PR #228, opened 2026-07-02
fix(build): print actual coverage percentage before failing gate

leseb:
## Summary
- The `coverage-check` Makefile target used `--fail-under-lines` which exits non-zero without printing the actual percentage, making it impossible to see how far off coverage is from CI logs
- Split into two steps: generate the JSON report first, then parse and print the line coverage percentage before checking the threshold
- This will let us see the actual percentage in CI so we can set a realistic temporary threshold

## Test plan
- [ ] Verify coverage workflow prints `Line coverage: XX.XX%` in CI logs before pass/fail

Signed-off-by: Sébastien Han <seb@redhat.com>

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -97,8 +97,8 @@ coverage-check:
 		--output-path coverage.json
 	@LINE_PCT=$$(jq '.data[0].totals.lines.percent' coverage.json); \
 	echo "Line coverage: $${LINE_PCT}%"; \
-	if [ $$(echo "$${LINE_PCT} < 95" | bc -l) -eq 1 ]; then \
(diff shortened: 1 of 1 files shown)


leseb:
I have a few PRs in flight to bump the coverage back to 95%:

- https://github.com/praxis-proxy/ai/pull/234
-  https://github.com/praxis-proxy/ai/pull/235
- https://github.com/praxis-proxy/ai/pull/236
- https://github.com/praxis-proxy/ai/pull/237

But i'd like to move fw with this temporarily so we can unblock main and other PRs can go through.

shaneutt (approved):

PR #241, opened 2026-07-02
fix(apis): match OpenAI JSON field ordering in conversation responses

leseb:
## Summary

Our conversation API responses serialized JSON fields in alphabetical order (`created_at`, `id`, `metadata`, `object`) instead of matching OpenAI's wire format (`id`, `object`, `created_at`, `metadata`).

- Enable `serde_json`'s `preserve_order` feature so `json!({})` macros emit fields in declaration order
- Deduplicate the create handler's inline JSON by reusing `conversation_to_json()`
- Add a regression test asserting field order matches OpenAI's API

## Test plan

- [x] `make lint` passes
- [x] `make build` passes
- [x] `make test` passes (all 1213 tests)
- [x] Manual verification: started server with conversations config, confirmed `curl` response has correct field ordering

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -36,7 +36,7 @@ regex = "1.12.4"
 reqwest = { version = "0.12.28", default-features = false, features = ["rustls-tls"] }
 secrecy = { version = "0.10.3", features = ["serde"] }
 serde = { version = "1.0.228", features = ["derive", "rc"] }
-serde_json = "1.0.150"
+serde_json = { version = "1.0.150", features = ["preserve_order"] }
 syn = { version = "2.0.118", features = ["full", "extra-traits", "visit"] }
 serde_yaml = { package = "yaml_serde", version = "0.10.4" }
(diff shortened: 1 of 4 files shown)


Cargo.toml
```diff
@@ -36,7 +36,7 @@ regex = "1.12.4"
 reqwest = { version = "0.12.28", default-features = false, features = ["rustls-tls"] }
 secrecy = { version = "0.10.3", features = ["serde"] }
 serde = { version = "1.0.228", features = ["derive", "rc"] }
-serde_json = "1.0.150"
+serde_json = { version = "1.0.150", features = ["preserve_order"] }
```
aslakknutsen, line 39: We're ok with preserving order of json globally, or should this fix be scoped to the conversation api?

leseb: the responses api also needs that, so i'd would have this globally, are you concerned about a potential performance hit from preserve_order or something else?

aslakknutsen: performance/memory was the base concern. But it's probably minimal and can always be adjusted later if proven a problem.

aslakknutsen (approved):

Issue #148, opened 2026-04-03
Epic: AI Agentic Protocol Support


shaneutt:
Protocol support for AI agent communication patterns: MCP (Model Context Protocol) and A2A (Agent-to-Agent).

## Status

The initial protocol foundation and useful production paths are implemented. The remaining work is primarily sessionful MCP backend routing, shared/distributed state, and explicitly scoped advanced/security integrations.

This epic was originally written as a nine-PR implementation plan. That plan is now historical: the filters moved from Praxis core into this repository, the MCP stateless profile was added, and several original tracking issues were completed or superseded. The sections below describe the current state.

## Shipped

- [x] Shared JSON-RPC parsing and durable filter metadata (praxis-proxy/ai#152, praxis-proxy/praxis#101)
- [x] Reserved internal-header hygiene (praxis-proxy/praxis#180)
- [x] Reusable MCP and A2A integration-test servers (praxis-proxy/praxis#369)
- [x] MCP classification: method, name, protocol version, session presence, and header/body validation (praxis-proxy/ai#154)
- [x] Static MCP catalog aggregation and namespace prefixes
- [x] Stateless MCP `tools/call` routing, prefix removal, backend path rewrite, and framing repair (praxis-proxy/ai#247, praxis-proxy/ai#340)
- [x] A2A classification, task/context extraction, local ownership routing, and SSE response capture (praxis-proxy/ai#157, praxis-proxy/ai#226)
- [x] StreamBuffer body-mutation adapter behavior (praxis-proxy/ai#212)
- [x] Broker-to-load-balancer pipeline validation without a dummy router (praxis-proxy/ai#213)

## Operating modes on `main`

| Mode | Pipeline | Current behavior |
| --- | --- | --- |
| Pass-through | `router -> load_balancer` | Proxies MCP/A2A HTTP traffic without protocol interpretation. |
| Classifier | `mcp` or `a2a` -> `router` -> `load_balancer` | Promotes body-derived protocol metadata for normal routing. |
| Stateless MCP broker | `mcp` -> `load_balancer` | Serves discovery/catalog operations and routes `tools/call` through the configured static catalog. |
| Current/sessionful MCP broker | `mcp` -> `load_balancer` | Handles local initialize/catalog operations, but backend `tools/call` remains unsupported until backend-session infrastructure lands. |
| A2A ownership routing | `a2a` -> `router` -> `load_balancer` | Captures task/context ownership and routes follow-up traffic using the local store. |

[...]

PR #247, opened 2026-07-02
feat(mcp): add configurable 2026-07-28 stateless profile

nerdalert:
## Summary

 Adds an opt-in MCP stateless broker profile targeting the upcoming [2026-07-28 release candidate](https://modelcontextprotocol.io/specification/draft), while preserving the existing session-based MCP broker as the default. The profile-based configuration can add explicitly implemented session-based or legacy HTTP+SSE MCP support later, while keeping each supported version and its transport semantics independently selectable. We may want to drop legacy support once stateless is GA, but either way, having a swappable profile is a clean approach.

  The implementation centralizes MCP version and profile handling so future finalized MCP versions can be added as explicit supported profiles without redesigning broker configuration or request dispatch. Configuration cannot enable arbitrary version strings: a version must be implemented and registered with its corresponding protocol behavior.

  This PR adds:

  - `current` and `stateless` MCP protocol profiles
  - Centralized profile-specific default and supported protocol versions
  - Profile/version config validation that rejects unsupported or cross-profile versions
  - Stateless `server/discover` capability discovery
  - Stateless `tools/list` responses with cache metadata
  - Required stateless request validation for `MCP-Protocol-Version`, `Mcp-Method`, and `Mcp-Name`
  - Validation that standard headers match the JSON-RPC body and `params._meta` metadata
  - Stateless validation of `clientInfo` and `clientCapabilities`
  - Controlled errors for header mismatches, unsupported versions, and unsupported methods
  - Stateless behavior with no `initialize` session handshake, no `Mcp-Session-Id`, and `DELETE` rejected
  - Existing current-profile behavior retained as the default, including session-based `initialize`
  - Stateless broker example configuration, generated filter documentation, unit tests, integration tests, and example-config coverage

  ## Profiles

  MCP protocol versions change more than a version string. The `2026-07-28` profile changes request metadata, discovery, session behavior, and cache semantics.

  `ProtocolProfile` keeps those behaviors tied to the version they implement. Adding a future MCP version means defining its supported version set and protocol semantics in one place, rather than accepting a configured string that the broker cannot actually serve correctly.

  ## Scope

  This PR implements the static broker portion of the `2026-07-28` stateless profile.

  The existing `current` profile remains the default for compatibility. It continues to support the established session-based broker behavior.

  This does not yet add:

  - Backend `tools/call` forwarding
  - Tool-prefix lookup or prefix stripping
  - Backend path rewrite or multi-backend tool routing
  - Dynamic catalog discovery or subscriptions
  - `Mcp-Param-*` / `x-mcp-*` validation
  - Shared distributed broker state

  ## Validation:
  - `cargo test -p praxis-ai-filters a2a`
  - `cargo clippy -p praxis-ai-filters -- -D warnings`
  - `cargo +nightly fmt --all --check`
  - `cargo xtask lint-example-tests`
  - `git diff --check`

  Refs praxis-proxy/ai#148

### DIFF
--- a/filters/src/agentic/mcp/broker/mod.rs
+++ b/filters/src/agentic/mcp/broker/mod.rs
@@ -35,15 +35,30 @@ use praxis_filter::{
 };
 use tracing::{debug, trace};
 
-use self::config::{CatalogTool, McpBrokerConfig, build_config};
-use super::protocol;
+use self::config::{CacheScope, CatalogTool, McpBrokerConfig, build_config};
+use super::protocol::ProtocolProfile;
 
 // -----------------------------------------------------------------------------
 // Constants
 // -----------------------------------------------------------------------------
 
-/// Server name reported in MCP initialize responses.
+/// Server name reported in MCP responses.
 const SERVER_NAME: &str = "praxis";
+
+/// Server version reported in MCP responses.
+const SERVER_VERSION: &str = env!("CARGO_PKG_VERSION");
+
+/// JSON-RPC error code for stateless header mismatch.
+const ERR_HEADER_MISMATCH: i32 = -32020;
+
+/// JSON-RPC error code for unsupported protocol version.
+const ERR_UNSUPPORTED_VERSION: i32 = -32022;
+
+/// MCP `=?base64?` sentinel prefix.
+const BASE64_SENTINEL_PREFIX: &str = "=?base64?";
+
+/// MCP `=?base64?` sentinel suffix.
+const BASE64_SENTINEL_SUFFIX: &str = "?=";
 
 // -----------------------------------------------------------------------------
 // McpBrokerFilter
@@ -53,9 +68,9 @@ const SERVER_NAME: &str = "praxis";
 /// MCP servers and handles `initialize`, `tools/list`, `tools/call`, `ping`,
 /// and `notifications/initialized` directly as a static broker.
 ///
-/// This first MCP static catalog change short-circuits all methods: no request is
-/// forwarded to backends. `tools/call` returns a controlled `-32601` error
-/// until backend routing is added.
+/// The broker serves configured catalog operations locally while backend tool
+/// routing is not implemented. It deliberately returns `-32601` for
+/// `tools/call` rather than forwarding a request whose target is unresolved.
 ///
 /// # YAML
 ///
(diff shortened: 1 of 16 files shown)


filters/src/agentic/mcp/broker/mod.rs
```diff
@@ -126,6 +144,234 @@ impl McpBrokerFilter {
...
+        reason = "signature matches dispatch_current for consistent dispatch_method call"
+    )]
+    fn dispatch_stateless(
+        &self,
+        ctx: &mut HttpFilterContext<'_>,
+        value: &serde_json::Value,
+        envelope: &JsonRpcEnvelope,
+        method_str: &str,
+    ) -> FilterAction {
+        if method_str.starts_with("notifications/") {
```
aslakknutsen: SEP-2243 - headers still apply to notifications. The header table says Mcp-Method is required for “All requests and notifications”. Missing or mismatched headers -> 400 

SEP-2575 - notifications/initialized should not exist in stateless mode. Per-request _meta replaces it. Reject it?

nerdalert: @aslakknutsen  good catch! Stateless dispatch now validates the required MCP stateless headers before handling notification methods, then rejects `notifications/initialized` as unsupported with 404 / -32601. Current-profile `notifications/initialized` behavior is unchanged.

Added unit coverage for unsupported stateless notifications plus missing/mismatched `Mcp-Method`, and added an integration-style test for the stateless wire behavior.

Ty!

aslakknutsen: @nerdalert This effectively "disables" any notification support doesn't it? It's perfectly fine for the scope of the PR mind you. Just wondering :)

aslakknutsen (approved):

filters/src/agentic/mcp/broker/mod.rs
nerdalert: @aslakknutsen Apologies, I should have referenced this in my response to be more clear. The ignore is only in the stateless profile. Current-profile notification behavior is unchanged. For this PR that seems safest because the stateless draft removes `notifications/initialized` and does not define core client-to-server notifications over Streamable HTTP. 

The spec text I’m following in the [changelog](https://modelcontextprotocol.io/specification/draft/changelog) is: `“Make MCP stateless: remove the initialize/notifications/initialized handshake. Every request now carries its protocol version, client identity, and client capabilities in _meta (io.modelcontextprotocol/protocolVersion, io.modelcontextprotocol/clientInfo, io.modelcontextprotocol/clientCapabilities). Version mismatches return
  UnsupportedProtocolVersionError (SEP-2575).”`

  - current profile = existing MCP 2025-03-26 style broker behavior, including initialize, Mcp-Session-Id, notifications/initialized, and DELETE session cleanup.
  - stateless profile = upcoming 2026-07-28 behavior, no protocol session, no initialize handshake, request metadata in _meta on each request.

### Summary

  - Existing/session-based profile still accepts notifications/initialized with 202.
  - Stateless profile validates headers first, then rejects notifications/initialized with 404.
  - Missing/mismatched Mcp-Method on stateless notifications returns 400.
  
Ty!

aslakknutsen: Perfect! Thanks for the explanation.

PR #255, opened 2026-07-03
ci(coverage): bump line coverage threshold to 93%

leseb:
## Summary

- Raise the minimum line coverage gate from 90% to 93% in the Makefile `coverage-check` target
- Correct the coverage workflow step name which previously said ≥95% but enforced 90%
- Both the Makefile threshold and CI workflow comment now consistently say 93%


leseb:
almost back to 95%!

aslakknutsen (approved):

PR #245, opened 2026-07-02
chore: remove stale ai-inference feature flag

leseb:
## Summary

- Remove the `ai-inference` feature from test manifests (`tests/integration`, `tests/schema`, `tests/utils`)
- Remove all `#[cfg(feature = "ai-inference")]` gates from test source files
- Remove stale "Requires the ai-inference feature" comments from 6 example YAML configs

The feature was a leftover from when AI code lived inside praxis core. It forwards to `praxis-filter/ai-inference` which does not exist upstream, making it dead wiring that would break compilation when test crates are re-enabled.

## Test plan

- [x] `make lint` passes (clippy + nightly fmt check)
- [x] Verified `praxis-filter` does not define `ai-inference`
- [x] No CI or Makefile targets use `--no-default-features` or `--features ai-inference`

PR #244, opened 2026-07-02
ci(integration): add integration test workflow and enable test crates

leseb:
## Summary

- Add `.github/workflows/integration.yaml` adapted from praxis core, running `make test-schema` and `make test-integration` on push/PR to main
- Re-enable `tests/integration`, `tests/schema`, and `tests/utils` as workspace members (previously commented out pending migration)
- Wire test utilities to use the AI filter registry (`praxis_ai::build_full_registry()`) instead of core-only builtins
- Prune test modules and config path references that belong to praxis core, not this repo

## Changes

- **Workflow**: New `integration.yaml` with checkout, clone-praxis, stable Rust, cargo cache, schema + integration test steps
- **Makefile**: Add `test-schema` target, fix `test-integration` package name (`praxis-tests-integration`)
- **Cargo.toml**: Uncomment test workspace members, add missing workspace deps (`h2`, `rcgen`, `rustls`, `rustls-pemfile`, `tokio-rustls`, `tokio-tungstenite`, `praxis`, `praxis-ai-proxy`), repoint `praxis-test-utils` to local `tests/utils`
- **Test crates**: Remove `ai-inference` feature forwarding to `praxis-filter`, strip `ai/` prefix from config paths, prune core-only test modules and configs
- **test-utils**: Use `praxis_ai::build_full_registry()` and `praxis_ai::run_server()` everywhere, fix unit tests to reference AI repo configs

## Test plan

- [x] `make lint` passes (clippy + fmt)
- [x] `make test-unit` passes (781+ tests)
- [x] `make test-schema` passes (101 tests)
- [x] `cargo test -p praxis-test-utils` passes (31 tests + 5 doc-tests)
- [x] `cargo test -p praxis-tests-integration --no-run` builds successfully
- [x] CI integration workflow runs green

PR #259, opened 2026-07-03
refactor: remove ai-inference feature gate

leseb:
## Summary

Remove the `ai-inference` Cargo feature flag that was re-introduced by PR #244 after PR #245 had removed it. All AI inference test modules are now compiled unconditionally — the feature gate is unnecessary in this dedicated AI repo.

Changes across 7 files: removed `ai-inference` feature definitions from three `Cargo.toml` files, removed all `#[cfg(feature = "ai-inference")]` gates from integration test module declarations, removed the "Feature Flags" section from the architecture doc, and updated an xtask test to use a generic example.

### DIFF
--- a/docs/architecture/ai-inference.md
+++ b/docs/architecture/ai-inference.md
@@ -141,8 +141,7 @@ IDs with `resp_` and `conv_` prefixes.
 ### `anthropic_messages_format`
 
 Classifies Anthropic Messages API requests and
-promotes format metadata. Feature-gated behind
-`ai-inference`.
+promotes format metadata.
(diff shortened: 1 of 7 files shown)


alexsnaps (approved):
What's with `docs/proposals` ? Do we keep these immutable? Is there some amendment procedures or...?

leseb:
> What's with `docs/proposals` ? Do we keep these immutable? Is there some amendment procedures or...?

I don't think we have any agreement on this, however, amending to reflect something broken makes sense to me :)

Issue #211, opened 2026-06-10
Billing and audit data pipeline


shaneutt:
## What

Structured data pipeline for billing-grade token usage records suitable for chargeback, compliance audit trails, and cost attribution.

## Why

Operational metrics (Prometheus) are designed for dashboards and alerting, not billing. Enterprise AI platforms need a separate audit pipeline with complete, durable records of every request's token usage, cost, tenant, model, and provider. Envoy AI Gateway runs dual pipelines: Kafka for billing/compliance, Prometheus for operations. Without this, operators cannot implement chargeback or satisfy compliance audits.

## Goals

- Structured per-request records: tenant, model, provider, route, token counts (all types), computed cost, timestamp
- Pluggable sink (structured log, webhook, message queue)
- Billing-grade durability: no silent drops
- Distinct from operational metrics - different cardinality, retention, and guarantees

PR #260, opened 2026-07-04
Implements issue 211: a new filter that extracts token usage from AI

szedan-rh:
Implements issue 211: a new filter that extracts token usage from AI
  inference responses across all five providers (OpenAI, Anthropic,
  Google, Bedrock, Azure). Supports both SSE streaming and JSON
  non-streaming modes via content-type detection. Reuses the existing
  SSE scanner from the A2A module and the token extraction APIs.

### DIFF
--- a/apis/src/token_usage/mod.rs
+++ b/apis/src/token_usage/mod.rs
@@ -7,6 +7,9 @@
 //! `Bedrock`, Azure) into a common [`TokenUsage`] representation.
 
 mod providers;
+mod streaming;
+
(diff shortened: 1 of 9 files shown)


aslakknutsen (commented):
This does not seem to implement #211 

Probably reframe the PR by narrowing the title/body to "token_count filter", with a why and mark it as related to #211

aslakknutsen:
Closing in favor for #261 & #262

Issue #84, opened 2026-05-15
Streaming token counting via SSE event parsing


shaneutt:
Parse SSE `data:` events for incremental token usage accumulation during streamed responses. Emit final aggregated count on stream completion (when `[DONE]` or equivalent sentinel is received).

szedan-rh:
Fix was merged.

PR #261, opened 2026-07-05
feat(apis): add Deserialize to TokenUsageProvider and streaming token…

szedan-rh:
add Deserialize to TokenUsageProvider and streaming token extraction
Add serde Deserialize to TokenUsageProvider for YAML config parsing.
Add extract_streaming_tokens for providers that spread token counts across multiple SSE events (Anthropic message_start/message_delta, Bedrock ConverseStream metadata).


apis/src/token_usage/streaming.rs
```diff
@@ -0,0 +1,149 @@
...
+
+/// Usage object inside `message_delta.usage`.
+#[derive(Deserialize)]
+struct AnthropicDeltaUsage {
+    /// Tokens in the output completion.
+    output_tokens: u64,
+}
+
+/// Parses Anthropic streaming events for partial token counts.
+fn parse_anthropic_event(data: &[u8]) -> Option<(Option<u64>, Option<u64>)> {
```
aslakknutsen: Anthropic events missing type field or with mismatched type + nested usage not explicitly tested

aslakknutsen (commented):
PR description should include trailer with reference to #84 , e.g. 

Related to #84

apis/src/token_usage/streaming.rs
```diff
@@ -0,0 +1,149 @@
...
+/// Returns `Some((input, output))` where either field may be `None`
+/// when only one count is present in the event. Returns `None` when
+/// the event contains no token usage data.
+///
+/// Used as a fallback when [`extract_token_usage`] returns `None` for
+/// providers that distribute token counts across multiple events
+/// (Anthropic, Bedrock).
+///
+/// [`extract_token_usage`]: super::extract_token_usage
+pub fn extract_streaming_tokens(provider: TokenUsageProvider, event_data: &[u8]) -> Option<(Option<u64>, Option<u64>)> {
```
alexsnaps: This may have been discussed else where, and I'm not sure I really understand the "either fields maybe be `None`", tho I read `either` as an `XOR` personally. 
But I guess my question is, what's the semantic of `Some((None, None))` ?

apis/src/token_usage/streaming.rs
szedan-rh: Acknowledged, I will add tests for those edge cases

apis/src/token_usage/streaming.rs
szedan-rh: Some((None, None)) is never produced by the current implementations — the outer Option means "no token data in this event" and the inner Options allow partial data (input-only or output-only). we can do either:
  - Tighten the doc comment to clarify "at least one field will be Some when the outer Option is Some"
  - Or refactor to an enum like Partial { input: Option<u64>, output: Option<u64> } with a constructor that enforces at least one is set — though that may be over-engineering for an internal API

szedan-rh: Would you prefer to do with the second option to handle that?

alexsnaps: I have so many questions 🤦 
I'd get rid of the outer `Option`, i.e. "just" have `(Option<u64>, Option<u64>)`, where then `None` becomes `(None, None)`. It's a case that has to be dealt with anyways on callsite with this current API - which is really what I'm challenging here: avoiding every caller to ask themselves the same question and having to deal with the `Some((None, None))` case. Does that make sense?

On the introduction of a `struct`, I'd be in favor, but not for the construction constrains (I think that doesn't address the problem of the current API, as described above), but rather that dealing with tuple is error prone, I'd rather _name_ the `tuple.0` & `tuple.1` "fields" explicitly to `input` and `output`. It's about code readability and conveying intents. 

Now if you _know_ we will need `Some((None, None))` eventually then, my question remains: what's the semantic of that. And how would that be best made explicit. 

And finally, is there a reason why `extract_streaming_tokens` isn't implemented on `TokenUsageProvider`, but is rather a "loose helper function"? It literally calls for it, no?

alexsnaps: > And finally, is there a reason why `extract_streaming_tokens` isn't implemented on `TokenUsageProvider`, but is rather a "loose helper function"? It literally calls for it, no?

Because the pattern was introduced with `extract_token_usage` is probably the answer, right?

alexsnaps: Right, and (because of these sigs?) `TokenUsageProvider` derives `Copy`... Which isn't "bad", but also I don't see why _really_. And it now closes the door to pushing any much more "knowledge" in the form of state into that type (or at least not without possibly paying quite a high price). 

@yehuditkerido Is there a reason for these choices that I'm missing?

alexsnaps: wdyt? https://github.com/praxis-proxy/ai/pull/284

szedan-rh: Good points across the board. For the return type, I agree — I'll flatten to (Option<u64>, Option<u64>) and introduce a StreamingTokens struct with named input/output fields so callers don't deal with  positional ambiguity.
On moving to methods on TokenUsageProvider — makes sense, the free functions are just dispatching on self anyway. Happy to rebase on top of #284 once it lands and move extract_streaming_tokens to a method as well.

yehuditkerido: @alexsnaps Hi, both choices were just simple/practical decisions at the time:

- Copy because the enum had no data
- Free function - no particular reason

No reason to keep either. PR #284 makes sense to me.

alexsnaps: >Copy because the enum had no data

That's the "wrong way" to think about it in my opinion. Cause if then `Copy` is leverage (which it was, given the usage in the free function sigs), introducing state becomes more challenging. 

Now, I'll resolve this conversation and consider it "addressed for now". I'll address the points in #284 myself _if no actual reason for the current design emerges_, but that discussion can happen there, so this gets unblocked and can progress.

apis/src/token_usage/streaming.rs
alexsnaps: Let's "just" 🔥 the outer `Option` and have `None` return `(None, None)` instead please 🙏

alexsnaps (dismissed):
... and addressing the other comments please too

apis/src/token_usage/streaming.rs
szedan-rh: Done

szedan-rh:
@alexsnaps - Comments was addressed.
Could you please take another look?

apis/src/token_usage/tests.rs
```diff
@@ -275,3 +275,164 @@ fn error_response_returns_none() {
...
+// -----------------------------------------------------------------------------
+
+#[test]
+fn provider_deserializes_lowercase() {
+    let providers = [
+        ("\"openai\"", TokenUsageProvider::OpenAi),
+        ("\"anthropic\"", TokenUsageProvider::Anthropic),
+        ("\"google\"", TokenUsageProvider::Google),
+        ("\"bedrock\"", TokenUsageProvider::Bedrock),
+        ("\"azure\"", TokenUsageProvider::Azure),
```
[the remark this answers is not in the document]
alexsnaps: Was that comment from a prior diff?

apis/src/token_usage/tests.rs
```diff
@@ -275,3 +275,164 @@ fn error_response_returns_none() {
...
+
+#[test]
+fn streaming_empty_returns_none() {
+    let result = extract_streaming_tokens(TokenUsageProvider::Anthropic, b"");
+
+    assert!(
+        result.is_none(),
+        "empty input should return None for streaming extraction"
+    );
+}
```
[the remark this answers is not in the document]
alexsnaps: I'd add these tests, if only to document the behavior

apis/src/token_usage/tests.rs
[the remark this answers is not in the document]
aslakknutsen: No, the GH view is just slightly off; 

```
       ("\"openai\"", TokenUsageProvider::OpenAi),
       ("\"open_ai\"", TokenUsageProvider::OpenAi),
```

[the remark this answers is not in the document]
szedan-rh: Addressed — `("\"open_ai\"", TokenUsageProvider::OpenAi)` is already in the current diff. GitHub line anchoring was just off after the force-push

apis/src/token_usage/tests.rs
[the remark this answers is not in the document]
szedan-rh: Added all six in the latest push: `anthropic_missing_type_field_returns_none`, `anthropic_wrong_type_with_usage_returns_none`, `anthropic_message_start_with_null_message`, `anthropic_message_start_with_null_usage`, `bedrock_null_metadata`, and `bedrock_null_usage_inside_metadata`.

aslakknutsen (approved):

PR #262, opened 2026-07-05
add token_count filter for streaming and non streaming reponses

szedan-rh:
add token_count filter for streaming and non-streaming responses Implements issue #84 . Extracts token usage from AI inference responses across all five providers via content-type detection. Reuses the SSE  scanner from A2A and the token extraction APIs. Writes unified counts to filter_metadata for downstream consumers like token_usage_headers.


szedan-rh:
Depends on #261. Please merge that first — this PR will be rebased after to show only the filter changes.

filters/src/token_count/mod.rs
```diff
@@ -0,0 +1,501 @@
...
+        .and_then(|v| v.parse().ok())
+        .unwrap_or(0);
+
+    if input > 0 || output > 0 {
+        set_token_usage(ctx, input, output, None);
+        debug!(input, output, "finalized streaming token counts");
+    }
+}
+
+// -----------------------------------------------------------------------------
```
aslakknutsen, line 348: Spec proposal [00211](https://github.com/praxis-proxy/ai/blob/main/docs/proposals/00211_streaming-token-counting.md#sse-scanner-reuse); suggested promoting shared hex helpers; duplication is consistent with current A2A pattern but increases drift risk.

server/src/lib.rs
```diff
@@ -64,6 +64,10 @@ fn register_general_ai_filters(registry: &mut praxis_filter::FilterRegistry) {
         @register registry,
         http "prompt_enrich" => praxis_ai_filters::PromptEnrichFilter::from_config
     );
+    praxis_filter::register_filters!(
+        @register registry,
+        http "token_count" => praxis_ai_filters::TokenCountFilter::from_config
```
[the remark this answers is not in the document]
szedan-rh: Good catch — added `tests/integration/tests/suite/examples/token_count.rs` following the `token_usage_headers` template. It verifies the proxy boots and proxies correctly with the `token_count` filter in the pipeline.

filters/src/token_count/tests.rs
```diff
@@ -0,0 +1,631 @@
...
+        ctx.get_metadata("token.input").is_none(),
+        "overflow should not set tokens"
+    );
+    let working_keys: Vec<_> = ctx
+        .filter_metadata
+        .keys()
+        .filter(|k| k.starts_with(META_PREFIX))
+        .collect();
+    assert!(working_keys.is_empty(), "overflow should clear all working metadata");
+}
```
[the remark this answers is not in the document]
szedan-rh: Agreed, the SSE overflow path has meaningfully different semantics from JSON overflow — it finalizes partial counts before clearing state. Added `sse_overflow_finalizes_partial_counts` which does exactly the four steps you outlined: sends valid usage events, triggers overflow via a chunk exceeding `DEFAULT_MAX_SCRATCH_BYTES`, asserts partial counts are finalized, and verifies all working metadata is cleaned up.

filters/src/token_count/tests.rs
```diff
@@ -0,0 +1,631 @@
...
+#[tokio::test]
+async fn sse_google_final_usage_event() {
+    let events = b"data: {\"candidates\":[{\"content\":{\"parts\":[{\"text\":\"Hi\"}]}}]}\n\ndata: {\"usageMetadata\":{\"promptTokenCount\":15,\"candidatesTokenCount\":42,\"totalTokenCount\":57}}\n\n";
+
+    let (input, output, total) = run_sse_extraction(TokenUsageProvider::Google, events).await;
+
+    assert_eq!(input.as_deref(), Some("15"), "Google SSE input tokens");
+    assert_eq!(output.as_deref(), Some("42"), "Google SSE output tokens");
+    assert_eq!(total.as_deref(), Some("57"), "Google SSE total tokens");
+}
```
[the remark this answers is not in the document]
szedan-rh: Fair point — having API-layer coverage without a filter-level integration was a gap. Added `sse_bedrock_metadata_event` exercising the full `on_response` → `on_response_body` cycle with a ConverseStream metadata event, consistent with the other provider tests.

examples/configs/token-counting.yaml
```diff
@@ -0,0 +1,41 @@
...
+#
+# Usage:
+#   cargo run -p praxis-ai-proxy -- -c examples/configs/token-counting.yaml
+#   curl http://localhost:8080/v1/chat/completions -d '{"model":"gpt-4","messages":[{"role":"user","content":"hi"}]}'
+#
+# The filter reads the response body (streaming SSE or buffered JSON),
+# extracts provider-specific token usage, and writes unified counts
+# to filter_metadata as token.input, token.output, and token.total.
+#
+# Downstream filters like token_usage_headers can then read these
```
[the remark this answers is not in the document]
szedan-rh: You are right — `token_usage_headers` reads during `on_response` which completes before any `on_response_body` runs, so the metadata would never be visible in the same pipeline. Updated the comment to clarify that counts are available to access logs, metrics, and filters that read metadata during the response body phase.

filters/src/token_count/mod.rs
```diff
@@ -0,0 +1,502 @@
...
+        trace!(input = inp, "partial input tokens from SSE event");
+    }
+    if let Some(out) = output {
+        merge_accumulated_count(ctx, META_OUTPUT, out);
+        trace!(output = out, "partial output tokens from SSE event");
+    }
+}
+
+/// Merge a partial count into accumulated state via max (last-writer wins
+/// for providers that repeat counts, additive for those that don't).
```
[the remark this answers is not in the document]
szedan-rh: Spot on — "additive" was misleading given the `.max()` implementation. Adopted your suggested wording verbatim, it captures the semantics much more precisely.

tests/integration/tests/suite/examples/token_count.rs
```diff
@@ -0,0 +1,36 @@
...
+        proxy_port,
+        HashMap::from([("127.0.0.1:3000", backend_port)]),
+    );
+    let proxy = start_proxy(&config);
+
+    let raw = http_send(
+        proxy.addr(),
+        "GET / HTTP/1.1\r\nHost: localhost\r\nConnection: close\r\n\r\n",
+    );
+    assert_eq!(parse_status(&raw), 200, "proxy should return 200");
```
aslakknutsen, line 35: This will pass with or without the token_filter. Is there a way to assert the tokens counted?

examples/configs/token-counting.yaml
```diff
@@ -0,0 +1,41 @@
...
+    filters:
+      - filter: router
+        routes:
+          - path_prefix: "/"
+            cluster: backend
+
+      - filter: token_count
+        provider: openai
+
+      - filter: token_usage_headers
```
aslakknutsen: In reference to https://github.com/praxis-proxy/ai/pull/262#discussion_r3550902379

Should this be removed as well to not cause any confusion or future copy/paste error on the user side?

aslakknutsen (changes requested):

filters/src/token_count/mod.rs
szedan-rh: Agree — the hex helpers (`decode_hex`, `hex_digit`, `set_hex_metadata`) and the SSE state persistence pattern are effectively duplicated between `token_count` and `a2a`. I kept it consistent with the existing A2A approach intentionally to keep this PR scoped, but I recognize the drift risk.

Happy to extract these into a shared utility module as a follow-up once this lands. That refactor would be cleaner as a standalone PR since it touches both filter modules.

tests/integration/tests/suite/examples/token_count.rs
szedan-rh: You are right — `token_count` writes to `filter_metadata` which is purely internal, so there is no HTTP-observable side effect to assert against from an integration test. The test as written validates that the config is accepted, the filter initializes without error, and the proxy proxies traffic correctly with `token_count` in the pipeline.

The actual extraction logic is thoroughly covered by unit tests at both the API layer (`apis/src/token_usage/tests.rs`) and the filter layer (`filters/src/token_count/tests.rs`) — including end-to-end `on_response` → `on_response_body` cycles for all five providers, chunked SSE, overflow, and edge cases.

To make token counts observable externally we would need a filter that reads metadata during `on_response_body` and surfaces it (e.g. via headers or access logs), but coupling that into this example would conflate the two concerns. Open to suggestions if you see a cleaner approach.

examples/configs/token-counting.yaml
szedan-rh: Good call — removed it. Having `token_usage_headers` in this example config was misleading given the timing semantics, and would have been a copy/paste trap for users. The example now only includes `token_count` itself.

aslakknutsen (approved):

PR #269, opened 2026-07-06
fix: point issue template discussions link to ai repo

leseb:
## Summary

The discussions contact link in the issue template config pointed to `praxis-proxy/praxis` instead of `praxis-proxy/ai`. This fixes it to point to the correct repository's discussions.


alexsnaps (approved):

alexsnaps:
I wasn't sure whether we maybe wanted all discussions under `/proxy`, but that's also my preference. so :shipit:

PR #272, opened 2026-07-06
fix(responses): case-insensitive SSE content-type detection

### LINKED ISSUE #267: Bug: Responses validation rewrites valid mixed-case SSE error content types
## Bug description
The Responses validation filter detects already-SSE backend error responses with a case-sensitive `starts_with("text/event-stream")` check. HTTP media types are case-insensitive, and the same feature already uses parameter-aware, case-insensitive helpers elsewhere.

A valid upstream error response with `Content-Type: Text/Event-Stream; charset=utf-8` can be treated as non-SSE and reformatted. For streaming Responses requests, `apply_error_response_headers` can then rewrite the status to `200` and replace the content type/body framing instead of passing the upstream SSE error through.

## Evidence
- `apis/src/openai/responses/validate/mod.rs:170-178` (`should_reformat_error` uses case-sensitive `starts_with`)
- `apis/src/openai/responses/validate/mod.rs:183-203` (`apply_error_response_headers` rewrites streaming errors to status `200` and `text/event-stream`)
- `apis/src/openai/responses/validate/mod.rs:1205-1227` existing test covers lowercase `text/event-stream` only
- `apis/src/openai/responses/store/filter.rs:541-558` shows the sibling parameter-aware, case-insensitive content-type helper pattern

## Steps to reproduce
1. Route a streaming OpenAI Responses request through the validation filter (`responses.stream=true`).
2. Have the backend return a non-2xx response with `Content-Type: Text/Event-Stream; charset=utf-8`.
3. Observe that `should_reformat_error` does not recognize the response as already SSE and sets `responses._reformat_error`; the filter can rewrite response headers/status instead of preserving the upstream SSE error.

## Expected behavior
Already-SSE error responses should be detected by parsing the media type, trimming parameters, and comparing `text/event-stream` case-insensitively.

## Suggested fix
Replace the `starts_with("text/event-stream")` check with a shared helper that splits `;`, trims whitespace, and uses `eq_ignore_ascii_case("text/event-stream")`. Add a mixed-case/parameterized SSE regression test for the validation filter.

## Source
Found by `clawpatch review --feature feat_apis_src_no_token_usage_codex_xhigh --limit 1` in run `20260706T070252-0f38c7`. `token_usage` was excluded from scope. No matching existing GitHub issue was found before filing.

leseb:
## Summary

- Replace case-sensitive `starts_with("text/event-stream")` in `should_reformat_error` with a parameter-aware, case-insensitive media type check that splits on `;`, trims whitespace, and uses `eq_ignore_ascii_case`
- Prevents the validation filter from rewriting valid upstream SSE error responses when the backend uses mixed-case or parameterized Content-Type headers (e.g. `Text/Event-Stream; charset=utf-8`)
- Adds a regression test for the mixed-case parameterized SSE content type

Closes #267

### DIFF
--- a/apis/src/openai/responses/validate/mod.rs
+++ b/apis/src/openai/responses/validate/mod.rs
@@ -174,7 +174,13 @@ fn should_reformat_error(ctx: &HttpFilterContext<'_>) -> bool {
             .headers
             .get(http::header::CONTENT_TYPE)
             .and_then(|v| v.to_str().ok())
-            .is_some_and(|ct| ct.starts_with("text/event-stream"));
+            .is_some_and(|ct| {
+                ct.split(';')
+                    .next()
+                    .unwrap_or_default()
+                    .trim()
+                    .eq_ignore_ascii_case("text/event-stream")
+            });
         is_error && !already_sse
     })
 }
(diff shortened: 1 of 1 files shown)


apis/src/openai/responses/validate/mod.rs
```diff
@@ -174,7 +174,13 @@ fn should_reformat_error(ctx: &HttpFilterContext<'_>) -> bool {
             .headers
             .get(http::header::CONTENT_TYPE)
             .and_then(|v| v.to_str().ok())
-            .is_some_and(|ct| ct.starts_with("text/event-stream"));
+            .is_some_and(|ct| {
```
aslakknutsen, line 177: nit; 	Fourth copy of is_event_stream_content_type logic (store/filter, anthropic/stream_events, filters/a2a, now inline here). Issue #267 suggested a shared helper

aslakknutsen (approved):

apis/src/openai/responses/validate/mod.rs
leseb: i'll send out a refactor pr after this

leseb: https://github.com/praxis-proxy/ai/pull/274

PR #284, opened 2026-07-06
refactor(token_usage): move functions to `TokenUsageProvider` directly

alexsnaps:
Maybe I'm missing something, but I don't see why we treat the `TokenUsageProvider` type and the associated functions as disjoint. Also, is there a reason why the type itself would _need to be_ `Clone` but mostly `Copy`?


aslakknutsen:
Closing, revisit later.

PR #295, opened 2026-07-07
Add agent Rust data ownership guidance

franciscojavierarceo:
## Summary

- add root AGENTS.md guidance for this repo
- document that Rust code should avoid unnecessary cloning of request, response, header, body, and SSE data in this high-performance proxy
- make .claude/CLAUDE.md a symlink to ../AGENTS.md so Claude and other agent guidance stay in sync
- correct the agent-guidance change that was previously opened against praxis-proxy/praxis instead of praxis-proxy/ai

## Validation

- cmp AGENTS.md .claude/CLAUDE.md
- git diff --check


leseb (approved):

aslakknutsen:
Removed from Queue for a second. The Unicode Safety Check is throwing up a lot of warnings about hidden chars, but not failing for some reason..

aslakknutsen:
Ah, the findings are categorized as MEDIUM, which is by default Warn only.  Only HIGH and CRITICAL blocks/fails the job. 

Found chars are not hidden smuggling characters, but typographic punctuations.

So, unless we want to enforce ASCII only; nothing to see here.. Move along.. :)

PR #300, opened 2026-07-07
Replace README banner

franciscojavierarceo:
## Summary

- Replace the README banner with the provided Praxis AI image.
- Store the banner as a tracked repository asset instead of a GitHub user-attachment URL.

## Validation

- Ran `git diff --check`.
- Confirmed the PNG dimensions/type with `file`.

ChatGPT Image Jul 7, 2026, 11_50_56 AM


shaneutt (commented):
Looks nice! Thoughts:

1. Did you want the banner to speak more to the _inference_ capabilities? The way I read this, it's AI Agents focused
2. "Server" doesn't feel like it belongs well to the lexicon here, shouldn't this be "Proxy" or "Proxy Server"?

franciscojavierarceo:
@shaneutt I was aiming to disambiguate between the proxy server (core Praxis) and the Stateful server (Praxis AI) which will serve the Conversations, Responses, Messages, and other APIs. 

LMK what you think.

shaneutt:
> disambiguate between the proxy server (core Praxis) and the Stateful server (Praxis AI)

I guess the problem I see is that the primary product of this repository _will in fact be a proxy server build_. :thinking:

franciscojavierarceo:
Yeah, that's very fair and that's true today. 

But as we start to incorporate more than just the proxy server, I think the primary product of this repo will no longer be just the proxy server (at least as a percentage of the codebase).

I could be wrong, we can discuss this more with @leseb as well but I'll close for now then.

PR #310, opened 2026-07-08
ci(fmt): include test crates in rustfmt checks

nerdalert:
## Summary

Smol PR to satisfy my OCD 😬 and save anyone else some time if they run `fmt --all` that adds rustfmt check used by `make lint` to also covers the AI test crates.

Before this, `make lint` formatted only the main AI crates and `xtask`, so integration/schema/test utility files could drift unless someone ran `cargo +nightly fmt --all --check` manually.

This also applies the rustfmt-required wrap in the OpenAI stream events integration test that exposed the gap.

Thanks!

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -77,16 +77,14 @@ test-integration:
 # Quality
 # -------------------------------------------------------------------
 
-AI_PKGS := -p praxis-ai-proxy -p praxis-ai-filters -p praxis-ai-apis -p xtask
-
 lint:
 	cargo clippy --workspace --all-targets -- -D warnings
-	cargo +nightly fmt $(AI_PKGS) -- --check
+	cargo +nightly fmt --all -- --check
 	cargo xtask lint-separators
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -77,7 +77,8 @@ test-integration:
 # Quality
 # -------------------------------------------------------------------
 
-AI_PKGS := -p praxis-ai-proxy -p praxis-ai-filters -p praxis-ai-apis -p xtask
+AI_PKGS := -p praxis-ai-proxy -p praxis-ai-filters -p praxis-ai-apis -p xtask \
```
leseb: why not just do fmt for everything instead those endless `-p`? is this possible?

leseb (commented):
noticed that same openai fix this morning on a local branch, thanks!

Makefile
nerdalert: Good call! done 🫡

leseb (approved):

PR #326, opened 2026-07-09
ci: add optional praxis main testing and known-issue label

leseb:
## Summary

- Adds `praxis_main` workflow_dispatch input to Tests and Integration workflows to optionally build/test against praxis core `main` branch instead of the crates.io release (uses `[patch.crates-io]` overrides)
- Adds `praxis-known-issue` PR label support: when applied, jobs use `continue-on-error` so transient praxis core failures don't block PRs
- Adds `patch-praxis` / `unpatch-praxis` Makefile targets for local development against a sibling `../praxis` checkout
- Restores `clone-praxis` composite action with configurable `ref` input and adds new `patch-praxis` composite action

## Test plan

- [ ] Manually trigger Tests workflow with `praxis_main: true` on this branch
- [ ] Apply `praxis-known-issue` label and verify jobs report green on failure
- [ ] Run `make patch-praxis` locally with `../praxis` present, verify build


leseb:
I'm opening this as a draft since i think it's valuable to test regularly against praxis main but would like inputs from other maintainers too - @franciscojavierarceo @aslakknutsen

franciscojavierarceo (approved):

PR #327, opened 2026-07-09
test: change auth test header to avoid tripping scanners

aslakknutsen:
Change test auth header to some non real looking token to avoid tripping security scanners.


aslakknutsen:
Sorry, old alert. File removed in main. :)

PR #340, opened 2026-07-15
feat(mcp): route stateless tools/call through static broker catalog

nerdalert:
## Summary

  Adds backend forwarding for stateless MCP `tools/call` requests using the static broker catalog.

  The stateless broker can now receive a `tools/call`, validate the stateless MCP headers/body, look up the exposed tool name, select the configured backend cluster, rewrite the backend path, and forward the call with the backend tool name. This turns the static catalog from discovery-only into a working broker path for configured tools.

  With this PR, the catalog becomes the routing table for stateless `tools/call`:

  1. The client calls an exposed tool name, for example `weather_get_forecast`.
  2. Praxis looks up that exposed name in the configured static catalog.
  3. The matched catalog entry tells Praxis which backend server/cluster owns the tool.
  4. Praxis selects that cluster, rewrites the upstream path, and rewrites `params.name` from the exposed name to the backend tool name, for
  example `get_forecast`.
  5. Praxis also updates the forwarded `Mcp-Name` header so the backend sees a consistent MCP request.
  6. The normal load-balancer path then forwards the request to the selected backend.

  The temporary Praxis dependency pin is needed because this forwarding happens after the request body has been inspected and rewritten. The MCP filter sets the rewritten `Mcp-Name` header during the body-processing phase, and `praxis-proxy/praxis#760` is the core fix that preserves those body-phase header mutations before the request is sent upstream.
  
## Changes

  - Routes stateless `tools/call` by static catalog entry.
  - Rewrites `params.name` from exposed name to backend/original tool name.
  - Preserves `params.arguments` unchanged.
  - Sets the upstream `Mcp-Name` header to the backend tool name.
  - Encodes non-visible-ASCII MCP names with the existing `=?base64?...?=` sentinel format.
  - Rewrites the upstream request path to the selected server path.
  - Repairs `Content-Length` after request body mutation.
  - Returns `-32602 Invalid Params` for malformed `tools/call` params before header mismatch validation.
  - Keeps the `current` (`2025-03-26`) compatibility-profile `tools/call` behavior unchanged.
  - Updates docs and the stateless broker example to reflect working forwarding behavior.

  ## Dependency Note

  This draft temporarily pins `praxis-proxy-*` crates to `praxis-proxy/praxis` commit `2bbce4386e84f6c8a8a0e64c7cc47b52f6d34cf9` because the latest crates.io release, `0.4.0`, predates `praxis-proxy/praxis#760`.

  `#760` is required for body-phase header mutations from StreamBuffer filters to reach upstream backends, which allows routed MCP `tools/call` requests to forward the rewritten `Mcp-Name` header.

  This pin should be replaced with the next published `praxis-proxy-*` crate version once available.

  ## Protocol Boundary

  This PR completes base stateless `tools/call` routing.

  The following are intentionally left to separate focused PRs:

  - Locked-RC method-set cleanup, including stateless `ping` removal
  - `logging/setLevel` and legacy notification conformance cleanup
  - Schema-driven `x-mcp-header` / `Mcp-Param-*` support
  - `subscriptions/listen`
  - MRTR / `input_required`
  - Dynamic catalog discovery
  - Legacy MCP sessions
  - Redis/Valkey state

  ## Validation

  - `cargo test -p praxis-ai-filters mcp` — 265 passed
  - `cargo test -p praxis-tests-integration --test suite mcp_broker` — 30 passed
  - `cargo test -p praxis-tests-schema` — 90 passed
  - `cargo clippy --workspace --all-targets -- -D warnings`
  - `cargo +nightly fmt --all --check`
  - `cargo xtask lint-filter-docs`
  - `cargo xtask lint-example-tests`
  - `cargo xtask sync-example-readme`
  - `git diff --check`

  Refs praxis-proxy/ai#148

### DIFF
--- a/apis/src/lib.rs
+++ b/apis/src/lib.rs
@@ -74,11 +74,14 @@ pub(crate) mod test_utils {
             request_headers_to_remove: Vec::new(),
             request_headers_to_set: Vec::new(),
             filter_metadata: std::collections::HashMap::new(),
+            pre_read_mutations: Vec::new(),
+            structured_metadata: std::collections::HashMap::new(),
             filter_results: std::collections::HashMap::new(),
             filter_state: std::collections::HashMap::new(),
             health_registry: None,
             id_generator: &TEST_ID_GENERATOR,
             kv_stores: None,
+            peer_identity: None,
             request: req,
             request_body_bytes: 0,
             request_body_mode: praxis_filter::BodyMode::Stream,
@@ -91,9 +94,6 @@ pub(crate) mod test_utils {
             selected_endpoint_index: None,
             time_source: &praxis_core::time::SystemTimeSource,
             upstream: None,
-            peer_identity: None,
-            pre_read_mutations: Vec::new(),
-            structured_metadata: std::collections::HashMap::new(),
         }
     }
 
--- a/docs/architecture/agentic-protocols.md
+++ b/docs/architecture/agentic-protocols.md
@@ -92,8 +92,9 @@ Client --> MCP Broker Filter
              +-- tools/list:  aggregated catalog
              |                from all servers
              +-- ping:        local response
-             +-- tools/call:  not yet supported
-             |                (returns -32601)
+             +-- tools/call:  stateless routes by
+             |                configured catalog tool
+             |                current profile returns -32601
              +-- DELETE:      session termination
 ```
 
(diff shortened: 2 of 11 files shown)


aslakknutsen (changes requested):
docs/architecture/agentic-protocols.md diagram line 95 still reads "tools/call: not yet supported (returns -32601)" for the broker path. Filter docs were updated but this architecture page contradicts merged behavior and will mislead operators.

nerdalert:
> docs/architecture/agentic-protocols.md diagram line 95 still reads "tools/call: not yet supported (returns -32601)" for the broker path. Filter docs were updated but this architecture page contradicts merged behavior and will mislead operators.

@aslakknutsen ty!

- Updated `docs/architecture/agentic-protocols.md:95` so the broker diagram no longer
    says tools/call is unsupported.
- It now says stateless tools/call routes by configured catalog tool, while current profile still returns -32601.

aslakknutsen (approved):

PR #347, opened 2026-07-15
feat: orchestrate OpenAI web search through vLLM

franciscojavierarceo:
## What changed

- Add a Praxis-owned blocking web-search orchestrator for direct vLLM Core Responses requests.
- Normalize OpenAI web-search tool declarations into a function tool, call vLLM, execute You.com searches, and continue the tool loop until vLLM returns the final response.
- Add OpenAI Responses web-search SSE event variants and a runnable example configuration.
- Add unit and integration coverage for the orchestration loop, error paths, streaming rejection, round limits, and example config registration.

## Why

This demonstrates the Praxis + vLLM Core deployment where Praxis owns the You.com call and credential boundary. Requests without web search continue through the normal vLLM route. Streaming web-search orchestration is explicitly rejected for this first blocking demo.

## Validation

- `cargo test -p praxis-ai-apis openai::responses::web_search --lib`
- `cargo test -p praxis-ai-apis openai::sse::responses::event --lib`
- `cargo test -p praxis-tests-integration openai_web_search --no-default-features`
- `cargo clippy -p praxis-ai-apis -p praxis-ai-proxy --all-targets -- -D warnings`
- `git diff --check`

## Follow-up

The vLLM Agentic API path remains separate: when that service owns web-search hydration, Praxis should bypass this direct orchestrator and proxy to Agentic API.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -10,17 +10,6 @@ checksum = "320119579fcad9c21884f5c4861d16174d0e06250625266f50fe6898340abefa"
 
 [[package]]
 name = "ahash"
-version = "0.7.8"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "891477e0c6a8957309ee5c45a6368af3ae14bb510732d2684ffa19af310920f9"
-dependencies = [
- "getrandom 0.2.17",
- "once_cell",
- "version_check",
-]
-
-[[package]]
-name = "ahash"
 version = "0.8.12"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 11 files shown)


leseb (changes requested):
a few issues: 

* needs to use the http callout crate
* needs https://github.com/praxis-proxy/praxis/pull/805 impl because now you are bypassing and running your own loop AND calling vllm directly - this cannot land as is
* conflicts with https://github.com/praxis-proxy/ai/pull/349 😢 - i sort feel like https://github.com/praxis-proxy/ai/pull/349 superseeds this - PTAL

PR #388, opened 2026-07-16
fix(a2a): capture task route for non-SSE responses to streaming methods

### LINKED ISSUE #352: fix(a2a): streaming methods with non-SSE response skip task route capture
In `filters/src/agentic/a2a/mod.rs`, when `SendStreamingMessage` or
`SubscribeToTask` receives a success response with `Content-Type:
application/json` (non-SSE), `on_response` returns `Continue` without
enabling any capture path (line ~258).

Per the A2A spec, these methods MAY return a non-streaming JSON-RPC
response containing a full `Task` object. The current code loses the
task route in this case, so subsequent `GetTask`/`CancelTask` calls
for that task may go to the wrong backend.

**Fix:** When `is_sse_capable` is true but `is_sse` is false, fall
through to the non-streaming capture path instead of returning early.

mkoushni:
## Summary
`on_response` in `filters/src/agentic/a2a/mod.rs` returned `FilterAction::Continue`
early whenever the A2A method was SSE-capable (`SendStreamingMessage` or
`SubscribeToTask`), **regardless of whether the response was actually SSE**.
Per the A2A spec, these methods *may* return a non-streaming JSON-RPC response
containing a full `Task` object (e.g. when the backend completes synchronously
instead of streaming). In that case, the early return skipped the
non-streaming task-route capture path entirely, so the task route was never
recorded. Subsequent `GetTask`/`CancelTask` calls for that task could then be
routed to the wrong backend.

## Fix
Only take the SSE capture branch when the response is **both** SSE-capable
**and** actually SSE (`is_sse_capable && is_sse`). Otherwise, execution now
falls through to the existing non-streaming capture logic, which records
`a2a.response.cluster` and `a2a.response.capture_enabled` so the task route
gets stored from the response body.
```rust
if is_sse_capable && is_sse {
    let cluster = ctx.cluster_name().map(str::to_owned);
    if let Some(cluster) = cluster {
        ctx.filter_metadata
            .insert("a2a.response.sse_capture_enabled".to_owned(), "true".to_owned());
        ctx.filter_metadata.insert("a2a.response.cluster".to_owned(), cluster);
    }
    return Ok(FilterAction::Continue);
}

Fixes #352

### DIFF
--- a/filters/src/agentic/a2a/mod.rs
+++ b/filters/src/agentic/a2a/mod.rs
@@ -246,14 +246,12 @@ impl HttpFilter for A2aFilter {
             .and_then(|v| v.to_str().ok())
             .is_some_and(is_event_stream_content_type);
 
-        if is_sse_capable {
-            if is_sse {
-                let cluster = ctx.cluster_name().map(str::to_owned);
-                if let Some(cluster) = cluster {
-                    ctx.filter_metadata
-                        .insert("a2a.response.sse_capture_enabled".to_owned(), "true".to_owned());
-                    ctx.filter_metadata.insert("a2a.response.cluster".to_owned(), cluster);
-                }
+        if is_sse_capable && is_sse {
+            let cluster = ctx.cluster_name().map(str::to_owned);
+            if let Some(cluster) = cluster {
+                ctx.filter_metadata
+                    .insert("a2a.response.sse_capture_enabled".to_owned(), "true".to_owned());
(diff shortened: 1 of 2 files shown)


leseb (dismissed):
thanks but let's first do this https://github.com/praxis-proxy/ai/issues/352#issuecomment-4989987607

leseb (approved):

PR #392, opened 2026-07-16
fix(conversations): pass ids to sync_conversation_messages instead of a partial ConversationRecord

Nitjsefnie:
Refactors `sync_conversation_messages` to accept `tenant_id` and `conversation_id` as `&str` instead of a full `ConversationRecord`. This eliminates the partial record constructed in `filter.rs` with `created_at: 0` and `messages: Value::Null`, which could silently corrupt conversation data if the callee ever accessed those fields.

Verification:
- `cargo fmt -p praxis-ai-apis` (nightly rustfmt) — clean
- `cargo clippy -p praxis-ai-apis --all-targets -- -D warnings` — clean
- `cargo test -p praxis-ai-apis` — all 1456 tests passed

Closes #358

Prepared with AI assistance (Kimi K2.7 Code), human-reviewed before submission.

PR #331, opened 2026-07-10
feat(filter): add mcp_tool_resolve filter for MCP tool discovery

leseb:
## Summary

Adds Filter 18 (`mcp_tool_resolve`) which resolves MCP tool entries in Responses API `tools` arrays into concrete tool definitions by calling `tools/list` on upstream MCP servers.

### Key capabilities

- **MCP client module** (`apis/src/mcp_client/`): Streamable HTTP transport via `rmcp`, with SSRF validation (loopback, link-local, unspecified, cloud metadata IPs, AWS IMDS IPv6), DNS pinning to prevent rebinding, URL userinfo rejection, hop-by-hop/framing header blocking, and reserved prefix stripping (`x-praxis-*`, `x-mcp-*`, `x-a2a-*`)
- **Tool resolution**: Paginated `tools/list` with configurable `max_tools` cap (default 128), `max_servers` limit (default 10), and per-server timeout (default 5s)
- **Cache layers**: Within-request dedup by `(label, url)` and cross-request reuse via `previous_tools` from rehydrated responses, with label-only matching for real API shapes
- **Filtering**: Supports both string-array and `MCPToolFilter` object forms of `allowed_tools`, including `read_only` annotation filtering
- **OAuth**: Dedicated `authorization` field injected as Bearer token; `Authorization` in `headers` is stripped
- **Continuation safety**: Skips `connector_id`-only entries, `defer_loading: true` entries, and entries without `server_label`

### Files changed (23)

- New `mcp_client` module with full SSRF/DNS/transport hardening
- New `mcp_tool_resolve` filter with config, resolution logic, and 113 unit tests
- Extended `ResponsesState` with `mcp_tool_map` and `previous_tools` fields
- Extended `rehydrate` to persist `server_url` in `mcp_list_tools` items
- Example config, integration tests, and generated filter docs

Closes #43

PR #394, opened 2026-07-16
fix(mcp): enforce length bounds on session_id and protocol_version

### LINKED ISSUE #361: fix(mcp): missing length bounds on session_id and protocol_version
In `filters/src/agentic/mcp/mod.rs`, `write_metadata()` enforces
`MAX_DYNAMIC_VALUE_LEN` on `method` and `name` but not on `session_id`
(line ~369) or `protocol_version` (line ~374). Only control characters
are checked.

Similarly, `promote_mcp_headers()` length-bounds `method` and `name`
before injecting into upstream headers, but `protocol_version` (line ~405)
skips the length check, allowing unbounded values in upstream headers.

**Fix:** Add `&& value.len() <= max_len` guards to `session_id` and
`protocol_version` in both `write_metadata` and `promote_mcp_headers`,
matching the existing pattern for `method` and `name`.

Nitjsefnie:
This change applies the existing `MAX_DYNAMIC_VALUE_LEN` guard to `session_id` and `protocol_version` in `write_metadata`, and to `protocol_version` in `promote_mcp_headers`, matching the pattern already used for `method` and `name`. Over-length values are now skipped instead of being stored in metadata or forwarded as upstream headers. Two unit tests verify that over-length `session_id` and `protocol_version` values are rejected/skipped while in-bounds values continue to pass.

Closes #361

Prepared with AI assistance (Kimi K2.7 Code), human-reviewed before submission.

### DIFF
--- a/filters/src/agentic/mcp/mod.rs
+++ b/filters/src/agentic/mcp/mod.rs
@@ -368,11 +368,13 @@ fn write_metadata(
     }
     if let Some(sid) = &mcp.session_id
         && !contains_control_chars(sid)
+        && sid.len() <= max_len
     {
         ctx.set_metadata("mcp.session_id", sid.clone());
     }
     if let Some(pv) = &mcp.protocol_version
         && !contains_control_chars(pv)
+        && pv.len() <= max_len
     {
         ctx.set_metadata("mcp.protocol_version", pv.clone());
     }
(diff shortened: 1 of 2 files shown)


leseb:
We need a human review on https://github.com/praxis-proxy/ai/issues/361#issuecomment-4991900871 first

Nitjsefnie:
Understood — leaving it closed, and I won't resubmit unless #361 gets a human review that says the change is wanted.

For what it's worth on timing: this PR went up before that note landed on #361, so it wasn't submitted around the gate. No objection to the outcome either way — your call on which issues are ready is the one that counts, and I'd rather you close work than carry review load you didn't ask for.

Also still happy to close any of the other four (#391, #392, #393, #395) if they're more than you want open right now — just say the word. Three of them are on issues you'd marked `Verdict: Valid`, and #395 is the deferred follow-up from your own #331 review, but that's context, not an argument.

Issue #30, opened 2026-06-01
Epic: Filter 11 - compact


leseb:
## Purpose

Token counting and context window management. Summarize conversation history when threshold is exceeded.

## Behavior

- Check `context_management` parameter for compaction config
- Count tokens via provider usage (from `previous_usage`) or tiktoken estimate
- If token count > threshold:
  - Build summarization prompt from conversation messages
  - Prepend `instructions` to summarization prompt
  - Call inference API with configured `compaction_model`
  - Extract summary text
  - Apply `summary_prefix` if configured
  - Create `OpenAIResponseCompaction` item with summary as `encrypted_content`
  - Replace conversation history with compaction item + preserved user messages
  - Store compaction response in response store (hidden, usable as `previous_response_id`)
  - Hide compaction items from `input_items` API response
- Update `responses.messages` with compacted messages
- Also handles explicit `POST /v1/responses/compact` requests

## Config

```yaml
filter: compact
default_model: gpt-4o-mini  # model for summarization
tiktoken_encoding: cl100k_base
```

## Praxis trait methods

- `on_request` — check token count, compact if needed

## Dependencies

- Inference backend (for summarization)
- Response store

## Reference

- [Proposal: Responses API Filters](docs/proposals/00354_responses-api-filters.md) — Filter 11
- Part of praxis-proxy/ai#93

PR #403, opened 2026-07-17
feat(compact): automatic compaction via context_management (#598)

### LINKED ISSUE #598: feat(compact): automatic compaction via context_management
## Parent

Part of #30 (compact filter)

## Context

The automatic compaction flow triggers when a multi-turn request includes a `context_management` array with a `type: compaction` entry and the conversation token count exceeds the configured threshold. This is the core compaction path — subsequent sub-issues build on top of it.

## Scope

- Parse `context_management` for compaction config (`compact_threshold`, `compaction_model`)
- Count tokens via tiktoken (`cl100k_base` / `o200k_base`), validated at startup
- Build summarization text from full conversation including tool calls and outputs
- Call inference backend with `compaction_model` override and `instructions` prepend
- Replace conversation history with canonical `{type: compaction, id, encrypted_content}` item
- Translate compaction items to assistant messages at backend boundary
- Respect `callout_failure_mode` (open/closed) for network and parse failures

## Reference

- PR #403 — implementation
- #595 — `summary_prefix` and `previous_usage` token counting (follow-up)
- #596 — hidden compaction store and `input_items` filtering (follow-up)
- #597 — `POST /v1/responses/compact` explicit endpoint (follow-up)

crstrn13:
## Summary

Implements automatic context window management for the OpenAI Responses API pipeline. Part of #30 — tracks as #598.

When a multi-turn request includes `context_management` with a `type: compaction` entry and the conversation token count exceeds the configured threshold, the filter summarizes the full conversation history via a Chat Completions callout and replaces it with a single canonical compaction item.

### How it works

1. **Token counting** — uses tiktoken (`cl100k_base` / `o200k_base`) to estimate the token count of the full conversation text, validated at startup.
2. **Threshold check** — if the count exceeds `compact_threshold`, compaction triggers. A missing or malformed threshold skips compaction.
3. **Summarization callout** — sends the conversation (including tool calls and outputs) to a Chat Completions endpoint and receives a summary.
4. **Message replacement** — replaces the full history with a canonical `{type: compaction, id, encrypted_content}` item (base64-encoded summary), preserving the current turn's input.
5. **Backend translation** — compaction items are translated to assistant-level messages at the backend boundary so backends receive a valid input item.

Runs after `openai_tool_parse` and `openai_responses_rehydrate`. Compaction only applies to multi-turn requests (`previous_response_id` or `conversation`) — single-turn requests are intentionally skipped.

### Files

- `apis/src/openai/responses/compact/` — filter implementation, config, unit tests
- `apis/src/openai/responses/config_validation.rs` — shared callout settings extracted from web_search
- `server/src/lib.rs` — filter registration
- `examples/configs/openai/responses/compact.yaml` — example config
- `tests/integration/tests/suite/examples/compact.rs` — integration tests

## Test plan

- [x] Unit tests: `cargo test -p praxis-ai-apis -- compact` (53 tests)
- [x] Integration tests: `cargo test -p praxis-tests-integration -- compact` (3 tests)
- [x] Full lint: `make lint` passes
- [x] End-to-end verified against Ollama (`qwen2.5:0.5b`)

## Scope

Closes #598. The remaining items from #30 are tracked as:

- #595 — `summary_prefix` config and `previous_usage` token counting
- #596 — persist compaction as hidden response and filter from `input_items`
- #597 — `POST /v1/responses/compact` explicit endpoint


apis/src/openai/responses/compact/mod.rs
```diff
@@ -0,0 +1,440 @@
...
+            .ok_or_else(|| "Chat Completions response missing choices[0].message.content".to_owned()),
+        Err(err) => Err(format!("failed to parse Chat Completions response JSON: {err}")),
+    }
+}
+
+/// Build the compaction output item.
+///
+/// Returns: `{"role": "system", "content": "<summary>"}`
+fn build_compaction_item(summary: &str) -> Value {
+    serde_json::json!({
```
franciscojavierarceo, line 494: Could we represent this as a canonical Responses compaction item rather than a system message? The summary is model-generated from user-controlled conversation history, so persisting it as `role: system` raises its instruction priority. It also cannot round-trip as the `OpenAIResponseCompaction` state described in #30. A safer shape would be `{type: "compaction", id, encrypted_content}`, translated to an assistant-level message only at a Chat Completions backend boundary if necessary.

apis/src/openai/responses/compact/mod.rs
```diff
@@ -0,0 +1,440 @@
...
+}
+
+/// Extract text content from a message's `content` field.
+///
+/// Content can be a plain string, an array of content parts
+/// (each with a `"text"` field), or absent/null.
+///
+/// Returns `Cow::Borrowed` for plain strings (zero-copy) and
+/// `Cow::Owned` for array content that must be joined.
+fn extract_content(msg: &Value) -> Cow<'_, str> {
```
franciscojavierarceo, line 604: Could this serialization cover the complete canonical conversation state? Items such as `function_call.arguments`, `function_call_output.output`, and prior compaction `encrypted_content` currently produce an empty string and disappear from both the token estimate and the summary; tool definitions also contribute to rendered context outside these message contents. That can trigger compaction too late and leave the compacted conversation unable to continue correctly. Tests covering a tool round-trip and repeated compaction would help lock this down.

apis/src/openai/responses/compact/config.rs
```diff
@@ -0,0 +1,111 @@
...
+    #[serde(default = "default_tiktoken_encoding")]
+    pub tiktoken_encoding: String,
+
+    /// Callout timeout in milliseconds.
+    #[serde(default)]
+    pub timeout_ms: Option<u64>,
+
+    /// Failure mode for the inference callout.
+    #[serde(default)]
+    pub failure_mode: Option<FailureMode>,
```
franciscojavierarceo: I don’t think this setting can currently be populated from filter YAML: `parse_filter_config` removes `failure_mode` as a pipeline structural key before deserializing the filter-specific config, leaving this as `None` and defaulting the callout to closed mode. Could we use a non-reserved name such as `callout_failure_mode`, or explicitly pass the pipeline policy into the filter, with coverage for both open and closed behavior?

apis/src/openai/responses/compact/mod.rs
```diff
@@ -0,0 +1,440 @@
...
+        params: &CompactionParams,
+        streaming: bool,
+        conversation_text: &str,
+    ) -> Result<Option<String>, FilterAction> {
+        let model = params.compaction_model.as_deref().unwrap_or(&self.config.default_model);
+        let instructions = state.request_body.get("instructions").and_then(Value::as_str);
+        let request = build_summarization_request(conversation_text, instructions, model, &self.config.inference_url);
+
+        match self.callout_client.execute(request).await {
+            CalloutResult::Success(resp) => parse_summarization_response(&resp.body).map(Some).or_else(|e| {
```
franciscojavierarceo: Could parse failures follow the configured open/closed policy as well? A 2xx response with invalid JSON or without `choices[0].message.content` currently returns `Ok(None)` and silently skips compaction even in closed mode. Routing this through the same failure-outcome logic as the other callout failures, with a test for each mode, would make the behavior consistent.

apis/src/openai/responses/compact/mod.rs
```diff
@@ -0,0 +1,440 @@
...
+    let array = context_management.as_ref()?.as_array()?;
+
+    for entry in array {
+        let Some(entry_type) = entry.get("type").and_then(|v| v.as_str()) else {
+            continue;
+        };
+        if entry_type != "compaction" {
+            continue;
+        }
+        let compact_threshold = entry.get("compact_threshold").and_then(Value::as_u64).unwrap_or(0);
```
franciscojavierarceo: Could we distinguish an absent threshold from an invalid one here? Missing, `null`, negative, and incorrectly typed values all become zero, which unexpectedly triggers a summarization call for any non-empty history. It seems safer for absence to use an explicit operator default or skip compaction, while malformed values are rejected; an explicit numeric zero could still intentionally mean “compact immediately.”

tests/integration/tests/suite/examples/compact.rs
```diff
@@ -0,0 +1,118 @@
...
+    let proxy2 = start_proxy(&config2);
+
+    let raw2 = http_send(
+        proxy2.addr(),
+        &json_post(
+            "/v1/responses",
+            r#"{"model":"gpt-4.1","input":"Compare with QUIC","previous_response_id":"resp_compact","context_management":[{"type":"compaction","compact_threshold":50}]}"#,
+        ),
+    );
+    let status2 = parse_status(&raw2);
```
franciscojavierarceo, line 191: Would it be possible for this test to assert the compaction behavior rather than only the final status? It currently passes even if compaction never runs because the mock returns a successful response for either call. A recording or sequenced backend could verify that the above-threshold case makes a summarization call followed by an inference call, and that the latter contains compacted state without the old history; the below-threshold case should make only the inference call.

apis/src/openai/responses/compact/mod.rs
```diff
@@ -0,0 +1,440 @@
...
+        end_of_stream: bool,
+    ) -> Result<FilterAction, FilterError> {
+        if !end_of_stream {
+            return Ok(FilterAction::Continue);
+        }
+        if !is_responses_request(ctx) {
+            return Ok(FilterAction::Release);
+        }
+        let streaming = is_streaming(ctx);
+        let Some(state) = ctx.extensions.get::<ResponsesState>() else {
```
franciscojavierarceo, line 254: Could you clarify whether auto-compaction is intentionally limited to requests rehydrated through `previous_response_id` or `conversation`? A direct full-input request with `store: false` and `context_management` has no `ResponsesState` here and is silently released. If that limitation is intentional, documenting and scoping it would be helpful; otherwise state likely needs to be initialized for every Responses request.

franciscojavierarceo (changes requested):
Thanks for putting this together—compaction is a substantial addition. I’m requesting changes on a few correctness and API-contract issues that I think should be addressed before merge:

- The generated summary is persisted as a system message rather than canonical Responses compaction state.
- Token estimation and summarization omit supported conversation items such as tool calls, tool outputs, and previous compaction state.
- The documented failure-mode behavior is not currently reachable or applied consistently.
- Missing or malformed thresholds trigger immediate compaction.
- The integration test does not yet demonstrate that compaction actually occurred.

There is also a scope mismatch worth resolving: this PR says `Closes #30`, while #30 explicitly includes the standalone `POST /v1/responses/compact` flow, `summary_prefix`, a canonical `OpenAIResponseCompaction` item, and hidden chainable storage. If those pieces are intentionally deferred or handled through passthrough, I’m happy for this PR to be narrowed and for #30 to remain open; otherwise they should be completed here.

For context, I previously implemented and shipped analogous Responses compaction support in [ogx-ai/ogx#5327](https://github.com/ogx-ai/ogx/pull/5327), including the explicit compact endpoint, automatic `context_management` compaction, canonical compaction items, persistence/chaining, and Codex CLI end-to-end testing. I compared that lifecycle while reviewing this PR. The comments here are grounded in this PR’s linked issue and the Responses API contract rather than a request to reproduce OGX’s implementation exactly.

apis/src/openai/responses/compact/mod.rs
crstrn13: documented for now, as I've decided to narrow down the scope. Will track the support of that path under #597

apis/src/openai/responses/compact/mod.rs
franciscojavierarceo: this handles the tool-call, tool-output, and previous-compaction cases nicely. one piece still seems unaccounted for: `should_compact` estimates only `state.messages`, while tool definitions and instructions also contribute to the rendered context. could we either include those in the estimate or track that limitation explicitly as follow-up work? once that’s scoped, this thread looks good to me.

crstrn13: right, I missed that the last time. It should be handled in the latest commit. Thanks a bunch for pointing it out.

franciscojavierarceo (approved):
looks good now — thanks for addressing the feedback.

Issue #26, opened 2026-06-01
Filter 7: agentic_loop


leseb:
## Purpose

Track the complete agentic inference-loop lifecycle for the OpenAI Responses API.

The `agentic_loop` filter is deliberately a loop controller: it observes completed model output, checks exit conditions, accumulates output and usage, and signals whether iterative request routing should re-enter inference or return to the client. Tool classification and execution belong to their dedicated filters.

## Current architecture

The non-streaming loop is implemented with Praxis `iterative_request_router`:

1. `openai_mcp_tool_resolve` resolves request-supplied MCP definitions.
2. An IRR inference step runs server-side executors, `agentic_loop`, `openai_responses_proxy`, routing, and load balancing.
3. The model response is accumulated by `agentic_loop`.
4. `openai_mcp_dispatch` or `openai_web_search` detects a pending server-side call and publishes `action = loop`.
5. IRR re-enters the inference step; the execution filter appends the tool result before `openai_responses_proxy` serializes the next model request.
6. A model response without pending server-side calls exits to the client.

`max_infer_iters` is the application-level limit and IRR `max_iterations` is the infrastructure safety limit. The latter must be at least as large as the former.

## Implemented

- [x] Loop controller, output accumulation, usage accumulation, and exit conditions (#294, merged in PR #623)
- [x] Praxis IRR request re-entry using the released agentic-loop support
- [x] Non-streaming MCP discovery and model → `tools/call` → model execution
- [x] Non-streaming web-search execution and re-entry
- [x] Client-owned function calls return without proxy execution
- [x] Example configuration and focused integration coverage

## Remaining children

- [ ] #633 — streaming through IRR-based agentic inference
- [ ] #276 — emit client-visible SSE for locally executed tool activity
- [ ] #637 — complete MCP approval request/response round trip
- [ ] #100 — remaining integration and example coverage

## Filter behavior

- Parse a non-streaming Responses model response and collect completed output items.
- If there are no pending server-side calls, set the loop action to done and return the accumulated final response.
- Exit on model-owned incomplete/max-output-token status.
- Return a 508 Responses-compatible error when `max_infer_iters` is exceeded.
- On re-entry, force `parallel_tool_calls = false` and reset `tool_choice` to `auto`.
- Preserve accumulated output items and usage across inference rounds.

## Scope boundaries

- MCP discovery: `openai_mcp_tool_resolve`
- MCP approval and execution: `openai_mcp_dispatch`
[...]

PR #404, opened 2026-07-17
feat(mcp): add openai_mcp_dispatch filter for MCP tool execution

### LINKED ISSUE #27: Epic: OpenAI Responses MCP dispatch
## Parent

Part of #93.

## Purpose

Track the OpenAI Responses MCP execution filter and its remaining contract, lifecycle, performance, and conformance work.

The original Filter 8 was named `mcp_tool`. The implemented filter is now `openai_mcp_dispatch` and runs inside the `iterative_request_router` agentic loop.

## Current architecture

`openai_mcp_dispatch` operates in two phases:

1. On the model response path, identify completed function calls backed by `ResponsesState::mcp_tool_map`, apply approval policy, and decide whether the loop should continue.
2. On the next IRR request-body pass, execute eligible MCP `tools/call` operations through `mcp_client`, append model-facing `function_call_output` plus client-facing `mcp_call` items, and resume inference.

Client-owned function calls do not match the MCP tool map and are returned without proxy execution.

## Implemented

- [x] Match rewritten MCP function names back to their original server/tool entries.
- [x] Parse object and string-valued arguments without double encoding.
- [x] Apply `require_approval` policy and return `mcp_approval_request` output.
- [x] Execute MCP calls sequentially or concurrently inside the owning request future.
- [x] Preserve text, image, and embedded-resource result content losslessly.
- [x] Append complete MCP execution traces across buffered and logical-streaming agentic rounds.
- [x] Cover automatic execution, credential isolation, failures, usage accumulation, and multi-round behavior with unit, integration, and live-SDK tests.

## Remaining sub-issues

- [ ] #1018 — enforce response-wide `max_tool_calls` for MCP dispatch.
- [ ] #1019 — reuse MCP dispatch sessions across agentic rounds.
- [ ] #1020 — pin audio and resource-link result preservation.

## Related work owned elsewhere

- #276 / #985 — synthesize client-visible SSE for locally executed tool activity.
- #637 / #982 — complete approval approve, deny, replay prevention, and resume.
- #990 — accept model-produced batches safely when a backend ignores `parallel_tool_calls=false`.
- #957 — route MCP exchanges through configured filtered outbound chains.

## Scope boundaries

- MCP discovery and tool rewriting: #43 / `openai_mcp_tool_resolve`.
- Agentic-loop extraction, accumulation, and continuation: #26 / `openai_agentic_loop`.
- Event ordering and sequence allocation: #276 / `openai_stream_events`.
- Inbound MCP classifier, broker, and gateway sessions: #148 and its gateway children.

## Reference

- [Proposal: Responses API Filters](https://github.com/praxis-proxy/ai/blob/main/docs/proposals/00354_responses-api-filters.md)

leseb:
## Summary

Add the `openai_mcp_dispatch` filter that executes MCP tool calls against upstream MCP servers within the Responses API agentic loop. The filter checks approval policies (default: always, fail-closed), resolves tool calls from the `mcp_tool_map`, executes `tools/call` via the `mcp_client` module, and appends results to `ResponsesState`.

- Add `call_tool()` to `mcp_client` with defensive argument parsing and SSRF protection
- Implement approval policy engine with ambiguity detection for duplicate tool names
- Add 34 unit tests + 11 integration tests against a real rmcp-based MCP server
- Add standalone example config (`mcp-dispatch.yaml`)

Requires `agentic_loop` (#26) for end-to-end execution; without it the filter validates config at startup but cannot be triggered at runtime.

Closes #27


apis/src/openai/responses/mcp_dispatch/mod.rs
```diff
@@ -0,0 +1,641 @@
...
+    ) -> Result<FilterAction, FilterError> {
+        if !end_of_stream {
+            return Ok(FilterAction::Release);
+        }
+
+        let Some(state) = ctx.extensions.get::<ResponsesState>() else {
+            return Ok(FilterAction::Continue);
+        };
+
+        let mcp_calls = extract_mcp_tool_calls(&state.tool_calls, &state.mcp_tool_map);
```
franciscojavierarceo, line 195: I think this path currently never fires for native MCP responses. `mcp_dispatch` only looks at `ResponsesState::tool_calls`, but the stream accumulator currently adds entries there only for `function_call` events; native `mcp_call` events remain in `output_items`. So when the resolved MCP tools are sent upstream, this sees no pending calls and never issues `tools/call`. We should either accumulate native `mcp_call` events here or rewrite these tools as function tools before relying on this filter.

leseb: Good catch, I need to rewrite mcp to function, but that needs to happen inside openai_mcp_tool_resolve filter. I've opened https://github.com/praxis-proxy/ai/issues/425 and will followup with a new PR to fix this. Thanks!

examples/configs/openai/responses/mcp-dispatch.yaml
```diff
@@ -0,0 +1,45 @@
...
+listeners:
+  - name: ai-gateway
+    address: "127.0.0.1:8080"
+    filter_chains: [mcp-dispatch-pipeline]
+
+filter_chains:
+  - name: mcp-dispatch-pipeline
+    filters:
+      - filter: openai_tool_parse
+      - filter: mcp_tool_resolve
```
aslakknutsen: openai_mcp_tool_resolve?

examples/configs/openai/responses/mcp-dispatch.yaml
```diff
@@ -0,0 +1,45 @@
...
+        timeout_ms: 5000
+      - filter: openai_mcp_dispatch
+        timeout_ms: 30000
+        allow_loopback: false
+      - filter: openai_stream_events
```
aslakknutsen: the comment says the order should be openai_stream_events > openai_mcp_dispatch ?

examples/configs/openai/responses/mcp-dispatch.yaml
```diff
@@ -0,0 +1,45 @@
+# MCP Dispatch Filter
```
aslakknutsen, line 1: Do we have an integration test that run this?

apis/src/openai/responses/mcp_dispatch/mod.rs
```diff
@@ -0,0 +1,641 @@
...
+        if mcp_calls.is_empty() {
+            return Ok(FilterAction::Continue);
+        }
+
+        debug!(count = mcp_calls.len(), "executing pending MCP tool calls");
+
+        let parallel = state.parallel_tool_calls;
+        let tool_map = state.mcp_tool_map.clone();
```
aslakknutsen: execute_mcp_calls also clones the tool_map

apis/src/openai/responses/mcp_dispatch/mod.rs
```diff
@@ -0,0 +1,641 @@
...
+            .iter()
+            .map(|tc| {
+                let tc = tc.clone();
+                let map = std::sync::Arc::clone(&shared_map);
+                tokio::spawn(async move { execute_single_call(&tc, &map, timeout, allow_loopback).await })
+            })
+            .collect();
+        let mut results = Vec::with_capacity(handles.len());
+        for handle in handles {
+            if let Ok(Some(result)) = handle.await {
```
aslakknutsen: "JoinError or Ok(None) results are ignored. on_request then unconditionally retain-filters all MCP entries out of tool_calls (line 176) regardless of whether every call produced a result. Scenario: two parallel tools/call, one task panics or is cancelled -> one result appended, both tool_calls removed -> model never sees output or error for the dropped call and cannot retry. Sequential path has the same retain behavior for None returns."

aslakknutsen (changes requested):

examples/configs/openai/responses/mcp-dispatch.yaml
leseb: yes we spawn a fake mcp server

aslakknutsen: hmm, right.. there is no tests/schema/tests/suite/examples/ == examples/configs setup in the ai repo like in core.  But still a req in AGENTS.md https://github.com/praxis-proxy/ai/blob/main/AGENTS.md#test-requirements

leseb: will add integration test to load the filter too

examples/configs/openai/responses/mcp-dispatch.yaml
leseb: duh yes :)

apis/src/openai/responses/mcp_dispatch/mod.rs
leseb: https://github.com/praxis-proxy/ai/pull/404/commits/b09463dc0283f207d69636bf7b2517aa1bf5bd9b

apis/src/openai/responses/mcp_dispatch/mod.rs
leseb: https://github.com/praxis-proxy/ai/pull/404/commits/b09463dc0283f207d69636bf7b2517aa1bf5bd9b

examples/configs/openai/responses/mcp-dispatch.yaml
leseb: https://github.com/praxis-proxy/ai/pull/404/commits/b09463dc0283f207d69636bf7b2517aa1bf5bd9b

examples/configs/openai/responses/mcp-dispatch.yaml
leseb: https://github.com/praxis-proxy/ai/pull/404/commits/b09463dc0283f207d69636bf7b2517aa1bf5bd9b

examples/configs/openai/responses/mcp-dispatch.yaml
leseb: https://github.com/praxis-proxy/ai/pull/404/commits/b09463dc0283f207d69636bf7b2517aa1bf5bd9b

apis/src/openai/responses/mcp_dispatch/mod.rs
```diff
@@ -0,0 +1,694 @@
...
+            "output": output_text,
+        })
+    };
+
+    McpCallResult { message, output_item }
+}
+
+/// Build an error result for a tool call that was dropped
+/// (task panic, cancellation, or missing fields).
+fn error_result_for_dropped_call(tool_call: &serde_json::Value, reason: &str) -> Option<McpCallResult> {
```
aslakknutsen: "When resolve_tool_entry returns Err(None) for an unknown tool, execute_single_call returns None. error_result_for_dropped_call also returns None if the tool call lacks call_id and id fields. execute_mcp_calls then omits a result, yet on_request unconditionally retain-filters all MCP entries out of tool_calls (mod.rs:176). The model receives neither output nor error and cannot retry. Guard error_result_for_dropped_call to synthesize a call_id (as extract_call_id already defaults to "unknown") or retain only tool_calls that produced a result."

apis/src/openai/responses/mcp_dispatch/tests.rs
```diff
@@ -0,0 +1,912 @@
...
+    let map = std::sync::Arc::new(sample_tool_map());
+    let calls = vec![json!({"name": "get_weather", "call_id": "c1", "arguments": {}})];
+    let timeout = std::time::Duration::from_millis(200);
+    let results = execute_mcp_calls(&calls, &map, true, timeout, true).await;
+    assert_eq!(results.len(), 1);
+    assert_eq!(results[0].output_item["type"], "mcp_call");
+}
+
+#[tokio::test]
+async fn execute_mcp_calls_emits_error_for_unknown_tools() {
```
aslakknutsen, line 764: A regression test for {"name":"nonexistent"} without id fields would catch bug-silent-drop-missing-call-id.

apis/src/openai/responses/mcp_dispatch/tests.rs
```diff
@@ -0,0 +1,912 @@
...
+    let map = std::sync::Arc::new(sample_tool_map());
+    let calls = vec![json!({"name": "get_weather", "call_id": "c1", "arguments": {}})];
+    let timeout = std::time::Duration::from_millis(200);
+    let results = execute_mcp_calls(&calls, &map, false, timeout, true).await;
+    assert_eq!(results.len(), 1);
+    assert_eq!(results[0].output_item["type"], "mcp_call");
+}
+
+#[tokio::test]
+async fn execute_mcp_calls_parallel() {
```
aslakknutsen, line 754: No unit test spawns a failing tokio task or simulates JoinError to assert one error result per dropped call and that retain does not outpace results.

apis/src/openai/responses/mcp_dispatch/mod.rs
```diff
@@ -0,0 +1,694 @@
...
+        let mcp_calls = extract_mcp_tool_calls(&state.tool_calls, &state.mcp_tool_map);
+        if mcp_calls.is_empty() {
+            return Ok(FilterAction::Continue);
+        }
+
+        if let Some(pending) = find_approval_required(&mcp_calls, &state.mcp_tool_map) {
+            return Ok(Self::handle_approval_required(ctx, &pending));
+        }
+
+        ctx.set_metadata("openai_mcp_dispatch.action".to_owned(), "execute_mcp".to_owned());
```
aslakknutsen, line 204: Who's reading this?

aslakknutsen (changes requested):

apis/src/openai/responses/mcp_dispatch/mod.rs
```diff
@@ -0,0 +1,694 @@
...
+
+        let mcp_calls = extract_mcp_tool_calls(&state.tool_calls, &state.mcp_tool_map);
+        if mcp_calls.is_empty() {
+            return Ok(FilterAction::Continue);
+        }
+
+        debug!(count = mcp_calls.len(), "executing pending MCP tool calls");
+
+        let parallel = state.parallel_tool_calls;
+        let tool_map = std::sync::Arc::new(state.mcp_tool_map.clone());
```
aslakknutsen, line 162: "For large tool maps or hot agentic loops this duplicates HashMap data that could be borrowed or shared via existing ResponsesState ownership boundaries. Consider Arc in ResponsesState at resolve time or pass a shared reference when parallel is false."

I'm not sure how hot path or how large the map realistically(or in an evil scenario) could get if the comment if worth looking into.

apis/src/openai/responses/mcp_dispatch/tests.rs
leseb: I think this would only test tokio's internal, plus it's hard to do. won't do

apis/src/openai/responses/mcp_dispatch/mod.rs
leseb: agentic_loop will #26

apis/src/openai/responses/mcp_dispatch/tests.rs
leseb: 7a9a143c4453f55fd0bdb9e5dc9d06e07b64a74b

apis/src/openai/responses/mcp_dispatch/mod.rs
leseb: 7a9a143c4453f55fd0bdb9e5dc9d06e07b64a74b

apis/src/openai/responses/mcp_dispatch/mod.rs
leseb: a4aa782a

aslakknutsen (approved):

PR #429, opened 2026-07-20
fix(openai): preserve first-pass Responses requests

### LINKED ISSUE #426: fix(openai): responses_proxy rebuilds request body unconditionally and strips previous_response_id prematurely
## Bug Description

`openai_responses_proxy` unconditionally rebuilds the outbound request body from `ResponsesState::messages` whenever a `ResponsesState` extension is present, even when rehydration did not modify the conversation history. It also strips `previous_response_id` regardless of whether the rehydrate filter actually resolved it.

This causes two problems:

1. **Body corruption**: downstream filters that modified the request body (e.g. model rewrite, prompt enrichment) have their changes silently discarded because the proxy serializes from the original parsed state rather than the current body.
2. **Premature field stripping**: `previous_response_id` is removed before the upstream provider can see it in cases where Praxis did not resolve it locally (e.g. when the provider owns the response store).

## Expected Behavior

- The proxy should only replace `input` in the outbound body when `ResponsesState` indicates the conversation history actually changed (rehydration injected items).
- `previous_response_id` should only be stripped after the rehydrate filter has resolved it locally.
- When no rebuild is needed, the proxy should preserve the current request body as-is (apart from removing the Praxis-owned `conversation` field).

## Actual Behavior

Every request with a `ResponsesState` extension gets its body re-serialized from the parsed state, discarding any downstream body rewrites. `previous_response_id` is always stripped.

## Affected Code

- `apis/src/openai/responses/openai_responses_proxy/mod.rs` — `on_request_body`, `serialize_body`, `rebuild_outbound_body`
- `apis/src/openai/responses/rehydrate/mod.rs`
- `apis/src/openai/responses/state.rs`

leseb:
## Summary

- Rebuild request JSON only after local history changes it, preserving downstream body rewrites from other filters (model rewrite, prompt enrichment)
- Strip `previous_response_id` only when the rehydrate filter has actually resolved it locally
- Add `request_needs_rebuild` check to `ResponsesState` so the proxy can skip unnecessary serialization

Closes #426

### DIFF
--- a/apis/src/openai/responses/openai_responses_proxy/mod.rs
+++ b/apis/src/openai/responses/openai_responses_proxy/mod.rs
@@ -9,11 +9,11 @@
 //! Named `inference` in pipeline configs so branch chains can
 //! `rejoin` here for the agentic tool loop.
 //!
-//! When `ResponsesState` is present in `RequestExtensions`,
-//! rebuilds the request body with the full conversation history
-//! from `state.messages` and strips `previous_response_id` (already
-//! resolved by the rehydrate filter). When no state is present,
-//! passes the body through unchanged.
+//! When `ResponsesState` is present in `RequestExtensions`, replaces
+//! the request input with `state.messages` only after conversation
+//! history has changed it. It strips `previous_response_id` only after
+//! the rehydrate filter has resolved it locally. Every path removes the
+//! Praxis-owned `conversation` field before forwarding upstream.
 
 mod config;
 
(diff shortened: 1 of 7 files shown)


apis/src/openai/responses/openai_responses_proxy/mod.rs
```diff
@@ -51,12 +51,12 @@ use super::{error::responses_error_rejection, state::ResponsesState};
 ///
 /// Reads the assembled conversation history from
 /// `ResponsesState::messages` and replaces the `input` field in
-/// the outbound body. Strips `previous_response_id` since Praxis
-/// already resolved it locally via the rehydrate filter.
+/// the outbound body when it differs from the original normalized
+/// input. Strips `previous_response_id` after Praxis resolves it
+/// locally via the rehydrate filter.
```
aslakknutsen, line 57: "write_tool_map still documents skipping ResponsesState creation for continuation requests to avoid openai_responses_proxy stripping previous_response_id. That rationale is obsolete now that stripping is gated on history_rehydrated. Not a functional bug but may mislead future edits."

https://github.com/praxis-proxy/ai/blob/0de807e8fd3dea94fd3448d4bd3b4f13336d85cb/apis/src/openai/responses/openai_mcp_tool_resolve/mod.rs#L335

aslakknutsen (commented):
"

apis/src/openai/responses/openai_responses_proxy/mod.rs
leseb: makes sense, good doc is important

leseb: https://github.com/praxis-proxy/ai/pull/429/commits/cd3b9ec2f3b45fdc1b5c7008a5871c4dea1509b1

aslakknutsen (approved):

PR #449, opened 2026-07-20
Keep Anthropic tool-error replay compatible

franciscojavierarceo:
## Summary

- keep the native Claude replay fixture on its original typed bash tool
- derive an untyped custom tool only for the Anthropic-to-OpenAI translation test
- assert both native fidelity and translation compatibility

## Why

The translation replay reused a typed Anthropic built-in tool and expected HTTP 200. Translation compatibility validation rejects typed built-ins because Chat Completions cannot preserve their semantics, so the test conflicted with that policy.

## Impact

Native Anthropic replay coverage remains faithful to the captured fixture, while the translation test isolates tool_result.is_error behavior without depending on unsupported built-in-tool semantics.

## Validation

- cargo +nightly fmt --all -- --check
- cargo test -p praxis-tests-integration replay_claude_messages_tool_error_translates_error_marker_for_openai -- --nocapture
- cargo test -p praxis-tests-integration session_replay -- --nocapture
- cargo test -p praxis-ai-apis anthropic::to_openai::request -- --nocapture
- make lint
- independent read-only Claude Code review: no actionable findings


tests/integration/tests/suite/examples/session_replay.rs
```diff
@@ -416,6 +427,14 @@ fn replay_claude_messages_tool_error_translates_error_marker_for_openai() {
         .as_array()
         .expect("OpenAI request should contain messages");
 
+    assert!(
```
aslakknutsen, line 430: nit: The test verifies translation_request input shape and forwarded message content, but never checks forwarded["tools"] contains the converted function definition. With the pre-change payload the suite still returned HTTP 200 because convert_tool_definition silently drops bash_* tools rather than rejecting the request. Adding assert_eq!(forwarded["tools"][0]["function"]["name"], "bash") (or similar) would prove the custom-tool substitution actually affects the backend payload, not just the client-side setup.

leseb (approved):

aslakknutsen (approved):

tests/integration/tests/suite/examples/session_replay.rs
franciscojavierarceo: addressed

PR #396, opened 2026-07-16
docs(proposals): add What?/Why? for A2A task route capture parse comp…

mkoushni:
## Summary

`try_capture_from_buffer` in `filters/src/agentic/a2a/mod.rs` (the
non-streaming A2A task-route capture path) hex-decodes and attempts a
full `serde_json::from_slice` parse over the **entire accumulated
response buffer** on every `on_response_body` chunk, discarding the
result whenever the JSON is still incomplete. For a body delivered in
`k` chunks, this repeats full decode+parse work over an ever-growing
buffer on every chunk, producing O(n²) total work in response body
bytes rather than O(n) — see #353.

This PR adds the proposal's `What?`/`Why?` sections per the
`docs/proposals/template.md` convention. Per the template note, the
`How?` section is intentionally omitted here and will follow in a
separate PR once the proposal direction is agreed.

## What's in this PR

- New proposal: `docs/proposals/00353_a2a-task-route-capture-parse-complexity.md`
- Describes bounding the per-chunk capture cost to O(chunk) via an
  incremental, string-aware brace/bracket depth scan, so the
  expensive hex-decode + JSON parse is only attempted once the buffer
  plausibly holds a complete, balanced JSON value.
- Explicitly scoped to the **non-streaming** capture path only — the
  SSE capture path (`process_sse_response_chunk`) already scans
  incrementally and is unaffected.
- Preserves the existing opportunistic, pre-EOS capture requirement
  (a route must be captured as soon as the JSON body is complete,
  without waiting for `end_of_stream`, since Pingora may not deliver
  a separate EOS callback after the final data chunk).
- Non-goals: no change to the hex-encoding scheme used for multibyte
  UTF-8 chunk-boundary safety, no change to the SSE scanner, and no
  overlap with #352 (handled separately).

## Test plan

- [x] `make lint` passes (clippy, fmt-check, separator/filter-docs lints)
- [x] No code changes in this PR — documentation only

Relates to #353

PR #450, opened 2026-07-21
perf(a2a): bound task route capture parsing to O(n) with a balance scan

### LINKED ISSUE #353: perf(a2a): O(n²) hex-decode and JSON parse on every response body chunk
`try_capture_from_buffer` in `filters/src/agentic/a2a/mod.rs` is called
on every `on_response_body` chunk. Each invocation hex-decodes the entire
accumulated buffer and attempts a full `serde_json::from_slice` parse,
producing O(n²) work in the number of chunks.

Note: parsing only on `end_of_stream` is not viable due to a Pingora
constraint (may not deliver a separate EOS callback after the final data
chunk). A heuristic approach (e.g. check for trailing `}`) or running
depth counter would avoid the quadratic behavior.

mkoushni:
## Summary

`try_capture_from_buffer` in `filters/src/agentic/a2a/mod.rs` (the
non-streaming A2A task-route capture path) runs on every
`on_response_body` chunk while capture is enabled. Each invocation
hex-decodes the **entire accumulated buffer** and attempts a full
`serde_json::from_slice` parse over it, discarding the result whenever
the JSON is still incomplete. For a body delivered in `k` chunks, this
repeats full decode+parse work over an ever-growing buffer on every
chunk, producing O(n²) total work in response body bytes rather than
O(n).

Parsing only at `end_of_stream` is not viable: Pingora may not deliver
a separate EOS callback after the final data chunk, so the route must
be captured opportunistically as soon as the JSON body is complete.

## Fix

Add a `JsonBalanceState` scanner that tracks `{}`/`[]` depth
incrementally, skipping bytes inside string literals (respecting
backslash escapes). It is updated only with the bytes in each new
chunk and persisted via `filter_metadata` between calls, so total scan
work across all chunks is O(n) rather than O(buffer²):

```rust
fn scan_json_balance(state: &mut JsonBalanceState, chunk: &[u8]) {
    if state.is_complete() {
        return;
    }
    for &byte in chunk {
        // track in_string / escaped / depth incrementally
        // ...
    }
}
```

The expensive hex-decode + JSON parse in `try_capture_from_buffer` is
now only attempted once the scanner reports the buffer holds a
balanced top-level value, or at `end_of_stream` as before:

```rust
if balance.is_complete() || end_of_stream {
    try_capture_from_buffer(ctx, store, &self.config.task_routing, end_of_stream);
}
```

This preserves the existing opportunistic, pre-EOS capture guarantee
while bounding total per-response scan work to O(n). `clear_capture_metadata`
is extended to also clear the new scanner scratch keys, and reused in
`accumulate_response_hex`'s overflow path to remove the previous
duplicated inline `.remove()` calls.

Scoped to the non-streaming capture path only — the SSE capture path
(`process_sse_response_chunk`) already scans incrementally and is
unaffected. No change to the hex-encoding scheme used for multibyte
UTF-8 chunk-boundary safety.

Implements the proposal in #396.

## Test plan

- [x] Unit tests for `JsonBalanceState`/`scan_json_balance`: minimal
  object, no-object-seen, nested objects/arrays, structural bytes
  inside strings, escaped quotes inside strings, no-op once complete,
  byte-by-byte accumulation matching a single-shot scan
- [x] New regression test `many_single_byte_chunks_still_capture_route_before_eos`
  covering capture before `end_of_stream` when a body arrives as many
  single-byte chunks
- [x] Existing A2A response-body unit test coverage passes unmodified
- [x] `make test` — full workspace suite passes (0 failed)
- [x] `make lint` — clippy, nightly fmt-check, separator/filter-docs
  lints all pass

Fixes #353


mkoushni:
@aslakknutsen the doc proposal for this fix is https://github.com/praxis-proxy/ai/pull/396

filters/src/agentic/a2a/mod.rs
```diff
@@ -511,6 +523,120 @@ fn accumulate_response_hex(ctx: &mut HttpFilterContext<'_>, chunk: &[u8], max_by
...
+        if state.in_string {
+            match byte {
+                b'\\' => state.escaped = true,
+                b'"' => state.in_string = false,
+                _ => {},
+            }
+            continue;
+        }
+
+        match byte {
```
aslakknutsen, line 641: {} and [] are not interchangeable and Depth alone does not care which closer matches which opener leading to is_complete == true

filters/src/agentic/a2a/mod.rs
```diff
@@ -289,7 +289,15 @@ impl HttpFilter for A2aFilter {
...
             }
 
-            try_capture_from_buffer(ctx, store, &self.config.task_routing, end_of_stream);
+            let mut balance = load_json_balance_state(ctx);
+            if let Some(chunk) = body.as_ref() {
+                scan_json_balance(&mut balance, chunk);
+            }
+            save_json_balance_state(ctx, balance);
+
+            if balance.is_complete() || end_of_stream {
```
aslakknutsen: After the scanner says “JSON looks complete,” a failed real parse leaves capture armed. Every later chunk still does a full hex-decode + `serde_json` of the whole buffer — the cost pattern #353 was meant to kill.

**Happy path (fine):**
1. Chunks arrive → scan updates depth  
2. Depth hits 0 → `is_complete`  
3. `try_capture_from_buffer` → parse Ok → `clear_capture_metadata`  
4. Later chunks: capture off → cheap

**Attack / sticky path:**

1. Backend sends something the **scanner** treats as complete but **serde** rejects, e.g. `{]` or a full object plus `trailing` in the same buffer.  
2. `is_complete == true` → enter `try_capture_from_buffer` (```298:300:filters/src/agentic/a2a/mod.rs```).  
3. `from_slice` fails → neither branch clears (```415:422:filters/src/agentic/a2a/mod.rs```): clear only on Ok or EOS.  
4. Scanner is a no-op once complete, so depth stays “complete.”  
5. Next chunk: still `is_complete` → **full decode+parse again** on a larger hex buffer.  
6. Repeat until body hits `max_response_body_bytes` or EOS.

**Concrete exploit:** Send `{]` (2 bytes), then ~64KiB as 1-byte chunks. First step is cheap; then each of ~65k chunks re-parses an ever-growing buffer → Θ(n·k) CPU on the proxy, inside the configured cap.

**Why “sticky”:** `is_complete` never resets on parse failure; the gate keeps calling the expensive path.

**Fix shape:** First time complete + parse Err → clear/abandon (or one-shot flag). Do not re-run full parse on every later chunk while complete-but-failed.

aslakknutsen (changes requested):

filters/src/agentic/a2a/mod.rs
```diff
@@ -409,6 +444,9 @@ fn try_capture_from_buffer(
         clear_capture_metadata(ctx);
     } else if end_of_stream {
         clear_capture_metadata(ctx);
+    } else {
+        ctx.filter_metadata
```
aslakknutsen: Nit'ish

"After capture_gate_exhausted latches, capture stays armed: later chunks still hex-accumulate into metadata (up to max_response_body_bytes) and skip mid-stream parse, with one final decode+parse at EOS on the full buffer. That preserves the O(n) CPU bound and avoids wrong route stores, but holds ~2× body-size metadata and hex-append work for the rest of the response instead of freeing on the first failed parse. This is bounded by the existing max_response_body_bytes cap, but can be avoided when parse failed. "

Calling clear_capture_metadata(ctx) here instead should avoid the hex buffer accumulating junk as well as the final EOS full buffer decode+parse that has already failed to parse the content.

filters/src/agentic/a2a/mod.rs
```diff
@@ -289,7 +289,16 @@ impl HttpFilter for A2aFilter {
                 return Ok(FilterAction::Continue);
             }
 
-            try_capture_from_buffer(ctx, store, &self.config.task_routing, end_of_stream);
+            let mut balance = load_json_balance_state(ctx);
```
aslakknutsen: When JsonBalanceState sets invalid (e.g. {]), is_complete stays false, so try_capture_from_buffer (and thus clear_capture_metadata) only runs at end_of_stream. Until then capture_enabled remains set and every chunk still hex-accumulates up to max_response_body_bytes—bounded memory/append cost, no per-chunk re-parse. Fix: clear (or stop accumulating) as soon as invalid becomes true.

aslakknutsen (approved):

filters/src/agentic/a2a/mod.rs
```diff
@@ -387,29 +381,75 @@ fn lookup_task_route(
...
+        return;
+    }
+
+    let mut balance = load_json_balance_state(ctx);
+    if let Some(chunk) = body.as_ref() {
+        scan_json_balance(&mut balance, chunk);
+    }
+    let is_complete = balance.is_complete();
+    let is_invalid = balance.invalid;
+    save_json_balance_state(ctx, balance);
```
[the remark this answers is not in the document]
mkoushni: nope

filters/src/agentic/a2a/mod.rs
```diff
@@ -511,6 +554,168 @@ fn accumulate_response_hex(ctx: &mut HttpFilterContext<'_>, chunk: &[u8], max_by
...
+            .is_some_and(|v| v == "true"),
+    }
+}
+
+/// Persists [`JsonBalanceState`] back to `filter_metadata` for the next chunk.
+fn save_json_balance_state(ctx: &mut HttpFilterContext<'_>, state: JsonBalanceState) {
+    ctx.filter_metadata.insert(
+        "a2a.response.json_stack".to_owned(),
+        // Stack bytes are always `}`/`]`, so this is always valid UTF-8.
+        String::from_utf8(state.stack).unwrap_or_default(),
```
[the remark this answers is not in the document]
mkoushni: no needs

PR #453, opened 2026-07-21
feat(apis): convert Responses input_file to input_text for vLLM

### LINKED ISSUE #397: Support Responses input_file documents for vLLM backends
## Purpose

Make OpenAI Responses `input_file` documents usable when the inference backend is vLLM or llm-d.

## Context

The `openai_file_resolve` work in #37 can turn a Praxis-owned `file_id` into standard inline file content. #452 extends the same transport-resolution boundary to remote `file_url` sources. OpenAI backends natively understand Responses `input_file` content, but vLLM does not currently process `input_file` documents. Merely replacing a reference with `file_data` or a data URL therefore does not make the document usable by vLLM.

This is an explicitly configured backend-adaptation step, not generic OpenAI request normalization. Requests routed to backends that natively support `input_file` must remain unchanged.

## Verified behavior (2026-07-17)

We tested the complete path with an OpenAI-compatible OGX Files API, Praxis `openai_file_resolve`, and vLLM 0.16.0 running `Qwen/Qwen3-0.6B`:

1. OGX returned the canonical OpenAI File object from `GET /v1/files/{file_id}` and the raw 1,004-byte body from `GET /v1/files/{file_id}/content`. As in the OpenAI contract, the metadata did not contain `content_type`.
2. `openai_file_resolve` fetched both responses, inferred `text/plain` from the `.txt` filename, removed `file_id`, and produced inline base64 `file_data`.
3. vLLM rejected the rewritten request with HTTP 400: `Unknown part type: input_file`.
4. Sending the same inline `input_file` and `file_data` directly to vLLM produced the same error. This isolates the failure to vLLM document-part handling rather than Files API lookup or Praxis rewriting.
5. A plain-text Responses request through the same Praxis-to-vLLM route succeeded.
6. Fetching the OGX file content, decoding it, and sending it as `input_text` succeeded; the model consumed and identified the document content.

This is not fixed by upgrading vLLM alone. The current vLLM v0.25.1 parser still has no `input_file` handling branch. The exact upstream feature request is open with no linked implementation PR or milestone:

- vLLM feature request: https://github.com/vllm-project/vllm/issues/46562
- Current parser: https://github.com/vllm-project/vllm/blob/v0.25.1/vllm/entrypoints/chat_utils.py#L1647-L1719
- Related Harmony PR #47052 only changes unsupported file parts from being silently ignored to being rejected; it does not implement document ingestion: https://github.com/vllm-project/vllm/pull/47052

The verified workaround is therefore document extraction followed by an explicit `input_file` to `input_text` adaptation. Images remain a separate path through `input_image` and require a vision-capable model. llm-d behavior still needs independent verification.

## Scope

- Define the vLLM-compatible representation for document inputs, initially covering the file types we explicitly support.
- Convert materialized Responses `input_file` content into modalities vLLM can consume, such as:
  - extracted document text as `input_text`
  - rendered document pages as `input_image` where appropriate
- Handle materialized content from all supported sources:
  - `file_id` resolved to inline content by #37
  - client-provided inline `file_data`
  - `file_url` materialized as inline content by #452
- Handle files in both message content arrays and tool-output content arrays without creating synthetic conversation turns.
- Preserve filenames and useful source metadata where the downstream representation permits it.
- Keep the rewritten request body and `ResponsesState` coherent so replay and `responses_proxy` cannot restore stale `input_file` items.
- Bound decoded content, page counts, rendered image sizes, and total rewritten request size.
- Make unsupported file types and conversion failures explicit and configurable rather than silently dropping content.
- Apply the conversion only for configured vLLM/llm-d routes; preserve pass-through fidelity for OpenAI-compatible backends with native document support.

## Design direction

- Keep `openai_file_resolve` standards-preserving: it resolves `file_id` and configured `file_url` references to standard inline content but does not change document semantics.
- Keep document extraction network-free. It consumes inline content produced by the client or `openai_file_resolve`.
- Add document extraction as a separate, explicitly configured backend adapter.
- Decode safe textual formats such as `text/*`, JSON, and CSV directly into `input_text`.
- Use a dedicated extraction service/library for PDF, DOCX, PPTX, and other structured document formats rather than embedding a broad document-processing stack into the file resolver.
- Treat rendered document pages as an opt-in vision path, not the default document representation.

## Dependencies

- #37 is required for Praxis-owned `file_id` inputs.
- #452 is required to materialize remote `file_url` inputs.
- Client-provided inline `file_data` can use the adapter independently.

## Non-goals

- Reimplementing the Files API or `file_id` lookup.
- Downloading or dereferencing `file_url`; that transport-resolution work belongs to #452.
- Rewriting valid `file_data`, `file_url`, or `input_file` inputs for providers that already support them.
- Enabling this translation implicitly in the default stateless pass-through path.

## Acceptance criteria

- A Responses request containing materialized `input_file` content reaches a vLLM-compatible backend as supported text and/or image input, with the document content actually available to inference.
- Coverage includes client-provided inline `file_data`, `file_id` resolved by #37, and `file_url` materialized by #452.
- Rehydrated history and tool-output file inputs follow the same conversion path.
- Resource limits and failure behavior have unit tests.
- A functional integration test exercises the conversion and downstream-consumable request shape; an echo-only passthrough test is insufficient.
- At least one functional test proves extracted text is consumed by an actual vLLM-compatible inference path, or an equivalent parser-level test proves the exact downstream representation.
- The composed path `file_url -> file_data -> input_text` is covered once #452 is available.
- An example config documents how to enable the vLLM adapter, supported formats, extraction dependency, and limits.

## Parent

- Parent: #22
- Related: #37, #70, #93, #164, #452

leseb:
## Summary

- Add `openai_doc_extract` filter that converts `input_file` content parts to `input_text` for inference backends that do not natively support `input_file` (e.g. vLLM, llm-d)
- Text-safe content (`text/*`, `application/json`, `application/xml`) is decoded from base64, validated as UTF-8, and forwarded as plain text with an optional `[Source: filename]` prefix
- The filter runs after `openai_file_resolve` and before `openai_responses_proxy` — it does not perform network I/O; parts without inline `file_data` (`file_url`, unresolved `file_id`) are skipped
- Enforces per-file content limit capped at the OpenAI `input_text.text` schema maximum (10,485,760 bytes), per-request file reference count, and aggregate text byte budgets
- Includes 56 unit tests, 9 integration tests with exact JSON equality assertions, and a live vLLM inference test (`test_doc_extract_inline_file_to_input_text`) that asserts a document marker appears in actual model output

## Test plan

- [x] `make build` — workspace compiles
- [x] `make lint` — clippy + nightly fmt + separator + filter docs checks pass
- [x] `make test` — 3,016 tests pass, 0 failures
- [ ] Live vLLM CI — `test_doc_extract_inline_file_to_input_text` sends `input_file` with inline `file_data` through the full proxy pipeline and asserts the model's output contains the document marker

Closes #397


apis/src/openai/responses/doc_extract/extract.rs
```diff
@@ -0,0 +1,302 @@
...
+
+    let Some(data) = file_data else {
+        return Ok(None);
+    };
+
+    budget.register_reference()?;
+
+    let (content_bytes, mime) = extract_from_file_data(data, filename, budget)?;
+
+    if !is_text_safe_mime(&mime) {
```
aslakknutsen, line 182: Split up extract_from_file_data  a bit so we can call is_text_safe_mime before decode to avoid the cost for unsupported mimes?

apis/src/openai/responses/doc_extract/tests.rs
```diff
@@ -0,0 +1,1274 @@
...
+        "should contain decoded raw base64 text: {text}"
+    );
+    assert!(
+        text.contains("[Source: readme.txt]"),
+        "should have filename prefix: {text}"
+    );
+}
+
+#[tokio::test]
+async fn unsupported_format_continue_leaves_input_file() {
```
aslakknutsen, line 459: add a unit test that a non-text data URI returns Ok(None) without allocating decoded content when is_text_safe_mime reorder comment above is fixed.

aslakknutsen (changes requested):

aslakknutsen:
This PR only partially satisfies the linked issue; create a follow up on the parent with regards to PDF/DOCX/PPTX ++ support?

leseb:
> This PR only partially satisfies the linked issue; create a follow up on the parent with regards to PDF/DOCX/PPTX ++ support?

You're right; other formats can be tricky and would grow the PR quite a lot so i've left them out. I'll register a new issue for those, thanks!

apis/src/openai/responses/doc_extract/tests.rs
leseb: https://github.com/praxis-proxy/ai/pull/453/commits/ba139d7a0f3de9dea5278328c5e760ed81d3a205

apis/src/openai/responses/doc_extract/extract.rs
leseb: https://github.com/praxis-proxy/ai/pull/453/commits/ba139d7a0f3de9dea5278328c5e760ed81d3a205

aslakknutsen (approved):

leseb:
> > This PR only partially satisfies the linked issue; create a follow up on the parent with regards to PDF/DOCX/PPTX ++ support?
> 
> You're right; other formats can be tricky and would grow the PR quite a lot so i've left them out. I'll register a new issue for those, thanks!

@aslakknutsen https://github.com/praxis-proxy/ai/issues/455

Issue #454, opened 2026-07-21
refactor(apis): upgrade praxis-core for bounded callout reads


leseb:
## Summary

`ApiClient::get_bytes` routes content downloads through `CalloutClient` for
uniform circuit breaking and callout-depth protection (#405). Published
`praxis-proxy-core` 0.4.0 collects the full response with
`response.bytes().await` before `ApiClient` can check `max_bytes`, so an
oversized or chunked upstream response is fully buffered in memory before
rejection.

## What's needed

Praxis-core main commit `praxis-proxy/praxis@80d5589` adds
`CalloutConfig::max_response_bytes` and enforces that client-wide limit during
chunk-by-chunk response collection.

Once a release containing that commit is published:

1. Bump `praxis-proxy-core` to the new version.
2. Set `CalloutConfig::max_response_bytes` to the filter's `max_body_bytes`.
3. Preserve response-size failures as `ResponseTooLarge` so
   `openai_file_resolve` continues returning 413 rather than 502.
4. Update the `get_bytes` documentation to distinguish the client-wide
   collection limit from the tighter per-request `max_bytes`, which remains
   post-collection unless core adds per-call bounded execution.
5. Add tests proving:
   - chunked responses stop at the client-wide limit;
   - responses below the configured limit can exceed 1 MiB;
   - size-limit failures remain 413;
   - ordinary callout failures remain 502.

## Current mitigation

`get_bytes` checks `max_bytes` after collection (post-buffer). This prevents
returning oversized content to callers but does not bound memory consumption
during the download.

## Blocked by

- praxis-proxy/praxis release containing commit `80d5589`

leseb:
CalloutClient is going away in https://github.com/praxis-proxy/praxis/pull/849 so closing.

Issue #312, opened 2026-07-08
file_search: core filter implementation


leseb:
## Parent

Part of #29

## Scope

Implement the core `file_search` filter that executes file_search tool calls against vector store APIs.

### Included
- Filter module with `HttpFilter` implementation (response-phase)
- Config struct: `vector_store_url`, `api_key`, `auth_type`, templates, `timeout_ms`
- `CalloutClient` from `praxis-core` for HTTP calls with circuit breaking
- Parallel fan-out across multiple vector stores
- Simple `{variable}` template interpolation for search/annotation/context formatting
- Citation extraction: `<|file-xxx|>` markers → annotation structs
- `citation_files` tracking (file_id → filename)
- Unit tests + integration tests + example config

### Not included
- SSE event emission (`file_search_call.in_progress`, `.searching`, `.completed`) — see sibling issue

leseb:
blocked by https://github.com/praxis-proxy/ai/issues/405

leseb:
closed by https://github.com/praxis-proxy/ai/pull/493

PR #456, opened 2026-07-21
refactor(openai): extract shared API client for callout filters

### LINKED ISSUE #405: refactor(openai): add hardened shared API client for OGX callouts
## Purpose

Add a shared, hardened `openai::api_client::ApiClient` for filters that call OGX or other OpenAI-compatible API endpoints.

The goal is to implement common outbound-call safety and HTTP patterns once instead of repeating them in every filter. `ApiClient` is an adapter over the core callout transport, not a domain client: each consuming filter continues to own its endpoint paths, wire types, limits, state mutations, and client-facing behavior.

## `ApiClient` Responsibilities

- Validate a fixed base URL against the shared SSRF-sensitive target policy.
- Encode resource IDs as opaque URL path segments and reject dot segments.
- Validate, normalize, deduplicate, and apply an explicit forwarded-header allowlist.
- Reject transport and Praxis-internal headers from configured forwarding.
- Provide reusable `get_json`, `post_json`, and bounded `get_bytes` operations.
- Disable redirects and environment proxy use for outbound API calls.
- Normalize transport, HTTP status, response-size, URL-construction, and JSON-decode failures for domain clients to map.
- Retain one client instance per consuming filter so circuit-breaker state is not shared across unrelated filters.

`CalloutClient` remains the underlying transport for ordinary JSON callouts. Published `praxis-proxy-core` 0.4.0 cannot enforce a response limit while collecting content downloads, so `get_bytes` temporarily uses a direct `reqwest::Client` with a full-transfer timeout and chunk-by-chunk bounds. Removing that temporary transport split is tracked separately in #454.

## Domain Ownership

Each filter or domain client remains responsible for:

- API routes and provider wire models.
- Search fan-out, ranking, aggregation, and citation behavior.
- File metadata, MIME handling, base64 expansion, and aggregate body budgets.
- `ResponsesState` synchronization, continuation, and replay semantics.
- Fail-open/fail-closed policy and downstream rejection formatting.

## Non-Goals

- Implementing or migrating vector-store search (#312).
- Adding Files API or vector-store domain behavior to `ApiClient`.
- Completing the core `CalloutClient` hardening tracked in `praxis-proxy/praxis#826`.
- Removing the temporary bounded-download transport before core exposes the required API (#454).
- Introducing a shared public YAML schema solely for deduplication.

## Acceptance Criteria

- `apis/src/openai/api_client/` owns the shared outbound API hardening and HTTP helpers.
- `FilesApiClient` composes `ApiClient` rather than duplicating URL, header, transport, and error-handling helpers.
- `ApiClient` contains no Files API, vector-store, Responses-state, or other domain-specific behavior.
- Base-URL validation covers malformed URLs, schemes, credentials, queries/fragments, private and local targets, mapped/legacy IP forms, and explicit private-target opt-in.
- Resource IDs cannot escape their URL path segment.
- Configured forwarded headers are normalized and cannot include hop-by-hop or Praxis-internal headers.
- Byte downloads reject redirects and enforce their per-call limit during response collection.
- JSON request/response behavior and error mapping have focused tests.
- Existing file-resolution behavior remains covered after migration.
- The temporary `get_bytes` transport split and its follow-up are documented explicitly.

## Related

- #312 — domain-specific vector-store search
- #454 — migrate bounded byte downloads back to `CalloutClient`
- praxis-proxy/praxis#826 — core callout hardening

leseb:
## Summary

Extracts common HTTP client functionality from `FilesApiClient` into a shared `api_client` module (`apis/src/openai/api_client/`) so both the Files API client and the future vector-store search client (#312) can reuse a single implementation.

- **Dual-transport design**: `CalloutClient` for JSON requests (`get_json`/`post_json`) with circuit breaking and depth protection; direct `reqwest::Client` for bounded byte downloads (`get_bytes`) with chunk-by-chunk size enforcement via `read_bounded_body`.
- **Extracted shared utilities**: SSRF validation, resource-ID path-segment encoding, and forward-header validation moved to `api_client::url`.
- **`FilesApiClient` simplified**: now wraps `ApiClient` instead of owning transport clients directly, reducing ~500 lines from `file_resolve/`.
- Once praxis-core publishes `max_response_bytes` on `CalloutConfig`, both paths can be unified behind `CalloutClient` (tracked in #454).

Closes #405

## Test plan

- [x] All existing `file_resolve` tests pass unchanged (behavioral preservation)
- [x] New `api_client` unit tests cover: JSON parsing, bounded reads, redirect rejection, transport failures, >1 MiB downloads, forward-header copying, SSRF validation, resource-ID encoding
- [x] `make build` / `make lint` / `make test` / `make doc` all pass

### DIFF
--- a/apis/src/openai/api_client/mod.rs
+++ b/apis/src/openai/api_client/mod.rs
@@ -5,8 +5,9 @@
 //!
 //! Provides URL construction, SSRF-safe base-URL validation,
 //! resource-ID path-segment encoding, header forwarding, bounded
-//! JSON and byte reads, and normalized error mapping. Used by
-//! [`FilesApiClient`] and (in the future) vector-store search.
+//! byte reads, JSON metadata fetches, and normalized error
+//! mapping. Used by [`FilesApiClient`] and (in the future)
+//! vector-store search.
 //!
 //! JSON requests route through `praxis_core::callout::CalloutClient`
 //! for circuit breaking, callout-depth protection, and timeout.
@@ -51,9 +52,9 @@ pub(crate) struct ApiClientConfig {
 
 /// Shared HTTP client for OpenAI-compatible API callouts.
 ///
-/// JSON requests (`get_json`, `post_json`) route through
-/// [`CalloutClient`] for circuit breaking and callout-depth
-/// protection. Content downloads (`get_bytes`) use a direct
+/// JSON requests (`get_json`) route through [`CalloutClient`]
+/// for circuit breaking and callout-depth protection. Content
+/// downloads (`get_bytes`) use a direct
 /// `reqwest::Client` with chunk-by-chunk bounded reads so
 /// oversized responses are rejected during collection without
 /// unbounded memory consumption. Unifying both paths behind
@@ -140,7 +141,7 @@ impl ApiClient {
     /// response body as JSON.
     pub(crate) async fn get_json(
         &self,
-        url: &str,
+        url: String,
         request_headers: &http::HeaderMap,
     ) -> Result<serde_json::Value, ApiClientError> {
         let headers = self.forward_headers(request_headers);
@@ -149,41 +150,7 @@ impl ApiClient {
             depth: 0,
             headers,
             method: http::Method::GET,
-            url: url.to_owned(),
-        };
-
-        let response = execute_callout(&self.client, request).await?;
-
-        serde_json::from_slice(&response.body).map_err(|e| ApiClientError::DecodeFailed {
-            detail: format!("JSON decode failed: {e}"),
-        })
-    }
-
-    /// Send a POST request with a JSON body via the callout client
-    /// and parse the response body as JSON.
-    pub(crate) async fn post_json(
-        &self,
-        url: &str,
-        body: &serde_json::Value,
-        request_headers: &http::HeaderMap,
-    ) -> Result<serde_json::Value, ApiClientError> {
-        let mut headers = self.forward_headers(request_headers);
-        headers.retain(|(name, _)| name != http::header::CONTENT_TYPE);
-        headers.push((
-            http::header::CONTENT_TYPE,
-            http::HeaderValue::from_static("application/json"),
-        ));
-
-        let serialized = serde_json::to_vec(body).map_err(|e| ApiClientError::DecodeFailed {
-            detail: format!("request body serialization failed: {e}"),
-        })?;
-
-        let request = CalloutRequest {
-            body: Some(serialized),
-            depth: 0,
-            headers,
-            method: http::Method::POST,
-            url: url.to_owned(),
+            url,
         };
 
         let response = execute_callout(&self.client, request).await?;
@@ -490,7 +457,7 @@ mod tests {
         let client = test_client(&format!("http://{address}"));
 
         let json = client
-            .get_json(&format!("http://{address}/v1/files/file-abc"), &http::HeaderMap::new())
+            .get_json(format!("http://{address}/v1/files/file-abc"), &http::HeaderMap::new())
             .await
             .unwrap();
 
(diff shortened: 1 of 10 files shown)


apis/src/openai/api_client/mod.rs
```diff
@@ -0,0 +1,710 @@
...
+
+        let request = CalloutRequest {
+            body: Some(serialized),
+            depth: 0,
+            headers,
+            method: http::Method::POST,
+            url: url.to_owned(),
+        };
+
+        let response = execute_callout(&self.client, request).await?;
```
franciscojavierarceo, line 156: I think this path is missing a deadline and size limit while reading the response body. `CalloutClient` applies the timeout while sending the request, then buffers the body outside that timeout. Since this is being added for the vector-store client, can we either leave `post_json` out until bounded callout reads are available, or enforce the body timeout and limit here?

apis/src/openai/api_client/mod.rs
```diff
@@ -0,0 +1,710 @@
...
+        url: &str,
+        request_headers: &http::HeaderMap,
+    ) -> Result<serde_json::Value, ApiClientError> {
+        let headers = self.forward_headers(request_headers);
+        let request = CalloutRequest {
+            body: None,
+            depth: 0,
+            headers,
+            method: http::Method::GET,
+            url: url.to_owned(),
```
franciscojavierarceo: This makes another `String` allocation even though the Files caller already owns the `String` returned by `resource_url`. Can `get_json`/`post_json` take a `String` (or `impl Into<String>`) so the caller can move it into `CalloutRequest`?

apis/src/openai/api_client/mod.rs
```diff
@@ -0,0 +1,710 @@
...
+    }
+
+    /// Send a GET request and return the response body with
+    /// chunk-by-chunk bounded reads.
+    ///
+    /// Uses a direct `reqwest::Client` so the `max_bytes` limit
+    /// is enforced during collection — oversized responses are
+    /// rejected without unbounded memory consumption. Redirects
+    /// are rejected, and the configured timeout spans the full
+    /// request including body transfer.
```
franciscojavierarceo, line 170: We had a `content_download_timeout_covers_response_body` regression test before this move, and I don't see an equivalent in `api_client`. Can we keep that coverage here? The current reqwest path does cover body transfer; I mostly want to keep the guardrail for the behavior this doc promises.

franciscojavierarceo (commented):
One tracking thing: this looks like part of [#405](https://github.com/praxis-proxy/ai/issues/405) rather than closing it. #405 also includes bounded CalloutClient reads and the vector-store migration, which are deferred here. Can we change the PR to `Part of #405` and leave it open? [#454](https://github.com/praxis-proxy/ai/issues/454) also looks like it needs an update now that `get_bytes` uses direct reqwest.

apis/src/openai/api_client/mod.rs
leseb: Good catch!

apis/src/openai/api_client/mod.rs
leseb: i have this tracked in https://github.com/praxis-proxy/ai/issues/454 but will leave post_json too thanks!

leseb:
> One tracking thing: this looks like part of [#405](https://github.com/praxis-proxy/ai/issues/405) rather than closing it. #405 also includes bounded CalloutClient reads and the vector-store migration, which are deferred here. Can we change the PR to `Part of #405` and leave it open? [#454](https://github.com/praxis-proxy/ai/issues/454) also looks like it needs an update now that `get_bytes` uses direct reqwest.

Actually went to far on the scope of 405, good catch, my intent was to have a simple ApiClient for all the filters callout to ogx, that's it, then the domain specific will be added by each filter. Essentially, ApiClient is a harden version of HTTPCallout to avoid repeating the same patterns on every filter. I have a proposal to reconcile some of the HttpCallout hardening into core here https://github.com/praxis-proxy/praxis/issues/826. Scoping down 405 so this PR should close it!

apis/src/openai/api_client/mod.rs
leseb: https://github.com/praxis-proxy/ai/pull/456/commits/894713ce6b9766981b3aaad48c19ca976206ddbe

apis/src/openai/api_client/mod.rs
leseb: https://github.com/praxis-proxy/ai/pull/456/commits/894713ce6b9766981b3aaad48c19ca976206ddbe

apis/src/openai/api_client/mod.rs
leseb: https://github.com/praxis-proxy/ai/pull/456/commits/894713ce6b9766981b3aaad48c19ca976206ddbe

franciscojavierarceo (approved):

PR #491, opened 2026-07-23
feat(ci): add OGX file_id integration tests to vLLM workflow

leseb:
## Summary

- Adds an end-to-end integration test (`test_file_id_resolution_through_ogx`) that validates the `openai_file_resolve` filter against a real OGX Files API, exercising the full pipeline: upload file to OGX, reference by `file_id` through Praxis, file_resolve fetches from OGX, doc_extract converts to input_text, vLLM inference.
- Updates the vLLM CI workflow to start and manage an OGX server alongside vLLM, with readiness checks, failure logging, and cleanup.
- Patches `_write_config` to point the `files_api_url` at OGX (configurable via `OGX_BASE_URL` env var, default `http://127.0.0.1:8321`).

## Test plan

- [ ] CI: vLLM integration workflow passes with OGX started and file_id test exercised
- [ ] Local: `uv run ogx run starter --insecure` + vLLM running, then `uv run tests/integration/sdk/openai/test_openai_responses_vllm.py -s -k test_file_id`
- [ ] Existing tests remain unaffected (file_resolve filter only activates on `file_id` references)

### DIFF
--- a/tests/integration/sdk/openai/test_openai_responses_vllm.py
+++ b/tests/integration/sdk/openai/test_openai_responses_vllm.py
@@ -164,17 +164,6 @@ def openai_client(praxis_proxy):
     )
 
 
-@pytest.fixture(scope="session")
-def ogx_client():
-    """Return an OpenAI client pointed directly at the OGX Files API."""
-    return OpenAI(
-        base_url=f"{OGX_BASE_URL}/v1",
-        api_key="test",
-        max_retries=0,
-        timeout=30,
-    )
-
-
 # ---------------------------------------------------------------------------
 # Tests
 # ---------------------------------------------------------------------------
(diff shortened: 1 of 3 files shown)


tests/integration/sdk/openai/test_openai_responses_vllm.py
```diff
@@ -269,6 +289,60 @@ def test_doc_extract_inline_file_to_input_text(self, openai_client):
...
+        verify vLLM output contains the file content.
+
+        Pipeline: file_resolve (OGX) -> doc_extract -> responses_proxy -> vLLM
+        """
+        import io
+
+        marker = "PRAXIS-OGX-FILE-4829"
+        file_content = f"The secret marker is: {marker}"
+
+        uploaded = ogx_client.files.create(
```
franciscojavierarceo: why not use openai client instead btw?

franciscojavierarceo (commented):
let's use openai client instead of ogx client

tests/integration/sdk/openai/test_openai_responses_vllm.py
leseb: it's actually OpenAI, yeah it shoulnd't be called ogx

leseb:
> let's use openai client instead of ogx client

it's just the name that was confusing, it always OpenAI client :) renamed ogx_client -> openai_client

franciscojavierarceo (approved):

PR #492, opened 2026-07-23
refactor(api_client): unify get_bytes behind CalloutClient

### LINKED ISSUE #454: refactor(apis): upgrade praxis-core for bounded callout reads

leseb:
## Summary

Unifies `get_bytes` behind `CalloutClient` instead of a separate `reqwest::Client`, completing the work unblocked by the praxis-core 0.4.1 bump (#486). Content downloads now share circuit-breaker, callout-depth protection, and bounded reads via `CalloutConfig::max_response_bytes`. A per-request `max_bytes` post-collection check preserves `ResponseTooLarge` → 413 for file-level budgets.

- Removed `content_client` field and `read_bounded_body` from `ApiClient`
- Routed `get_bytes` through `CalloutClient` with `tokio::time::timeout` wrapping the full operation
- Set `max_response_bytes` to the filter's `max_body_bytes` in `openai_file_resolve`
- Added test for client-wide `max_response_bytes` rejection

Closes #454

### DIFF
--- a/apis/src/openai/api_client/mod.rs
+++ b/apis/src/openai/api_client/mod.rs
@@ -8,17 +8,14 @@
 //! JSON and byte reads, and normalized error mapping. Used by
 //! [`FilesApiClient`] and vector-store search.
 //!
-//! JSON requests route through `praxis_core::callout::CalloutClient`
-//! for circuit breaking, callout-depth protection, and timeout.
-//! Content downloads use a direct `reqwest::Client` with
-//! chunk-by-chunk bounded reads so oversized responses are rejected
-//! during collection. Unifying both paths behind `CalloutClient`
-//! is tracked in [#454].
+//! All requests route through `praxis_core::callout::CalloutClient`
+//! for circuit breaking, callout-depth protection, timeout, and
+//! bounded reads via `CalloutConfig::max_response_bytes`. Content
+//! downloads (`get_bytes`) apply a tighter per-request `max_bytes`
+//! check post-collection so callers can enforce file-level budgets.
 //!
 //! Each consuming filter retains its own [`ApiClient`] instance so
 //! circuit-breaker state is isolated per filter.
-//!
-//! [#454]: https://github.com/praxis-proxy/ai/issues/454
 //!
 //! [`FilesApiClient`]: super::responses::file_resolve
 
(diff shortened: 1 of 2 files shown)


apis/src/openai/api_client/mod.rs
aslakknutsen, line 257: Issue #454 requires preserving ResponseTooLarge so openai_file_resolve returns 413 for size limits. The per-request max_bytes post-check correctly maps to ResponseTooLarge -> 413 when the body is fully collected under max_response_bytes. When collection hits the client-wide max_response_bytes first, praxis-core read_body fails, execute_callout maps CalloutResult::Rejected to CalloutFailed, and resolve_error_response returns 502. Example: max_body_bytes=64 with a 70 MiB download and absent or understated metadata.bytes collects until the client-wide cap, then returns 502 instead of 413. get_bytes_client_wide_limit_rejects_oversized_chunked_response encodes CalloutFailed for this path rather than ResponseTooLarge. Map client-wide size rejections to ResponseTooLarge (or propagate a distinct size error from execute_callout) before they reach the filter.

aslakknutsen (changes requested):

leseb:
CalloutClient is going away in https://github.com/praxis-proxy/praxis/pull/849 so closing.

PR #536, opened 2026-07-24
refactor(token_usage): move token usage module from apis to filters

leseb:
## Summary

Move the `token_usage` module from `apis/src/token_usage/` into `filters/src/token_usage/`, consolidating it alongside its only production consumers (`token_count` and `token_usage_headers` filters). This eliminates duplicate provider enums and conversion glue between the two crates, since token usage extraction writes `HttpFilterContext` metadata which is filter behavior rather than API-domain logic.

The new `filters/src/token_usage/` subsystem contains:
- `mod.rs` — shared metadata contract and provider configuration
- `count.rs` — `TokenCountFilter` (renamed from `token_count`)
- `headers.rs` — `TokenUsageHeadersFilter` (renamed from `token_usage_headers`)
- `providers.rs` — provider-specific token extraction
- `streaming.rs` — SSE streaming token accumulation

Documentation (proposals, filter reference, AGENTS.md) updated to reflect the new paths.


AGENTS.md
```diff
@@ -134,12 +134,11 @@ New capabilities require:
   SSE, model rewrite, store, rehydrate, validate, proxy
```
aslakknutsen, line 134: Architecture crate summary still lists "token usage extraction" under apis and "header injection" under filters. Update `AGENTS.md:65-72` to match both under "filter"

aslakknutsen (changes requested):

AGENTS.md
leseb: https://github.com/praxis-proxy/ai/pull/536/commits/40dbcd3e5174cfc9609d2c692fe31f791517370c

aslakknutsen (approved):

PR #570, opened 2026-07-27
fix(conversations): atomic item mutations with transactional message cache sync

### LINKED ISSUE #358: fix(conversations): refresh_message_cache constructs misleading partial ConversationRecord
`filter.rs:583-588` constructs a `ConversationRecord` with `created_at: 0`
and `messages: Value::Null` just to pass `tenant_id` and `conversation_id`
to `sync_conversation_messages`. If that function ever accesses other
fields, the bogus values would silently corrupt data.

**Fix:** Refactor `sync_conversation_messages` to accept
`(tenant_id, conversation_id)` directly instead of a full record.

eoinfennessy:
## Summary

- Adds `create_items_and_sync_messages` and `delete_item_and_sync_messages` to the `ConversationItemStore` trait, combining position assignment, item mutation, and message cache rebuild into single database transactions. This eliminate races under concurrent appends and stale cache overwrites
- Removes `refresh_message_cache` and `sync_conversation_messages`, which constructed a `ConversationRecord` with bogus field values just to thread `tenant_id` and `conversation_id` through to the sync logic
- Adds a unique index on `(tenant_id, conversation_id, position)` as defense-in-depth against position collisions
- Uses `BEGIN IMMEDIATE` (SQLite) and `SELECT ... FOR UPDATE` (PostgreSQL) to serialize concurrent writes

## Test plan

- [x] 5 new SQLite tests covering position assignment, multi-batch appends, deletion with cache sync, empty batches, and nonexistent items
- [x] 2 new PostgreSQL tests covering position assignment and deletion with cache sync
- [x] Existing `duplicate_positions` test replaced with `duplicate_position_rejected_by_unique_constraint`
- [x] Full test suite passes (2005 tests)

Closes #358

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -155,7 +155,7 @@ audit:
 coverage-check:
 	cargo llvm-cov --workspace --json \
 		--exclude xtask \
-		--ignore-filename-regex '(target/|tests/)' \
+		--ignore-filename-regex '(target/|tests/|store/postgres\.rs)' \
 		--output-path coverage.json
 	@LINE_PCT=$$(jq '.data[0].totals.lines.percent' coverage.json); \
 	echo "Line coverage: $${LINE_PCT}%"; \
--- a/apis/src/openai/conversations/filter.rs
+++ b/apis/src/openai/conversations/filter.rs
@@ -410,6 +410,7 @@ impl HttpFilter for OpenaiConversationsFilter {
         true
     }
 
+    #[expect(clippy::too_many_lines, reason = "dispatcher with one arm per endpoint")]
     async fn on_request(&self, ctx: &mut HttpFilterContext<'_>) -> Result<FilterAction, FilterError> {
         let Some(route) = routes::match_route(ctx.request.method.as_str(), ctx.request.uri.path()) else {
             if should_append_back(ctx) {
@@ -625,14 +626,9 @@ async fn persist_items(
     ctx: &HttpFilterContext<'_>,
     items: Vec<Value>,
 ) -> Result<(), FilterError> {
-    let max_pos = store
-        .max_item_position(tenant_id, conversation_id)
-        .await
-        .map_err(|e| -> FilterError { Box::new(e) })?;
-    let start_position = max_pos.saturating_add(1);
     let created_at = handlers::current_timestamp(ctx);
 
-    let records = handlers::build_item_records(ctx, tenant_id, conversation_id, created_at, start_position, items)
+    let records = handlers::build_item_records(ctx, tenant_id, conversation_id, created_at, 0, items)
         .map_err(|e| -> FilterError { e.into() })?;
 
     if records.is_empty() {
@@ -641,49 +637,16 @@ async fn persist_items(
 
     let count = records.len();
     store
-        .create_conversation_items(&records)
+        .create_items_and_sync_messages(&records)
         .await
         .map_err(|e| -> FilterError { Box::new(e) })?;
 
-    // Item rows are durable past this point. The message cache is a self-healing
-    // projection rebuilt from the items table on the next successful sync, so a
-    // refresh failure is not data loss and must not fail the turn: propagating it
-    // would abort a request whose items already committed and drive a client retry
-    // that re-appends the same (typically id-less) input items as duplicates
-    // (#837). Log and continue; a later successful cache-refresh re-syncs the cache.
-    refresh_message_cache(store, tenant_id, conversation_id).await;
     debug!(
         conversation_id,
         tenant_id, count, "conversation items appended from response"
     );
 
     Ok(())
-}
-
-/// Refresh the denormalized conversation message cache after item mutation.
-///
-/// The cache is a projection of the items table that `openai_responses_rehydrate`
-/// replays on the next turn. It runs *after* the item rows have committed, so it
-/// is deliberately best-effort: the sync is idempotent — it rebuilds `messages`
-/// from the durable items table — so a later successful sync re-syncs whatever this
-/// attempt left behind (a later *append* is not sufficient: its own refresh may
-/// also fail). Failing the turn here would abort a request whose
-/// items already persisted and push the client into a retry that re-appends the
-/// same items as duplicates, so a refresh failure is logged and swallowed rather
-/// than propagated (#837). A transient stale-cache window remains until the next
-/// successful sync; closing it fully (atomic item-insert + projection) is tracked
-/// separately.
-async fn refresh_message_cache(store: &dyn ConversationItemStore, tenant_id: &str, conversation_id: &str) {
-    // The append-back path holds no pre-mutation cache snapshot, so it syncs with
-    // `None`; the refresh then re-reads the live cache as the swap's expected value.
-    if let Err(e) = handlers::sync_conversation_messages(store, tenant_id, conversation_id, None).await {
-        warn!(
-            error = %e,
-            conversation_id,
-            "conversation message cache refresh failed after append-back; items are durable and the \
-             cache re-syncs on a later successful refresh (#837)"
-        );
-    }
 }
 
 // -----------------------------------------------------------------------------
--- a/apis/src/openai/conversations/filter.rs
+++ b/apis/src/openai/conversations/filter.rs
@@ -637,7 +637,7 @@ async fn persist_items(
 
     let count = records.len();
     store
-        .create_items_and_sync_messages(&records)
+        .create_items_and_sync_messages(tenant_id, conversation_id, &records)
         .await
         .map_err(|e| -> FilterError { Box::new(e) })?;
 
--- a/apis/src/openai/conversations/filter.rs
+++ b/apis/src/openai/conversations/filter.rs
@@ -410,7 +410,6 @@ impl HttpFilter for OpenaiConversationsFilter {
         true
     }
 
-    #[expect(clippy::too_many_lines, reason = "dispatcher with one arm per endpoint")]
     async fn on_request(&self, ctx: &mut HttpFilterContext<'_>) -> Result<FilterAction, FilterError> {
         let Some(route) = routes::match_route(ctx.request.method.as_str(), ctx.request.uri.path()) else {
             if should_append_back(ctx) {
--- a/apis/src/openai/conversations/filter.rs
+++ b/apis/src/openai/conversations/filter.rs
@@ -526,10 +526,8 @@ impl HttpFilter for OpenaiConversationsFilter {
         // yields a 2xx followed by a reset — but either way the body is withheld.
         // `failure_mode: open` is an explicit operator opt-out of that guarantee:
         // the pipeline logs this error and converts it to Continue, releasing the
-        // body even though items were lost. Only pre-commit failures (item insertion
-        // and earlier) reach this `?`: a post-commit message-cache refresh failure
-        // is tolerated inside `persist_items` because the cache is a self-healing
-        // projection (see `refresh_message_cache`).
+        // body even though items were lost. Transactional item insertion and cache
+        // rebuild failures reach this `?` before any append-back bytes are released.
         self.append_items_blocking(&items.tenant_id, &conv_id, ctx, items.all_items)
             .inspect_err(|e| warn!(error = %e, conversation_id = %conv_id, "conversation append-back failed"))?;
 
--- a/apis/src/openai/conversations/handlers.rs
+++ b/apis/src/openai/conversations/handlers.rs
@@ -12,7 +12,9 @@ use serde::{
     de::{DeserializeOwned, MapAccess, Visitor, value::MapAccessDeserializer},
 };
 use serde_json::{Map, Value};
-use tracing::{debug, warn};
+use tracing::debug;
+#[cfg(test)]
+use tracing::warn;
 
 use super::{
     contracts::{
@@ -41,6 +43,7 @@ use crate::{
 /// up on refreshing the cache. Each failed swap means another writer refreshed
 /// the cache first from the same authoritative rows, so a small bound absorbs
 /// realistic append contention.
+#[cfg(test)]
 const MAX_SYNC_ATTEMPTS: usize = 8;
 
 // -----------------------------------------------------------------------------
@@ -948,6 +951,7 @@ fn store_error_response(error: &StoreError) -> Result<Rejection, FilterError> {
 /// assumed to be small: the OpenAI contract has no cumulative item or byte
 /// ceiling. Replace this full-history rebuild with incremental processing; do
 /// not add a non-spec conversation limit as a workaround. Tracked in #532.
+#[cfg(test)]
 pub(super) async fn sync_conversation_messages(
     store: &dyn ConversationItemStore,
     tenant_id: &str,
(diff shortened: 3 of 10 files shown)


apis/src/store/tests.rs
```diff
@@ -1314,6 +1303,165 @@ async fn conversation_item_methods_fail_without_items_table() {
...
+    let conv_record = ConversationItemStore::get_conversation(&store, "tenant_a", "conv_1")
+        .await
+        .expect("get should succeed")
+        .expect("conversation should exist");
+    let messages = conv_record.messages.as_array().expect("messages should be an array");
+    assert_eq!(messages.len(), 2, "messages cache should have 2 items");
+}
+
+#[tokio::test]
+async fn create_items_and_sync_messages_continues_from_max_position() {
```
aslakknutsen, line 1394: Sequential tests prove position assignment and cache coherence on happy paths, but nothing interleaves two `create_items_and_sync_messages` calls against the same conversation. A lightweight SQLite test spawning two tasks appending one item each and asserting distinct positions plus a two-item message cache would directly guard the regression described in the PR.

apis/src/openai/conversations/handlers.rs
```diff
@@ -104,26 +104,33 @@ pub(super) async fn handle_create_conversation(
...
     }
     if !item_records.is_empty()
-        && let Err(e) = store.create_conversation_items(&item_records).await
+        && let Err(e) = store.create_items_and_sync_messages(&item_records).await
     {
         return Ok(FilterAction::Reject(store_error_response(&e)?));
     }
     debug!(conversation_id, tenant_id, "conversation created");
 
+    let record = if item_records.is_empty() {
```
aslakknutsen: After `create_items_and_sync_messages` succeeds, `handle_create_conversation` re-fetches the conversation but falls back to the pre-sync record (empty `messages`) on `get_conversation` error or `None`. That yields a 200 with an empty messages array even though items and cache were written. Propagate the fetch error or build the response from the inserted item data instead of silently degrading.

apis/src/store/sqlite.rs
```diff
@@ -561,6 +561,231 @@ impl ConversationItemStore for SqliteResponseStore {
...
+            return Ok(());
+        };
+
+        let items_table = self
+            .tables
+            .items
+            .as_deref()
+            .ok_or_else(|| StoreError::Unavailable("items table not configured".to_owned()))?;
+        let conv_table = &self.tables.conversations;
+        let tenant_id = &first.tenant_id;
```
aslakknutsen: `create_items_and_sync_messages` derives `tenant_id` and `conversation_id` from `items.first()` for the MAX-position query and cache rebuild, yet inserts each record with its own `tenant_id` / `conversation_id` fields. Current callers always pass a homogeneous batch, so behavior is correct today. A future caller mixing conversations in one slice would assign positions from one conversation's max while inserting into another. Document the invariant on the trait or validate homogeneity before opening the transaction. Other options?

aslakknutsen (changes requested):

apis/src/store/tests.rs
eoinfennessy: Done; Added a concurrent write test in [7fbaeb5](https://github.com/praxis-proxy/ai/pull/570/commits/7fbaeb5749496ab1a71aeddc2fa95a4793208862)

apis/src/openai/conversations/handlers.rs
eoinfennessy: Done; Commit [f747003](https://github.com/praxis-proxy/ai/pull/570/commits/f74700342e21e7f3c15da2d781bea78787b309bb) builds the response from the inserted item data

apis/src/store/sqlite.rs
eoinfennessy: The API boundary already enforces homogeneity. `build_item_records` stamps every item with the same `tenant_id` and `conversation_id` from the request, so a caller can't mix conversations through the external API.

To ensure future internal callers behave the same way, I added a `debug_assert!` to both implementations in [1c9836e](https://github.com/praxis-proxy/ai/pull/570/commits/1c9836e9f478c0e40d1c54372706e6ccd9dbed8c). This guards against a new call site wiring things up incorrectly without adding runtime cost in release builds. I also added a `# Panics` section on the trait doc.

apis/src/openai/conversations/filter.rs
```diff
@@ -583,11 +578,10 @@ async fn persist_items(
 
     let count = records.len();
     store
-        .create_conversation_items(&records)
+        .create_items_and_sync_messages(&records)
```
aslakknutsen: Wouldn't it be better to make it `create_items_and_sync_messages(tenant_id, conversation_id, items)?

* You already know the tenant_id/conversation_id's needed
* No need to ignore the start_position argument
* makes debug_assert! unnecessary
* matches what the method already does

apis/src/store/schemas.rs
```diff
@@ -155,6 +155,10 @@ fn append_items_ddl(stmts: &mut Vec<String>, i: &str) {
         "CREATE INDEX IF NOT EXISTS idx_{i}_conversation \
          ON {i}(conversation_id, tenant_id, position, item_id)"
     ));
+    stmts.push(format!(
+        "CREATE UNIQUE INDEX IF NOT EXISTS idx_{i}_position \
```
franciscojavierarceo, line 192: **[P1] This automatic unique-index creation is unsafe for upgrades.** The previous schema and tests allowed duplicate `(tenant_id, conversation_id, position)` rows, and the race being fixed could have produced them in deployed databases. On restart, creating this index then fails store initialization; I reproduced SQLite returning `UNIQUE constraint failed`. Please add an explicit migration that deterministically resequences legacy duplicates before creating the constraint, with upgrade coverage for both backends.

apis/src/store/sqlite.rs
```diff
@@ -561,6 +561,235 @@ impl ConversationItemStore for SqliteResponseStore {
...
+            "all items must belong to the same tenant and conversation"
+        );
+
+        let mut conn = self
+            .pool
+            .acquire()
+            .await
+            .map_err(|e| StoreError::Database(e.to_string()))?;
+
+        sqlx::query("BEGIN IMMEDIATE")
```
franciscojavierarceo: **[P1] Use an SQLx transaction guard for `BEGIN IMMEDIATE`.** Issuing `BEGIN`/`COMMIT`/`ROLLBACK` as raw queries bypasses SQLx transaction bookkeeping and rollback-on-drop. If this future is cancelled after `BEGIN`, or `COMMIT` fails, the pooled connection can be returned with an open transaction holding SQLite writer lock. The delete path has the same pattern. `pool.begin_with("BEGIN IMMEDIATE")` provides the required RAII rollback behavior; please add a cancellation/abort regression test.

apis/src/store/sqlite.rs
```diff
@@ -561,6 +561,235 @@ impl ConversationItemStore for SqliteResponseStore {
...
+) -> Result<(), StoreError> {
+    let select_sql = format!(
+        "SELECT item_data FROM {items_table} \
+         WHERE tenant_id = ? AND conversation_id = ? \
+         ORDER BY position ASC, item_id ASC"
+    );
+    let rows = sqlx::query(AssertSqlSafe(select_sql.as_str()))
+        .bind(tenant_id)
+        .bind(conversation_id)
+        .fetch_all(&mut **conn)
```
franciscojavierarceo: **[P2] This full-history rebuild now runs while `BEGIN IMMEDIATE` holds SQLite database-wide writer lock.** Every append/delete blocks unrelated tenant writes through the complete fetch, parse, and serialization pass. `fetch_all` also keeps all raw rows alive alongside the parsed `Vec<Value>` and final JSON string. Please consume rows incrementally to reduce peak memory and avoid rebuilding the entire cache inside the writer-critical section where possible.

franciscojavierarceo (changes requested):
Requesting changes for two merge-blocking correctness issues and one SQLite performance regression. The existing mixed-batch API thread at filter.rs:581 also remains valid and unresolved.\n\nI reproduced the legacy-duplicate initialization failure on SQLite. Targeted transactional tests otherwise passed locally, while the PR is currently also failing nightly rustfmt and the coverage threshold.

apis/src/store/sqlite.rs
eoinfennessy: The rebuild (fetch all items → serialize → update messages column) is pre-existing behavior from `refresh_message_cache` / `sync_conversation_messages` in the filter layer. This PR moves it inside the transaction so the items rows and the denormalized messages cache are updated atomically — previously they were two independent operations with no consistency guarantee.

On the writer lock concern: SQLite's `BEGIN IMMEDIATE` is database-wide regardless, so writers were already serialized before this PR. The difference is holding the lock slightly longer to include the rebuild, which keeps the items and messages cache consistent. Splitting them back out would reintroduce the original atomicity gap.

The memory profile of the rebuild is unchanged from the existing `collect_conversation_messages` / `sync_conversation_messages` path on main — it already collects all item values into a `Vec<Value>` and serializes the full array. This PR moves that work inside the transaction but doesn't change the allocation pattern. Optimizing it (streaming rows, incremental serialization) is worth doing but is a separate, out of scope concern.

apis/src/store/schemas.rs
eoinfennessy: Are upgrades a concern at this early stage of the project? Are there any deployed databases?

I considered adding a dedup step from the old schema, which incorrectly allowed duplicate `(tenant_id, conversation_id, position)` entries, but it kind of seems like it would be adding noise for an upgrade scenario that does not yet exist.

apis/src/store/sqlite.rs
eoinfennessy: Done in [83da276](https://github.com/praxis-proxy/ai/pull/570/commits/83da276a3ae4ec5efa439061f3c2d4681c46b3ab)

apis/src/store/schemas.rs
```diff
@@ -155,6 +155,10 @@ fn append_items_ddl(stmts: &mut Vec<String>, i: &str) {
         "CREATE INDEX IF NOT EXISTS idx_{i}_conversation \
          ON {i}(conversation_id, tenant_id, position, item_id)"
```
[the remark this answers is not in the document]
eoinfennessy: See https://github.com/praxis-proxy/ai/pull/570#discussion_r3660871963

apis/src/openai/conversations/filter.rs
eoinfennessy: Done in [efa1a40](https://github.com/praxis-proxy/ai/pull/570/commits/efa1a40). The signature now takes explicit `tenant_id` and `conversation_id` params, and the `debug_assert!` is removed.

One thing to note: the INSERT still binds `item.tenant_id` and `item.conversation_id` from each record rather than the method parameters. In practice these always match (all callers go through `build_item_records` which stamps every item uniformly), but should we go further and use the method params for the INSERT too? That would make the item's own `tenant_id`/`conversation_id` fully ignored, the same as `position` already is.

apis/src/openai/conversations/handlers.rs
```diff
@@ -105,21 +105,20 @@ pub(super) async fn handle_create_conversation(
...
     }
-    let messages = Value::Array(item_records.iter().map(|item| item.item_data.clone()).collect());
 
     let record = ConversationRecord {
         conversation_id: conversation_id.clone(),
         tenant_id: tenant_id.to_owned(),
         created_at,
         metadata,
-        messages,
+        messages: Value::Array(Vec::new()),
```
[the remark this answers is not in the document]
eoinfennessy: Not applicable. `conversation_response()` converts the record to a `ConversationResource` ([contracts.rs:125–136](vscode-webview://04i0j7o52hjad5egt0if053mldhici9ha4in2heg8ikcb0meip4m/apis/src/openai/conversations/contracts.rs#L125-L136)), which only serializes `id`, `object`, `created_at`, and `metadata`. There is no `messages` field on the response type, so the value in the `ConversationRecord` is never surfaced to the caller.

The messages cache is correctly populated in the database by `create_items_and_sync_messages` on line 122.

apis/src/store/schemas.rs
leseb: This is greenfield, no migration.

apis/src/store/postgres.rs
```diff
@@ -622,6 +622,188 @@ impl ConversationItemStore for PostgresResponseStore {
...
+            "SELECT 1 FROM {conv_table} \
+             WHERE conversation_id = $1 AND tenant_id = $2 \
+             FOR UPDATE"
+        );
+        sqlx::query(AssertSqlSafe(lock_sql.as_str()))
+            .bind(conversation_id)
+            .bind(tenant_id)
+            .fetch_optional(&mut *tx)
+            .await
+            .map_err(|e| StoreError::Database(e.to_string()))?;
```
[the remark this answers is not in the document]
eoinfennessy: Done in [da27ad4](https://github.com/praxis-proxy/ai/pull/570/commits/da27ad43c49deb95a981caee8f5d5e81e8a592c2)

eoinfennessy:
Hi @aslakknutsen @franciscojavierarceo, I have addressed your requested changes. PTAL when you get a chance

leseb:
@aslakknutsen @franciscojavierarceo PTAL to unblock this, thanks

Makefile
```diff
@@ -113,7 +113,7 @@ audit:
 coverage-check:
 	cargo llvm-cov --workspace --json \
 		--exclude xtask \
-		--ignore-filename-regex '(target/|tests/)' \
+		--ignore-filename-regex '(target/|tests/|store/postgres\.rs)' \
```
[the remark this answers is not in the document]
aslakknutsen: @eoinfennessy Could you add a tracker issue to ensure we get some postgres testing in ci at some point?

apis/src/store/postgres.rs
```diff
@@ -694,6 +694,200 @@ impl ConversationItemStore for PostgresResponseStore {
...
+        );
+        for (i, item) in items.iter().enumerate() {
+            let offset = i64::try_from(i).unwrap_or(i64::MAX);
+            let position = max_pos.saturating_add(1).saturating_add(offset);
+            let item_data =
+                serde_json::to_string(&item.item_data).map_err(|e| StoreError::Serialization(e.to_string()))?;
+
+            sqlx::query(AssertSqlSafe(insert_sql.as_str()))
+                .bind(&item.item_id)
+                .bind(&item.tenant_id)
```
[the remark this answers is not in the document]
aslakknutsen: @eoinfennessy This sounds like a sane enough "security" feature...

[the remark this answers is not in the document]
eoinfennessy: Done in [583e480](https://github.com/praxis-proxy/ai/pull/570/commits/583e480cf2b1b2a40e60bf5a4d2436e736711118)

Makefile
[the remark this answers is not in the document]
eoinfennessy: #739 set up postgres testing for CI. I filed #759 to track wiring `store/postgres.rs` into the coverage gate itself. The exclusion should stay in place until that lands.

eoinfennessy:
@aslakknutsen @franciscojavierarceo can you please review again to unblock this?

`test-praxis-main` now fails, but I don't think this is related to my changes.

aslakknutsen (approved):

franciscojavierarceo (approved):

leseb:
@eoinfennessy please rebase

Issue #577, opened 2026-07-27
feat(filter): add external metering filter for usage reporting and balance checks


noyitz:
## Description

Add an `external_metering` filter that integrates with an external metering/billing service for:

1. **Pre-request balance check**: Queries the metering service to verify the tenant has available token budget before forwarding the request to the LLM provider. Rejects with 429 if budget is exhausted.
2. **Post-response usage reporting**: After the response completes, reads token counts from `filter_metadata` (set by the existing `token_count` filter) and sends a CloudEvents-formatted usage report to the metering service.

### Key design points

- **Single filter** (not two like the Go/ext_proc predecessor) — leverages the existing `token_count` filter for usage extraction from all providers (OpenAI, Anthropic, Google, Bedrock, Azure)
- **Identity header capture**: Captures configurable identity headers (default prefix: `x-tenant-`) containing username, group, subscription. Strips them before forwarding upstream.
- **Fail-open by default**: If the metering service is unavailable, requests proceed normally (configurable to fail-closed)
- **Fire-and-forget usage reporting**: Usage events are sent asynchronously via `tokio::spawn` — never blocks the response path
- **CloudEvents format**: Standard CloudEvents 1.0 envelope with `inference.tokens.used` and `inference.request.error` event types
- Uses `CalloutClient` from praxis-core for HTTP calls with circuit breaking and timeout

### API contract

**Balance check (pre-request):**
```
GET {metering_url}/api/v1/customers/{username}/entitlements/{feature_key}/value?model={model}
Response: { "hasAccess": true, "balance": 9000.0, "usage": 1000.0, "overage": 0.0 }
```

**Usage report (post-response):**
```
POST {metering_url}/api/v1/events
Body: CloudEvent JSON with token usage data
```

### Configuration

```yaml
filter: external_metering
metering_url: "http://metering-service:8080"
timeout_seconds: 5
feature_key: "inference-tokens"
source: "ai-gateway"
fail_open: true
identity_header_prefix: "x-tenant-"
```

### Pipeline ordering

```yaml
filters:
  - filter: external_metering   # request: first, response: LAST
  - filter: token_count          # response: writes metadata before metering reads it
  - filter: router
  - filter: load_balancer
```

## Motivation

Production AI gateway deployments need per-tenant token quota enforcement and usage tracking for billing/chargeback. Praxis has token extraction (`token_count`) but no way to report usage to an external metering or billing system.

[...]

PR #581, opened 2026-07-28
feat(metering): add external metering filter with balance checks and usage reporting

noyitz:
Adds an `external_metering` HTTP filter: pre-request balance checks and
post-response token usage reporting against an external metering service.

Part of #577.

## What the filter does

**Request phase** — resolves tenant identity, strips identity headers and
client credentials (`authorization`, `x-api-key`), and asks the metering
service whether the tenant may spend more tokens:

- `GET {metering_url}/api/v1/customers/{username}/entitlements/{feature_key}/value?model={model}`
- Exhausted budget → 429. Metering unreachable → admit when `fail_open: true`
  (default), 503 otherwise, so a metering outage degrades to unmetered service
  rather than an inference outage.

**Response phase** — emits a CloudEvents 1.0 event
(`inference.tokens.used`, or `inference.request.error` on upstream failure) to
`POST {metering_url}/api/v1/events`. Delivery is fire-and-forget: a slow or
failing metering service never delays a response the upstream already
answered. Token counts come from the `token_count` filter's `filter_metadata`
keys (`token.input`/`output`/`total` and the prompt cache breakdown from #582).

## Identity resolution — three tiers, most trusted first

1. **Verified metadata** (unnamespaced `{prefix}*` keys) written by an
   authentication filter from verified credentials. When present, every lower
   tier is ignored entirely, so a client cannot spoof `subscription` or
   `model` via forged headers alongside valid credentials.
2. **Guard metadata** (namespaced `identity.{prefix}*` keys) written by the
   `identity_header_guard` filter (#709), which owns identity header capture.
3. **Raw `{prefix}*` headers**, for deployments where a trusted upstream auth
   layer (e.g. an external authorizer) injects them directly.

Identity headers are always stripped before the request reaches the upstream,
regardless of which tier supplied the identity.

## Relationship to #709 and #577

This is the metering half of the split: #709 owns identity header capture into
namespaced metadata; this filter consumes that metadata (tier 2) and owns
admission control, credential stripping, and usage reporting. #577 documents
the original single-filter design and needs an update to reflect the split.

HTTP callouts use `praxis-core`'s `SubRequestClient` (the `CalloutClient` named
in #577 was removed upstream in #849). `register_ai_filters` threads the shared
server-level client into the filter following the existing
`openai_file_resolve` pattern; without one the filter creates a private
connector, same as `openai_file_resolve::from_config`.
`praxis_ai_apis::subrequest::execute_url` is made public so filters can execute
full-URL sub-requests.

## Size

1,876 insertions, of which 861 are unit + integration tests, 103 generated
docs + README rows, and 64 the example config. Functional filter code is
~1,010 lines across `mod.rs`, `config.rs`, and registration. Balance check,
event construction, and identity resolution are cohesive enough that splitting
them again would leave non-functional intermediate states; happy to split if
reviewers prefer.

## Testing

- 41 unit tests (`filters/src/metering/tests.rs`), including spoofing-closure
  tests for the tier precedence rules
- 5 functional integration tests against `examples/configs/external-metering.yaml`
  (balance allowed, fail-closed rejection, header stripping, no-identity skip)
- `make test` — 4,905 passed, 0 failed; `make lint` green


leseb:
@noyitz please resolve the conflicts and address the bot's comments thanks

filters/src/metering/mod.rs
```diff
@@ -0,0 +1,857 @@
...
+            &url,
+            request,
+            MAX_CALLOUT_RESPONSE_BYTES,
+            self.timeout,
+        )
+        .await;
+
+        match result {
+            Ok(resp) if is_success(resp.status) => parse_balance_result(&resp.body, self.fail_open),
+            Ok(resp) => {
```
jordigilh, line 274: Any non-2xx from the metering service (403/404/500) hits the same `on_metering_unavailable()` path as a real outage (L239-242) -- under the default `fail_open: true`, an explicit denial gets silently admitted. Is a 4xx from this API ever meant to signal denial vs. unavailability? If so, worth splitting the two paths.

filters/src/metering/mod.rs
```diff
@@ -0,0 +1,857 @@
...
+///
+/// Identity headers are always marked for removal so tenant identity
+/// never leaks to the upstream provider, regardless of which tier
+/// supplied the identity.
+fn read_identity_headers(ctx: &mut HttpFilterContext<'_>, prefix: &str) -> Identity {
+    let prefix_lower = prefix.to_ascii_lowercase();
+
+    let mut identity = Identity::default();
+    read_metadata_identity(ctx, &prefix_lower, "", &mut identity);
+    let has_verified_identity = !identity.username.is_empty();
```
jordigilh: `has_verified_identity` only checks `username` (L554) -- if tier 1 sets `group`/`subscription`/`model` without `username`, tier 2 unconditionally overwrites those verified fields. Not just theoretical: `jwt_auth` (#769) sets each claim independently, and its `validate_config` doesn't require a username mapping. Worth gating on "any tier-1 field present" instead.

filters/src/metering/mod.rs
```diff
@@ -0,0 +1,857 @@
...
+
+    if !has_verified_identity {
+        read_metadata_identity(ctx, &prefix_lower, "identity.", &mut identity);
```
jordigilh: `"identity."` is hardcoded here, but `identity_header_guard` (#709) makes that namespace configurable (default `"identity"`). Works today only because both default -- if an operator changes it, tier 2 silently stops resolving with no fallback. Worth a matching config field.

filters/src/metering/mod.rs
```diff
@@ -0,0 +1,857 @@
...
+            headers,
+            body: Bytes::from(body),
+        };
+
+        report_delivery(subrequest::execute_url(&client, &url, request, MAX_CALLOUT_RESPONSE_BYTES, timeout).await);
+    });
+}
+
+/// Log the outcome of a usage report delivery.
+fn report_delivery(result: Result<SubResponse, SubRequestError>) {
```
jordigilh, line 809: Only signal on delivery failure is a `debug!` line -- no metric, and `SubRequestClient` doesn't emit one either. A metering blip during a traffic spike would be invisible. A `metrics::counter!` here (already a dependency) would fix that.

filters/src/metering/mod.rs
```diff
@@ -0,0 +1,857 @@
...
+// -----------------------------------------------------------------------------
+// Helpers
+// -----------------------------------------------------------------------------
+
+/// Extract the `model` field from a JSON body fragment without full parsing.
+///
+/// Request bodies arrive in chunks and can be megabytes long, so this scans for
+/// the field instead of buffering and deserializing the whole document. Returns
+/// `None` when the chunk does not contain a complete `"model": "..."` pair.
+fn extract_model_from_bytes(bytes: &[u8]) -> Option<String> {
```
jordigilh, line 912: Substring scan for `"model"` rather than JSON parsing -- a `content` field containing `"model":` ahead of the real field gets misattributed. Confirmed reachable:
```rust
let body = br#"{"messages":[{"role":"user","content":"hi","model":"decoy"}],"model":"gpt-4"}"#;
assert_eq!(extract_model_from_bytes(body).as_deref(), Some("decoy")); // should be "gpt-4"
```
`token_count` already parses this correctly (`serde_json::from_slice`) -- worth reusing that approach.

filters/src/metering/mod.rs
```diff
@@ -0,0 +1,857 @@
+// SPDX-License-Identifier: MIT
+// Copyright (c) 2026 Praxis Contributors
+
+//! External metering filter: pre-request balance checks and post-response
+//! token usage reporting via [`CloudEvents`] to an external metering service.
+//!
+//! Reads token counts from [`filter_metadata`] keys set by the `token_count`
+//! filter (`token.input`, `token.output`, `token.total`, and the prompt cache
+//! breakdown `token.cache_read` / `token.cache_write`). The metering filter
+//! must be declared *before* `token_count` in the YAML filter chain so that
```
jordigilh, line 10: This ordering requirement isn't enforced anywhere -- `praxis-core`'s `FilterPipeline::ordering_errors()` hard-fails at startup, but only checks core's own built-in filter names, not cross-crate `ai` filters. Wrong order -> every event ships with `token: 0`, silently, forever.

jordigilh (commented):
A few things flagged inline, plus some broader ones here. Identity stripping, the CloudEvents payload, and URL-encoding of client-influenced path segments all held up -- no SSRF or credential-leakage concerns.

**Circuit breaking is available but unused for this gateway.** `SubRequestConnector` supports one (`with_options`, praxis#879) and the base proxy wires it up; `praxis-ai`'s server never does (`server.rs#L141`, `commands.rs#L40` both use plain `::new()`). `runtime.subrequest_circuit_breaker` parses fine in an AI-gateway config today and is silently dropped. `external_metering` is the first mandatory-every-request consumer of that shared pool (the other 5 are opportunistic) -- worth a follow-up before a high-traffic rollout.

**`execute_url` (+ client/error types) are now `pub`**, not `pub(crate)` -- needed cross-crate, but worth flagging as a public API addition on `praxis-ai-apis`, semver-relevant going forward.

**Generated docs skip the 3-tier identity design** -- it only lives in a doc comment on the private `read_identity_headers`, which the doc generator never harvests (it reads a second paragraph of the module/struct doc -- `credential_inject.md` already does this). Worth moving it up.

**#577** (this filter's linked design doc) still describes the single-filter/`CalloutClient` design this PR (split across #581 + #709) supersedes.

**Minor:** `iterative_request_router`'s `validate()` doesn't block nesting `external_metering` inside a step (it does block nested IRR and `compression`) -- would double-report with no idempotency key if misconfigured. Not urgent, just flagging.

filters/src/metering/mod.rs
noyitz: @jordigilh good question — checked the service contract: the metering service expresses denial only through a 2xx response carrying `hasAccess: false` (its only non-2xx is a 400 for a malformed path), so a 4xx here always means the service itself is misbehaving, never an intentional denial. 79c3709 documents that contract on `check_balance` and raises the non-2xx log to `warn!` with the status so a misbehaving service is visible. If a future backend wants status-coded denials, splitting the paths would be a config-gated follow-up.

filters/src/metering/mod.rs
noyitz: @jordigilh fixed in 79c3709 — the gate is now "any verified field present" (username, group, subscription, or model), so an auth filter that maps only some claims (exactly the #769 case) blocks the lower tiers entirely. Covered by a new test: a tier-1 group-only identity rejects both a guard-metadata subscription and a raw-header username.

filters/src/metering/mod.rs
noyitz: @jordigilh fixed in 79c3709 — new `identity_metadata_namespace` config field (default `identity`), validated non-empty, with a test resolving tier 2 through a custom namespace. The rustdoc notes it must match `identity_header_guard`'s `metadata_namespace`.

filters/src/metering/mod.rs
noyitz: @jordigilh fixed in 79c3709 — added `praxis_ai_metering_report_failures_total`, incremented on both transport failure and non-2xx acknowledgement, and raised those logs to `warn!`. Naming follows the existing `praxis_ai_ttft_seconds` convention.

filters/src/metering/mod.rs
noyitz: @jordigilh confirmed and fixed in 79c3709 — replaced the substring scan with a string- and depth-aware scanner that only matches a `"model"` key belonging to the top-level object, still without buffering or deserializing the document. Your repro is now a test (`extract_model_ignores_nested_decoy`), plus decoy-inside-string-value and nested-only-decoy cases.

filters/src/metering/mod.rs
noyitz: @jordigilh agreed — core's `ordering_errors()` can't see ai-crate filters, so this is only documented today. Rather than a filter-local workaround, this seems worth a small upstream mechanism (filters declaring ordering constraints the pipeline validates at build). Happy to open a follow-up issue proposing that if you agree.

filters/src/metering/mod.rs
```diff
@@ -0,0 +1,105 @@
...
+
+/// Mark client-supplied credentials for removal.
+///
+/// The proxy authenticates to the provider with its own credentials, so a
+/// client-supplied key is never useful upstream and forwarding one would let a
+/// client bill an account the gateway does not control.
+fn strip_client_credentials(ctx: &mut HttpFilterContext<'_>) {
+    ctx.request_headers_to_remove.push(http::header::AUTHORIZATION);
+
+    if let Ok(name) = "x-api-key".parse::<HeaderName>() {
```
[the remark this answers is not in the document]
noyitz: @praxis-bot fixed in 79c3709 — `HeaderName::from_static("x-api-key")`, compile-time validated, no silently-swallowed branch.

filters/src/metering/config.rs
```diff
@@ -0,0 +1,40 @@
...
+    /// Prefix for tenant identity headers to strip.
+    /// Expected headers: `{prefix}username`, `{prefix}group`,
+    /// `{prefix}subscription`, `{prefix}model`.
+    #[serde(default = "default_identity_header_prefix")]
+    pub identity_header_prefix: String,
+}
+
+/// Validate config at construction time.
+pub(super) fn validate_config(cfg: &ExternalMeteringConfig) -> Result<(), FilterError> {
+    if cfg.identity_header_prefix.is_empty() {
```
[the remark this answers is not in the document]
noyitz: @praxis-bot fixed in 79c3709 — `validate_config` now rejects any prefix that is not a valid HTTP header name via `HeaderName::from_bytes`, with a test for a prefix containing spaces.

filters/src/metering/tests.rs
```diff
@@ -0,0 +1,127 @@
...
+    assert!(removed.contains(&"authorization"));
+    assert!(removed.contains(&"x-api-key"));
+}
+
+#[tokio::test]
+async fn strips_tenant_headers_under_a_custom_prefix() {
+    let mut req = make_request(http::Method::POST, "/v1/chat/completions");
+    req.headers.insert("x-myco-username", "bob".parse().unwrap());
+    let mut ctx = make_filter_context(&req);
+
```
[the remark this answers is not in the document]
noyitz: @praxis-bot added in 79c3709 — `multi_value_identity_header_is_fully_stripped` appends a duplicate `x-tenant-username` and asserts the name lands in `request_headers_to_remove` (removal by name drops every value).

filters/src/metering/mod.rs
```diff
@@ -0,0 +1,105 @@
...
+// -----------------------------------------------------------------------------
+// Header Removal
+// -----------------------------------------------------------------------------
+
+/// Mark every header carrying the tenant identity prefix for removal.
+fn strip_identity_headers(ctx: &mut HttpFilterContext<'_>, prefix: &str) {
+    let prefix_lower = prefix.to_ascii_lowercase();
+
+    for key in ctx.request.headers.keys() {
+        if key.as_str().to_ascii_lowercase().starts_with(prefix_lower.as_str()) {
```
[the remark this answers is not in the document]
noyitz: @praxis-bot fixed in 79c3709 — the prefix is lowercased once at construction and compared against `HeaderName::as_str()` directly; the per-header `to_ascii_lowercase()` allocation is gone.

filters/src/metering/tests.rs
```diff
@@ -0,0 +1,127 @@
...
+// -----------------------------------------------------------------------------
+// Config Parsing
+// -----------------------------------------------------------------------------
+
+#[test]
+fn default_config_parses() {
+    let filter = filter_from_yaml("{}");
+
+    assert_eq!(filter.name(), "external_metering");
+    assert_eq!(filter.identity_header_prefix, "x-tenant-");
```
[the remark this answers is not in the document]
noyitz: @praxis-bot `make lint` (clippy `--workspace --all-targets -D warnings`) is green on this branch, so the workspace deny does not fire on these asserts as written. All tests added since carry messages; happy to sweep the earlier ones too if maintainers want it in this PR.

noyitz:
@jordigilh thanks for the thorough pass — all six inline findings are addressed in 79c3709 (replied on each thread), `make test` 4,913 green, lint green. On the broader notes:

- **Circuit breaker**: agreed this filter is the first mandatory-per-request consumer of the shared pool. Wiring `SubRequestConnector::with_options` through the server (so `runtime.subrequest_circuit_breaker` stops being silently dropped) deserves its own PR — happy to open the follow-up issue.
- **`execute_url` going public**: acknowledged as a deliberate API addition on `praxis-ai-apis`; the alternative was duplicating the URL/DNS/peer-fallback logic in the filters crate.
- **Docs**: the three-tier identity design now lives in the struct documentation, so the generated `external_metering.md` carries it (regenerated in 79c3709).
- **#577**: updating it now to reflect the #581 + #709 split and the CalloutClient → SubRequestClient migration.
- **IRR nesting**: agreed, will fold an `external_metering` guard into the same follow-up as the ordering-validation idea.

filters/src/metering/mod.rs
```diff
@@ -0,0 +1,976 @@
...
+
+/// Upper bound on metering service response bodies.
+///
+/// Balance and event-acknowledgement responses are small JSON documents;
+/// anything larger indicates a misbehaving service and is cut off.
+const MAX_CALLOUT_RESPONSE_BYTES: usize = 64 * 1024;
+
+/// Keepalive pool size for the private per-filter sub-request connector
+/// created by [`ExternalMeteringFilter::from_config`].
+const PRIVATE_POOL_SIZE: usize = 4;
```
tkan145: I think 4 is a bit low? We need to send a report for each request that hit the gateway, which means only 4 connections can be reused simultaneously, so under heavy load, the gateway will need to open new connection each time.

filters/src/metering/mod.rs
```diff
@@ -0,0 +1,976 @@
...
+    .add(b'%')
+    .add(b'/')
+    .add(b':')
+    .add(b'<')
+    .add(b'>')
+    .add(b'?')
+    .add(b'@')
+    .add(b'`')
+    .add(b'{')
+    .add(b'}');
```
tkan145: I think you also need escape query characters `&`, `=`

filters/src/metering/mod.rs
```diff
@@ -0,0 +1,976 @@
...
+    fn report(&self, ctx: &HttpFilterContext<'_>, mut state: MeteringState) {
+        if state.model.is_empty() {
+            if let Some(model) = ctx.filter_metadata.get(META_METERING_MODEL) {
+                state.model.clone_from(model);
+            } else if let Some(fallback) = self.default_model.as_ref() {
+                state.model.clone_from(fallback);
+            }
+        }
+
+        let request_id = ctx.id_generator.generate(ctx.time_source);
```
tkan145: I'm just wondering if it makes more sense to use the actual request_id from the context instead of creating a new request_id here. The reason is that this makes it easier to associate the actual request with the event.

filters/src/metering/mod.rs
```diff
@@ -0,0 +1,976 @@
...
+// -----------------------------------------------------------------------------
+// Usage Reporting (fire-and-forget)
+// -----------------------------------------------------------------------------
+
+/// Deliver an event without blocking the response.
+///
+/// Metering is an observer: a slow or failing metering service must never add
+/// latency to, or fail, a request the upstream already answered.
+fn spawn_usage_report(client: SubRequestClient, metering_url: &str, timeout: Duration, event: &serde_json::Value) {
+    let url = format!("{}/api/v1/events", metering_url.trim_end_matches('/'));
```
tkan145: nit: `metering_url` is fixed; could we construct the final url it at init time instead of generating a new string for each request?

filters/src/metering/mod.rs
jordigilh: Agreed — opened praxis-proxy/ai#812 proposing filters declare ordering constraints that `FilterPipeline` validates at build, generalizing beyond this one identity_header_guard/external_metering pair. The other five fixes in 79c3709 all check out. LGTM on this PR.

jordigilh (approved):
Approving — all six findings resolved: balance-check contract documented, tier-gating fixed, configurable metadata namespace added, failure-reporting metric added, nested-JSON model extraction hardened (79c3709), and the cross-filter ordering-constraint gap now has a tracked follow-up (ai#812).

leseb (approved):
mostly stamping since @jordigilh did some thorough reviews already, thanks!

leseb:
@noyitz see ci failures

filters/src/metering/tests.rs
```diff
@@ -0,0 +1,689 @@
...
+    assert_eq!(filter.default_model.as_deref(), Some("unknown"));
+}
+
+#[test]
+fn config_without_fallbacks_leaves_them_unset() {
+    let yaml: serde_yaml::Value = serde_yaml::from_str(
+        r#"
+metering_url: "http://metering:8080"
+"#,
+    )
```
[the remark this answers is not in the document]
noyitz: Addressed in 0f17568b.

- `default_username` runtime: new `default_username_meters_anonymous_request` — a request with no identity headers and a configured `default_username` stores metering state (balance check pointed at an unroutable URL so it fails instantly; fail-open admits, state is still stored).
- `default_model` runtime: extracted the report-time resolution chain into `resolve_report_model` (captured model, body/header metadata, then configured default — behavior unchanged) and covered it with three unit tests: captured model wins, `default_model` used when nothing was captured, and empty stays empty without a default.

leseb:
@noyitz please rebase

leseb:
@noyitz rebased and fixed conflicts on your behalf.

PR #592, opened 2026-07-28
Fix/external filter discovery 478

### LINKED ISSUE #478: Fix external-filter discovery for optional Praxis AI dependencies
## Parent

Part of #136. This is a prerequisite for the optional local Candle filter, not
for native vLLM passthrough.

## Description

`server/build.rs::collect_server_deps` searches the Cargo resolve graph for
package `praxis-proxy`, but the AI server package is named
`praxis-ai-proxy`. Marked external-filter dependencies are therefore not
discovered and registered, including optional dependencies enabled through a
feature.

## Steps to reproduce

1. Add a runtime dependency carrying `[package.metadata.praxis-filters]`.
2. Put the dependency behind an optional `praxis-ai-proxy` feature.
3. Enable the feature and build `praxis-ai-proxy`.
4. Configure the exported filter.
5. Observe that the filter is unknown because no registration call was
   generated.

## Expected behavior

The build script identifies the AI server package and generates registration
calls for enabled normal runtime dependencies carrying the external-filter
marker. Disabled optional dependencies remain absent.

## Acceptance criteria

1. The package lookup resolves `praxis-ai-proxy` without relying on an
   ambiguous package name.
2. A regression test covers a marked normal runtime dependency.
3. A regression test covers an enabled optional marked dependency.
4. A disabled optional dependency is not registered.
5. Cargo dependency aliases and custom library names remain supported.
6. The generated registry contains the enabled external filter at runtime.
7. Default builds and existing external-filter discovery behavior remain
   compatible.

mkoushni:
## Summary

`praxis-ai-proxy`'s external-filter auto-discovery build script never actually discovered anything: `collect_server_deps` matched packages by the literal name `"praxis-proxy"` to identify the AI server crate, but that string is also the real package name of an unrelated dependency (the workspace alias `praxis = { package = "praxis-proxy" }` for Praxis core). The lookup silently matched the wrong graph node and scanned *its* dependency edges instead of the AI server's, so no external filter dependency — including ones gated behind an optional feature — was ever picked up.

Closes #478

## What changed

- **Root cause fix** — `collect_root_deps` now anchors on `cargo_metadata`'s resolved `Resolve::root` instead of a hardcoded name match, removing the ambiguity entirely. `MetadataCommand` is also given an explicit `manifest_path` rather than relying on the build script's working directory.
- **New `praxis-ai-build-support` crate** (`server/build_support/`) — the dependency-graph and marker-scanning logic moved out of `server/build.rs` into this library crate, since build scripts are never compiled as `cargo test` targets and couldn't carry unit tests. `server/build.rs` is now a thin orchestrator over the build-script environment.
- **Unit tests** (`server/build_support/src/tests.rs`) — exercise discovery against real `cargo metadata` output from on-disk fixture crates, covering normal/renamed marked dependencies, feature-gated optional dependencies, dev-dependency and unmarked dependencies being excluded, and the original collision bug (a marked crate reachable only transitively must not be discovered).
- **End-to-end regression test** (`server/tests/external_filter_discovery_e2e.rs`) — builds an isolated fixture Cargo project exercising the real `praxis-ai-build-support` functions and confirms the generated registration function compiles and registers the filter into a live `praxis_filter::FilterRegistry` at runtime. Skips itself with an explanatory message when the sibling `../praxis` checkout isn't present (e.g. the standard CI unit-test job).
- **Container build fix** — `Containerfile`'s dependency-caching stage now copies `server/build_support`'s manifest and source alongside the other workspace crates.
- `Makefile`'s `test-unit` target now also runs `cargo test -p praxis-ai-build-support`.

## Test plan

- [x] `cargo test -p praxis-ai-build-support`
- [x] `cargo test -p praxis-ai-proxy --test external_filter_discovery_e2e` (passes with `../praxis` present, skips cleanly without it)
- [x] `cargo check --workspace`
- [x] `cargo clippy --all-targets -- -D warnings` / `cargo +nightly fmt --all -- --check`
- [x] `make coverage-check` — 95%+ line coverage

### DIFF
--- a/Containerfile
+++ b/Containerfile
@@ -29,18 +29,24 @@ COPY server/build_support/Cargo.toml ./server/build_support/Cargo.toml
 
 # The server crate has a build.rs that discovers external filter
 # crates via cargo metadata for build-time auto-registration,
-# backed by the praxis-ai-build-support crate.
+# backed by the praxis-ai-build-support crate. build.rs is a
+# build-dependency of the server crate, not an ordinary crate under
+# test: cargo compiles it (and everything it depends on) up front,
+# so praxis-ai-build-support needs its real source here too, not a
+# stub — a stub would compile but export none of the functions
+# build.rs calls, breaking compilation of the build script itself.
 COPY server/build.rs ./server/build.rs
+COPY server/build_support/src ./server/build_support/src
 
 # Strip workspace members not needed for the binary.
 RUN sed -i '/xtask/d; /tests\//d' Cargo.toml
 
-# Create stub source files for all crates.
-RUN mkdir -p apis/src filters/src server/src server/build_support/src \
+# Create stub source files for the crates whose real source isn't
+# needed until after dependencies are cached.
+RUN mkdir -p apis/src filters/src server/src \
     && echo '//! stub' > apis/src/lib.rs \
     && echo '//! stub' > filters/src/lib.rs \
     && echo '//! stub' > server/src/lib.rs \
-    && echo '//! stub' > server/build_support/src/lib.rs \
     && printf '//! stub\nfn main() {}\n' > server/src/main.rs
 
 RUN --mount=type=cache,target=/usr/local/cargo/registry \
@@ -53,12 +59,13 @@ RUN --mount=type=cache,target=/usr/local/cargo/registry \
 
 # Replace stubs with real source, then rebuild. Only the project
 # crates recompile; all external dependencies are cached.
+# build_support/src was already real (see above), so it is not
+# copied again here.
 COPY apis/src ./apis/src
 COPY filters/src ./filters/src
 COPY server/src ./server/src
-COPY server/build_support/src ./server/build_support/src
 
-RUN find apis/src filters/src server/src server/build_support/src \
+RUN find apis/src filters/src server/src \
     -name '*.rs' -exec touch {} +
 
 # ------------------------------------------------------------------------------
--- a/Containerfile
+++ b/Containerfile
@@ -25,19 +25,22 @@ COPY Cargo.toml Cargo.lock ./
 COPY apis/Cargo.toml ./apis/Cargo.toml
 COPY filters/Cargo.toml ./filters/Cargo.toml
 COPY server/Cargo.toml ./server/Cargo.toml
+COPY server/build_support/Cargo.toml ./server/build_support/Cargo.toml
 
 # The server crate has a build.rs that discovers external filter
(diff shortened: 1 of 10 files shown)


Containerfile
```diff
@@ -25,19 +25,22 @@ COPY Cargo.toml Cargo.lock ./
...
 # Strip workspace members not needed for the binary.
 RUN sed -i '/xtask/d; /tests\//d' Cargo.toml
 
 # Create stub source files for all crates.
-RUN mkdir -p apis/src filters/src server/src \
+RUN mkdir -p apis/src filters/src server/src server/build_support/src \
     && echo '//! stub' > apis/src/lib.rs \
     && echo '//! stub' > filters/src/lib.rs \
     && echo '//! stub' > server/src/lib.rs \
+    && echo '//! stub' > server/build_support/src/lib.rs \
```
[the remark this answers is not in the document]
mkoushni: praxis-ai-build-support compiled with real source in the cache-build stage (before praxis-ai-proxy's build.rs needed it)

leseb (approved):
Our repo is the one whose filters need to be discovered by external hosts, not the one doing the discovering but I guess this is fine, thanks for fixing this!

server/build_support/src/lib.rs
```diff
@@ -0,0 +1,241 @@
...
+    let root = resolve.root.as_ref().expect(
+        "cargo metadata returned no root package; verify MetadataCommand is invoked with an \
+         explicit manifest_path and without --no-deps",
+    );
+    resolve
+        .nodes
+        .iter()
+        .find(|node| &node.id == root)
+        .map(|node| node.deps.iter().collect())
+        .unwrap_or_default()
```
aslakknutsen: If `resolve.root` is set but no matching entry exists in `resolve.nodes`, `collect_root_deps` returns an empty vector instead of panicking (`server/build_support/src/lib.rs:100-105`). That mirrors the old failure mode (silent no discovery). Cargo metadata should always include the root node; consider panicking with the same actionable message used elsewhere if the lookup fails.

mkoushni: fixed collect_root_deps in server/build_support/src/lib.rs so it panics with an actionable message if resolve.root doesn't match any node in resolve.nodes, instead of silently falling back to an empty vector via unwrap_or_default(). Added a regression test for that exact case, ran the full build_support test suite (10/10 pass), formatted, and clippy is clean.

leseb (approved):

PR #621, opened 2026-07-30
refactor(validate): make openai_responses_validate request-only

leseb:
### What does this PR do?

Makes `openai_responses_validate` a request-only filter by removing all upstream error interception and response body reconstruction. Backend error responses now pass through the proxy unchanged, preserving the original HTTP status code, content-type header, and response body (including provider-specific fields like numeric `code` and `param`).

**Removed:** `on_response`/`on_response_body` hooks, `BackendError` struct, SSE error synthesis (`build_sse_error_body`, `error_response_object`), JSON error reformatting (`build_json_error_body`, `error_payload`), and all related extraction helpers. Successful-response behavior is unchanged.

### Which issue(s) does this relate to?

N/A

### Checklist

- [x] Signed off all commits (`git commit -s`)
- [x] Tests added or updated
- [x] Documentation updated (if applicable)
- [x] `make lint && make test` passes locally

### Does this introduce a breaking change?

Yes — downstream consumers that relied on the validate filter reformatting backend errors into OpenAI Responses API format (SSE events for `stream: true`, normalized JSON for `stream: false`) will now receive the raw backend error response instead. Clients should handle provider-native error formats directly.


leseb:
openai_responses_validate should validate requests only and leave upstream responses untouched.

franciscojavierarceo (approved):

Issue #541, opened 2026-07-27
Add vLLM integration tests on GPU and expand CPU coverage


leseb:
## Summary

Expand vLLM integration testing to cover GPU instances alongside the existing CPU-only setup (`vllm-integration.yaml`), validating streaming, routing, and response handling under hardware acceleration.

## Current state

A single CI workflow runs vLLM on CPU (`quay.io/opendatahub/vllm-cpu`, Qwen3-0.6B) and exercises Python SDK tests against a live Praxis + vLLM stack.

## Scope

- Configure a GPU runner (self-hosted or cloud) with a CUDA-enabled vLLM image
- Add a GPU integration workflow (or extend the existing one)
- Expand test coverage for both CPU and GPU paths
- Decide on trigger strategy: per-PR vs. scheduled/manual to manage cost

leseb:
@skamenan7 PTAL

skamenan7:
@leseb please assign it to me. Thanks.

PR #640, opened 2026-07-31
ci: add vLLM GPU hello-world smoke (T4 / Qwen3-0.6B)

skamenan7:
## Summary
- Adds a `workflow_dispatch` GPU smoke workflow for Praxis issue #541
- Runs the existing `test_stateless_request` HELLO-PRAXIS check against CUDA vLLM (`Qwen/Qwen3-0.6B`, max 4096 tokens)
- Defaults to Ken’s offered `gpu-t4-4-core` larger-runner label; runner label is overridable
- Avoids the OGX `ec2-github-runner` / personal-PAT pattern

## Test plan
- [ ] Confirm `gpu-t4-4-core` (or equivalent) exists on `praxis-proxy`
- [ ] Actions → **vLLM GPU Smoke** → Run workflow
- [ ] Job shows `nvidia-smi`, vLLM becomes healthy, `test_stateless_request` passes
- [ ] Failure path uploads vLLM/OGX logs


skamenan7:
Dispatched **vLLM GPU Smoke** on this branch (`runner=gpu-t4-4-core`).

The job stayed **queued with no runner picked up**, so `praxis-proxy` does not appear to have a `gpu-t4-4-core` larger-runner yet (Ken offered that shape from another org).

Once that runner exists on this org (or we get the exact label), re-run:
Actions → **vLLM GPU Smoke** → `gpu-t4-4-core` (or override).

Smoke itself is the existing `test_stateless_request` / `HELLO-PRAXIS` check against CUDA vLLM `Qwen/Qwen3-0.6B`.

.github/workflows/vllm-gpu-smoke.yaml
```diff
@@ -0,0 +1,141 @@
+name: vLLM GPU Smoke
+
+# Manual GPU hello-world against Qwen/Qwen3-0.6B on a T4-class runner.
+# Intended for Ken's offered `gpu-t4-4-core` larger-runner shape (or equivalent).
+# Does not use the OGX ec2-github-runner / personal-PAT pattern.
+
+on:
+  workflow_dispatch:
```
nathan-weinberg, line 8: Is this just for testing? Presume we will want to also run this on a schedule/PRs/etc.

.github/workflows/vllm-gpu-smoke.yaml
```diff
@@ -0,0 +1,141 @@
...
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
+permissions: {}
+
+env:
+  CARGO_TERM_COLOR: always
+  VLLM_IMAGE: "docker.io/vllm/vllm-openai:v0.22.1-cu129@sha256:e1668bce9790a4b86682f8fcc99678153a13e12dc70e05348d8e239ffa474b05"
```
nathan-weinberg, line 24: @leseb @franciscojavierarceo hey guys, @skamenan7 and I were talking and wanted to loop y'all in on an idea

For the OGX distro, we have automation in place in the repo for building and pushing our own vLLM CPU images with baked-in models, including Dependabot configurations for automatically updating the vLLM base image via a PR. This prevents us from being reliant on HuggingFace credentials/uptime for our testing.

I could replicate those workflows here or in another repo somewhere in Praxis if we were interested in using the same strategy for the testing here, but with CUDA-based GPU images rather than CPU

Let me know your thoughts - if y'all agree I can open a new tracker issue and assign myself for this work - it shouldn't be a heavy lift

.github/workflows/vllm-gpu-smoke.yaml
```diff
@@ -0,0 +1,141 @@
...
+    runs-on: ${{ inputs.runner }}
+    timeout-minutes: 45
+    permissions:
+      contents: read
+    steps:
+      - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
+
+      - name: Verify GPU and container runtime
+        run: |
+          nvidia-smi
```
nathan-weinberg, line 38: Will the GHA run fail if `nvidia-smi` exits non-zero?

.github/workflows/vllm-gpu-smoke.yaml
```diff
@@ -0,0 +1,141 @@
...
+              -p 8000:8000 \
+              -e HUGGING_FACE_HUB_TOKEN="${{ secrets.HUGGING_FACE_HUB_TOKEN }}" \
+              "$VLLM_IMAGE" \
+              --model "$VLLM_MODEL" \
+              --max-model-len 4096 \
+              --served-model-name "$VLLM_MODEL" \
+              --enable-auto-tool-choice \
+              --tool-call-parser hermes \
+              --reasoning-parser deepseek_r1 \
+              --gpu-memory-utilization 0.7
```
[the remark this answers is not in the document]
nathan-weinberg: Agreed!

.github/workflows/vllm-gpu-smoke.yaml
nathan-weinberg: A PoC for this idea can be found here: https://github.com/praxis-proxy/ai/tree/vllm-gpu

.github/workflows/vllm-gpu-smoke.yaml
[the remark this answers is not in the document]
skamenan7: Agreed. I will move the runtime-specific GPU argument into the detection step and collapse this to one container command so the vLLM options only live in one place.

.github/workflows/vllm-gpu-smoke.yaml
skamenan7: Yes. GitHub Actions invokes bash with `-e -o pipefail` by default, so a non-zero exit from `nvidia-smi` stops the step and fails the job before the container starts.

.github/workflows/vllm-gpu-smoke.yaml
```diff
@@ -0,0 +1,128 @@
...
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
+permissions: {}
+
+env:
+  CARGO_TERM_COLOR: always
+  VLLM_IMAGE: "docker.io/vllm/vllm-openai:v0.22.1-cu129@sha256:e1668bce9790a4b86682f8fcc99678153a13e12dc70e05348d8e239ffa474b05"
+  VLLM_MODEL: "Qwen/Qwen3-0.6B"
```
leseb, line 25: Can't we use a more capable model?

.github/workflows/vllm-gpu-smoke.yaml
```diff
@@ -0,0 +1,128 @@
...
+            "$VLLM_IMAGE" \
+            --model "$VLLM_MODEL" \
+            --max-model-len 4096 \
+            --served-model-name "$VLLM_MODEL" \
+            --enable-auto-tool-choice \
+            --tool-call-parser hermes \
+            --reasoning-parser deepseek_r1 \
+            --gpu-memory-utilization 0.7
+
+      - name: Start OGX
```
leseb, line 77: why?

.github/workflows/vllm-gpu-smoke.yaml
leseb: That's a good idea, once this merges.

leseb (commented):
A few thoughts:

* We should use a more capable model, otherwise the value is low
* We need to use this instance with the existing vLLM integration test

.github/workflows/vllm-gpu-smoke.yaml
nathan-weinberg: Perhaps [Qwen/Qwen3.5-9B](https://huggingface.co/Qwen/Qwen3.5-9B)?

.github/workflows/vllm-gpu-smoke.yaml
nathan-weinberg: Opened an issue to track and assigned myself: https://github.com/praxis-proxy/ai/issues/692

leseb:
@skamenan7 what's the status on this one? Thanks

Issue #74, opened 2026-04-30
Epic: Intelligent Router


shaneutt:
As organizations deploy multiple specialized models, intelligent routing that considers model capability, cost, latency, and availability across providers and clusters becomes critical. This is a networking problem at its core: the proxy must understand model topology and make routing decisions based on request semantics, model performance characteristics, and business policy.

This epic tracks all work needed to handle intelligent routing to inference providers (and potentially agents and tools) to support robust mixture-of-models routing capabilities.

## Existing Praxis Primitives

These capabilities already exist and form the foundation:

- **StreamBuffer**: peek-then-stream body inspection for extracting routing signals
- **model_to_header / json_body_field**: JSON field extraction and header promotion
- **json_rpc**: JSON-RPC 2.0 parsing for MCP/A2A method routing
- **router**: path-prefix and host-based routing with header matching
- **load_balancer**: round-robin, least-connections, consistent-hash with weights
- **branch_chains**: conditional execution for fallback, escalation, and skip-to patterns
- **circuit_breaker**: circuit breaker pattern
- **guardrails**: string/regex matching on headers and body
- **credential_injection**: per-cluster API key injection with credential stripping
- **rate_limit**: token bucket with per-IP and global modes

---

### Upstream Targets

- Multi-cluster inference (self-hosted vLLM, SGLang, Triton, Ollama, etc.)
- 3rd party AI services (OpenAI, Anthropic, Google, Azure, Bedrock, etc.)
- Multi-cluster agents / agent mesh (MCP, A2A)
- 3rd party tool and agent services
- Mixtures of all of the above

### Content-Based Routing

Extract routing signals from request bodies using StreamBuffer.

- [ ] `model` field extraction and header promotion (exists: `model_to_header`, `json_body_field`)
- [ ] `stream` field detection (streaming vs non-streaming request routing)
- [ ] `tools` / function calling detection for tool-capable model routing
- [ ] Prompt length estimation for context window overflow routing (reroute to larger-context model)
- [ ] Multimodal content detection (image/audio URLs or base64 in messages)
- [ ] JSON-RPC method routing for MCP/A2A (exists: `json_rpc` filter)

### Semantic Routing

Classify prompts by content/intent via embedding similarity and route to the best-fit model.

- [ ] Embedding-based prompt classification (~100ms latency vs ~5s for LLM classification)
- [ ] Route definitions with seed utterances (similar to Aurelio Semantic Router)
[...]

PR #643, opened 2026-08-02
Add a proposal for generic model routing support.

usize:
## Summary

Model selection, for Mixture of Models, may incorporate many approaches--complexity analysis, semantic routing, etc...--from many sources. Meanwhile model routing hygeine and policy enforcement is needed across all possible approaches.

This proposal aims to define an API contract, allowing model selection approaches to plug and play with Praxis's Mixture of Models support.

That is, a model or list of models is proposed and the work of mutating a request or selecting a corresponding cluster to route to is handled downstream.

## Related issue

#74


usize:
This proposal doesn't include a "How" section, so there's no shape to the API contract that's suggested.

Here are some sketches of what sorts of policies we'd want to support, that might help guide the way:

https://gist.github.com/usize/6925f78eca5b2291caac651c7ccf8eb6

Issue #191, opened 2026-06-22
Agent budget enforcement


shaneutt:
Per-agent cost tracking and enforcement.

Budget limits cascade (organization → team → user → agent). When a budget is exhausted, the enforcer constrains the backend scorer to cheaper tiers rather than rejecting the request outright. Budget state integrates with the orchestration layer's G-Counter CRDTs for cross-grid visibility. Gateway-local tracking for single-site deployments.

jordigilh:
This overlaps #121's epic — both implement cascading org→team→user[→agent] budget limits — and neither issue references the other. Worth deciding whether this folds into #121 as its "agent" tier, or stays separate, since the enforcement action differs (#121 returns 429; this degrades to a cheaper model instead). Flagging before either gets implementation effort.

szedan-rh:
Following up on the overlap flag: #78 has now been merged into #121 (**Epic: Token Rate Limiting and Budgets**), so this and #125 sit under the same epic. The proposed resolution recorded there: #191 is the "agent" tier of #125's org → team → user cascade plus a `degrade` outcome. Whether `degrade` becomes a third tier `action.type` next to `deny` / `inject`, or stays `inject` with a downstream router reading the header, is listed as an open decision on the epic.

grokoloko:
Relevant to **agent budget enforcement**: Enforcement at the tool/call boundary is where budgets become real — pair it with transferable prepaid hours for burst Pix when a branch needs a short inference spike.

Pattern we use on agent fleets: attach a hard per-task hour/token ceiling to the task envelope, fail closed when empty, and treat leftover prepaid hours as inventory between branches. Uneven runs leave idle inference on one side and a short Pix burst need on the other — transferable hours beat locking another opaque monthly API plan overnight.

- Buy leftover hours: https://sparetoken.shop/?utm_source=github&utm_medium=comment&utm_campaign=sell&utm_content=buy-20260917-gh5
- Sell leftover hours: https://sparetoken.shop/?utm_source=github&utm_medium=comment&utm_campaign=sell&utm_content=sell-20260917-gh5

grokoloko:
Relevant to **agent budget enforcement**: Enforcement at the tool/call boundary is where budgets become real — pair it with transferable prepaid hours for burst Pix when a branch needs a short inference spike.

[...]

Issue #656, opened 2026-08-06
Epic: Tokenomics


shaneutt:
### Summary

End-to-end token awareness for praxis-ai: counting, metering, rate limiting, and cost attribution across all supported providers.

This epic covers the full lifecycle of token accounting - from extracting usage out of provider responses, through enforcing token-denominated quotas, to feeding accurate consumption data into billing and metering systems.

### Success Criteria

- Token Budgets
- Token counts extracted from all supported providers (streaming and non-streaming)
- Token count estimations
- Token-denominated rate limiting with reserve-then-reconcile admission
- Accurate metering data available for billing + charge-back
- Cross-cluster token quota enforcement

Issue #549, opened 2026-07-27
MCP tool expansion bypasses the configured request-body limit


leseb:
## Clawpatch finding

- **Severity:** medium
- **Category:** performance
- **Confidence:** high
- **Triage:** risk
- **Feature:** `feat_custom_openai_responses_tools_io`
- **Finding ID:** `fnd_sig-feat-custom-openai-responses_72b6e965c9`
- **Run ID:** `20260727T064345-425d3d`
- **Prior-run overlap:**
No prior-run overlap detected.

## Evidence

- `apis/src/openai/responses/openai_mcp_tool_resolve/config.rs:29-47` (`McpToolResolveConfig`) — max_body_bytes is the configured request-body bound.
- `apis/src/openai/responses/openai_mcp_tool_resolve/mod.rs:232-240` (`McpToolResolveFilter::request_body_mode`) — The bound is applied only while buffering the original client body.
- `apis/src/openai/responses/openai_mcp_tool_resolve/mod.rs:180-186` (`McpToolResolveFilter::resolve_all_entries`) — Opaque MCP definitions are expanded into function tools, including their schemas.
- `apis/src/openai/responses/openai_mcp_tool_resolve/mod.rs:389-430` (`rewrite_request_body`) — The expanded JSON is serialized and installed without comparing its length to max_body_bytes.

## Description

A small incoming request can expand substantially when external MCP tool descriptions and input schemas are inserted. The filter calculates the rewritten length only for Content-Length and forwards it even when it exceeds max_body_bytes, allowing an untrusted MCP server or cached listing to defeat the configured body cap and create large allocations or upstream request failures.

## Reproduction

Configure max_body_bytes to a small value, provide an incoming MCP request below that limit, and satisfy resolution from ResponsesState.previous_tools with a tool containing a sufficiently large description or input schema. The filter returns Continue and replaces the body with serialized JSON larger than max_body_bytes.

## Recommendation

Pass max_body_bytes into rewrite_request_body, reject the expanded request with a Responses-compatible 413 before replacing the body, and ideally enforce a cumulative byte limit while collecting MCP tool definitions to avoid building an oversized intermediate representation.

## Notes

Generated by Clawpatch review over `apis/src` excluding `apis/src/token_usage/**` after rebasing on `origin/main`. Please validate before assigning or starting implementation.

leseb:
This issue requires human review before any work can begin. **Do not self-assign** and **Do not submit a PR** before the issue has been validated.

Issue #547, opened 2026-07-27
Conversation pagination accepts invalid query values as successful defaults


leseb:
## Clawpatch finding

- **Severity:** medium
- **Category:** api-contract
- **Confidence:** high
- **Triage:** contract-mismatch
- **Feature:** `feat_custom_openai_core_protocol`
- **Finding ID:** `fnd_sig-feat-custom-openai-core-prot_fe748826f0`
- **Run ID:** `20260727T064001-aebbb7`
- **Prior-run overlap:**
No prior-run overlap detected.

## Evidence

- `apis/src/openai/conversations/handlers.rs:58-62` (`ItemListParams::effective_limit`) — self.limit.clamp(1, MAX_PAGE_LIMIT)
- `apis/src/openai/conversations/handlers.rs:621-650` (`parse_item_list_params`) — if let Ok(n) = value.parse::<u32>() { params.limit = n; }
- `apis/src/openai/conversations/handlers.rs:836-846` (`parse_params_unknown_order_stays_default`) — unknown order should keep default descending

## Description

Malformed `limit` values silently use the default, out-of-range values are clamped, and unknown `order` values silently become descending. Because this filter locally implements the endpoint rather than forwarding to an OpenAI backend, it must enforce the endpoint's integer and enum contract instead of converting invalid requests into successful but different queries; the official operation defines integer `limit` and `order` restricted to `asc` or `desc`.

## Reproduction

Request `GET /v1/conversations/conv_x/items?limit=abc&order=random`. The parser returns the default page limit and descending order, and the handler responds with a normal 200 list instead of an invalid-request response.

## Recommendation

Make query parsing fallible. Reject malformed, ~zero~, oversized, duplicate, or invalid-enum parameters with the standard 400 `invalid_request_error` response rather than defaulting or clamping.

## Notes

Generated by Clawpatch review over `apis/src` excluding `apis/src/token_usage/**` after rebasing on `origin/main`. Please validate before assigning or starting implementation.

leseb:
This issue requires human review before any work can begin. **Do not self-assign** and **Do not submit a PR** before the issue has been validated.

leseb:
As of 2026-07-27, OpenAI returned (live testing):

- Duplicate limit, order, or after → 400, code duplicate_parameter
- limit=abc → 400, code invalid_type
- limit=101 → 400, code integer_above_max_value
- order=random → 400, code invalid_value
[...]

PR #658, opened 2026-08-06
docs(proposal): add implementation to token rate-limiting

shaneutt:
This adds a high-level overview of the strategy for implementing Token Rate Limiting in Praxis.

### DIFF
--- a/docs/proposals/00121_token-rate-limiting.md
+++ b/docs/proposals/00121_token-rate-limiting.md
@@ -4,9 +4,11 @@ discussion: https://github.com/praxis-proxy/ai/issues/121
 status: proposed
 authors:
   - shaneutt
+  - jland-redhat
 graduation_criteria:
   - How? section with requirements and design
 stakeholders:
+  - jland-redhat
   - leseb
   - mkoushni
   - crstrn13
@@ -14,7 +16,7 @@ stakeholders:
   - alexsnaps
 ---
 
-# Token Rate Limiting
+# Tokenomics + Token Rate Limiting
 
 ## What?
 
@@ -26,18 +28,19 @@ wildly in terms of token cost. This capability lets
 operators define budgets in the unit that actually
 drives inference cost.
 
-The system requires reservation-based admission: requests
-are admitted by reserving an estimated cost up front,
-then the reservation is reconciled against actual usage
-reported by the provider after the response completes.
-The estimation method must be configurable because
-different deployments have different cost models.
+The system requires reservation-based admission:
+requests are admitted by reserving an estimated cost up
+front, then the reservation is reconciled against
+actual usage reported by the provider after the
+response completes. The estimation method must be
+configurable because different deployments have
+different cost models.
 
 Different token types must be accounted for separately
 with configurable weights. A cached input token does
 not carry the same cost as an uncached token, and
-quotas should reflect that difference. We also need
-to support multiple modular strategies for what to do
+quotas should reflect that difference. We also need to
+support multiple modular strategies for what to do
 about input token counting, output token counting,
 estimation, what to do with unused tokens, etc.
 
@@ -45,45 +48,61 @@ estimation, what to do with unused tokens, etc.
 
 **Must have (MVP):**
 
-- **[M1]** Bucketed `Application` concept which applies
-  `TokenBudget` to a specific application (or, by default
-  treats all requests as one `default` application.
-- **[M1]** Reservation-based admission: admit requests
-  by reserving an estimated token cost, then reconcile
-  against actual provider-reported usage after the
+- **[M1]** Bucket `Rules` which apply a `TokenBudget`
+  based on a specific condition such as a header (or,
+  by default, treats all requests under a `default`
+  rule).
+- **[M2]** Reservation-based admission: admit requests
+  while reserving an estimated token cost, then
+  reconcile count based on usage retrieved from the
   response.
-- **[M2]** Configurable estimation: how estimated cost
-  is calculated from request metadata must be
-  operator-defined, not a single hard-coded formula.
-- **[M3]** Token-type-aware accounting: different token
-  types (input, output, cached, uncached) tracked
-  separately with configurable weights so quotas
-  reflect real cost differences.
-- **[M4]** Flexible bucket keys: quotas keyed by
-  request headers, model identity, or compound keys
-  so different clients and models get independent
-  budgets. (TBD - this one may need to be teased out more)
-- **[M5]** Hard deny with 429 when a budget is
-  exhausted, with standard rate limit response
-  headers (`Retry-After`, `X-RateLimit-*`).
+- **[M3]** Configurable estimation: allow operator to
+  define how cost is estimated based on request
+  metadata.
+- **[M4]** Token-type-aware accounting: a flexible way
+  to capture different token types (input, output,
+  cached, thinking) from different providers. Each
+  tracked separately with configurable weights so
+  quotas reflect real cost differences.
+- **[M5]** Flexible bucket keys: quotas keyed by
+  request information. Headers, model identity, or
+  compound keys so different clients and models get
+  independent budgets. (TBD - may need further
+  scoping; see upstream [wg-ai-gateway-keys] effort
+  and related issues #123, #129, #232.)
+- **[M6]** Hard deny with 429 when a budget is
+  exhausted, with standard rate limit response headers
+  (`Retry-After`, `X-RateLimit-*`).
 - **[M7]** Observability: metrics and tracing
   distinguishing admitted vs. limited requests,
-  including estimated cost, actual cost, and
-  remaining budget. The ability to log tokenomic
-  results distinctly for accounting.
-- **[M7]** Soft limits: usage tiers that modify request
+  including estimated cost, actual cost, and remaining
+  budget. The ability to log tokenomic results
+  distinctly for accounting.
+
+
+**Should have (same effort if capacity):**
+
+- **[S0]** Multi-environment observability: aggregate
+  token counts from proxies spanning multiple
+  environments into one centralized view. (Moved from
+  MVP; requires its own design proposal covering
+  distributed counter replication -- see [#155] and
+  [ai#126].)
+
+- **[S1]** Soft limits: usage tiers that modify request
   headers instead of rejecting, enabling downstream
-  systems (e.g. llm-d `InferenceObjective`) to
-  degrade gracefully as a client approaches its
-  budget.
-- **[M8]** Batch workload awareness: separate quota
+  systems (e.g. llm-d `InferenceObjective`) to degrade
+  gracefully as a client approaches its budget.
+  (Related: #856 soft rate limiting, #549 quota
+  exhaustion failover.)
+- **[S2]** Batch workload awareness: separate quota
   rules or deferred accounting for batch/async API
   patterns so bulk jobs do not starve interactive
   traffic.
-- **[M9]** Exact token metering: an accounting path
+- **[S3]** Exact token metering: an accounting path
   that records precise actual-usage counts for billing
   and chargeback, independent of rate-limit counters.
-- **[M10]** Reservation refund on lost requests: release
+- **[S4]** Reservation refund on lost requests: release
   reserved capacity when a request times out, is
   dropped, or otherwise never completes.
 
@@ -97,61 +116,82 @@ estimation, what to do with unused tokens, etc.
   header by an upstream component. We can re-assess
   needs around this in later iterations.
 
-These are non-goals for this _iteration_ but are otherwise
-long term capabilities we do want.
+These are non-goals for this _iteration_ but are
+otherwise long term capabilities we do want.
 
-- Usage-based dynamic rates: adjust quotas or
-  refill rates based on observed usage patterns over
-  a sliding window.
+- Usage-based dynamic rates: adjust quotas or refill
+  rates based on observed usage patterns over a sliding
+  window.
 - Input message tokenization: fully tokenize request
   content before admission to use a real input token
-  count rather than relying on estimations, `max_tokens`
-  as a proxy, etc.
+  count rather than relying on estimations,
+  `max_tokens` as a proxy, etc.
+- Token Rate Limiting over a static window
+
+### Prior Art
+
+- **Kubernetes AI Gateway WG**: The upstream
+  [wg-ai-gateway] working group is exploring a
+  rate-limiting standard (see
+  [wg-ai-gateway#60][wg-ai-gateway-rl]). This proposal
+  should track that effort for compatibility where
+  feasible. A related key-calculation discussion is in
+  [wg-ai-gateway#57][wg-ai-gateway-keys].
+- **Praxis request-count rate limiting**: Praxis
+  already supports request-count rate limiting via the
+  `rate_limit` filter. Token rate limiting is a
+  complementary capability, not a replacement.
+
+[wg-ai-gateway]: https://github.com/kubernetes-sigs/wg-ai-gateway
+[wg-ai-gateway-rl]: https://github.com/kubernetes-sigs/wg-ai-gateway/pull/60
+[wg-ai-gateway-keys]: https://github.com/kubernetes-sigs/wg-ai-gateway/pull/57
 
 ## Why?
 
 ### Motivation
 
-AI inference is the first workload where request-count
-rate limiting is meaningfully wrong. Inference cost
-scales with token count, and a single request can vary
-by orders of magnitude. An operator limiting a team to
-100 req/min has no control over whether those requests
-consume 1,000 tokens or 10,000,000.
+AI inference is a unique type of workload where
+request-count rate limiting is meaningfully wrong.
+Inference cost scales with token count, and a single
+request can vary by orders of magnitude. An operator
+limiting a team to 100 req/min has no control over
+whether those requests consume 1,000 tokens or
+10,000,000.
 
 Three realities shape the requirements:
 
-1. **Precise token counts are _generally_ only available
-   after the response.** Providers report usage in the
-   response body or headers. By the time actual counts
-   are known, the tokens have been consumed. Admission
-   decisions must therefore rely on estimates, with
+1. **Precise token counts are only available after the
+   response.** Providers are unaware of output token
+   counts before receiving the output and are therefore
+   forced to report usage in the response body or
+   headers. By the time actual counts are known, the
+   tokens have been consumed. Admission decisions
+   however must happen at request time, and therefore
+   we must rely on token count estimates, with
    reconciliation after the fact. This is why a
    reservation-based model is necessary rather than
-   simple post-hoc accounting. (Caveat: in the future we
-   may be able to eliminate this by running tokenizers
-   early).
+   simple post-hoc accounting.
 
 2. **Not all tokens cost the same.** Providers charge
    significantly less for cached input tokens than
-   uncached (roughly 10x for Anthropic). A quota
-   system that treats all tokens equally will
-   over-restrict users who benefit from caching or
-   under-restrict those who do not. Token-type-aware
-   weighting is essential for fair quotas.
+   uncached (roughly 10x for Anthropic). A quota system
+   that treats all tokens equally will over-restrict
+   users who benefit from caching or under-restrict
+   those who do not. Token-type-aware weighting is
+   essential for fair quotas.
 
 3. **AI workloads span clusters.** Enterprise
    deployments run multiple proxy instances across
    availability zones. Per-instance budgets cause
-   effective quota to scale with instance count,
-   the opposite of the intended control.
+   effective quota to scale with instance count, the
+   opposite of the intended control.
 
 Beyond hard enforcement, operators need graduated
 controls. When a team approaches its budget the
-platform should be able to signal downstream
-schedulers to route to cheaper models or lower-priority
-queues, rather than rejecting outright. Hard deny is
-the backstop, not the first line of defense.
+platform should be able to signal downstream schedulers
+to route to cheaper models or lower-priority queues,
+rather than rejecting outright. Hard deny is the
+backstop, not the first line of defense.
 
 ### User Stories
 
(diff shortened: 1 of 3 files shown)


docs/proposals/00121_token-rate-limiting.md
```diff
@@ -184,41 +188,282 @@ the backstop, not the first line of defense.
   I need batch calls accounted for without starving
   real-time traffic.
 
+## How?
```
jland-redhat, line 226: Still not complete but the upstream AI Gateway WG is trying to look into creating a RateLimiting standard if we want to take inspiration from that?

https://github.com/kubernetes-sigs/wg-ai-gateway/pull/60

docs/proposals/00121_token-rate-limiting.md
```diff
@@ -26,64 +28,76 @@ wildly in terms of token cost. This capability lets
...
-- **[M1]** Bucketed `Application` concept which applies
-  `TokenBudget` to a specific application (or, by default
-  treats all requests as one `default` application.
-- **[M1]** Reservation-based admission: admit requests
-  by reserving an estimated token cost, then reconcile
-  against actual provider-reported usage after the
+- **[M1]** Bucket `Rules` which apply a `TokenBudget`
+  based on a specific condition such as a header (or,
+  by default, treats all requests under a `default`
+  rule).
```
alexsnaps, line 54: I'd suggest avoiding having "a default". Most generally buckets "stack", e.g. with hourly & daily counters when any hits the threshold, the call is considered "above budget". While it seems _easy enough_ to say "if there are no bucket rule matching for this request, then delegate to this other _set of buckets_", it actually comes at a very high cost in terms of complexity. otoh, if the user is to specify "how to qualify" the default, that complexity, while being pushed on the user, is gone and makes for a system that much easier to reason about and less prone to bugs & weird corner cases.

docs/proposals/00121_token-rate-limiting.md
```diff
@@ -26,64 +28,76 @@ wildly in terms of token cost. This capability lets
...
-  separately with configurable weights so quotas
-  reflect real cost differences.
-- **[M4]** Flexible bucket keys: quotas keyed by
-  request headers, model identity, or compound keys
-  so different clients and models get independent
-  budgets. (TBD - this one may need to be teased out more)
-- **[M5]** Hard deny with 429 when a budget is
-  exhausted, with standard rate limit response
-  headers (`Retry-After`, `X-RateLimit-*`).
+- **[M3]** Configurable estimation: allow opeartor to
```
alexsnaps: `operator` (not `opeartor`)

docs/proposals/00121_token-rate-limiting.md
```diff
@@ -26,64 +28,76 @@ wildly in terms of token cost. This capability lets
...
+  define how cost is estimated based on request
+  metadata.
+- **[M4]** Token-type-aware accounting: a flexible way
+  to capture different token types (input, output,
+  cached, thinking) from different providers. Each
+  tracked separately with configurable weights so
+  quotas reflect real cost differences.
+- **[M5]** Flexible bucket keys: quotas keyed by
+  request infromation. Headers, model identity, or
```
alexsnaps: `information` (not `infromation`)

docs/proposals/00121_token-rate-limiting.md
```diff
@@ -26,64 +28,76 @@ wildly in terms of token cost. This capability lets
...
+  (`Retry-After`, `X-RateLimit-*`).
 - **[M7]** Observability: metrics and tracing
   distinguishing admitted vs. limited requests,
-  including estimated cost, actual cost, and
-  remaining budget. The ability to log tokenomic
-  results distinctly for accounting.
-- **[M7]** Soft limits: usage tiers that modify request
+  including estimated cost, actual cost, and remaining
+  budget. The ability to log tokenomic results
+  distinctly for accounting.
```
alexsnaps, line 80: I'd suggest _not_ to compound "accounting" & TRLP. This makes for very brittle systems, mostly in the face of failure. It'd be preferable to have that (configurable?) auditing trail happen "outside this system", while it probably could be reusing some of the information from TRLP (e.g. the "bucket keying logic").

Now it can be part of this problem definition, but it would be desirable to model that separately in the implementation.

docs/proposals/00121_token-rate-limiting.md
```diff
@@ -26,64 +28,76 @@ wildly in terms of token cost. This capability lets
...
+  cached, thinking) from different providers. Each
+  tracked separately with configurable weights so
+  quotas reflect real cost differences.
+- **[M5]** Flexible bucket keys: quotas keyed by
+  request infromation. Headers, model identity, or
+  compound keys so different clients and models get
+  independent budgets. (TBD - this one may need to be
+  teased out more - possibly using CEL note there is an
+  upstream effor around  this found
+  [here](https://github.com/kubernetes-sigs/wg-ai-gateway/pull/57))
```
alexsnaps: This problem is interesting. In my experience, a key is "calculated" (vs. being a compound key, where all individual bits remains usable by themselves) - which is another reason to _not_ compound the rate limiting with the auditing (see below).

docs/proposals/00121_token-rate-limiting.md
```diff
@@ -167,58 +184,393 @@ the backstop, not the first line of defense.
...
+
+### Design
+
+#### Token Budgeting
+
+Define a set of `Rules` based on traffic information
+(headers, model, path). Each rule binds one or more
+`token_budgets` (for example an hourly budget and a
+daily budget). Unmatched requests fall to a `default`
+rule.
```
alexsnaps, line 252: Repeating myself, but I'd suggest leaving that "fall to a `default` rule" out, unless there is an actual real use case for it.

docs/proposals/00121_token-rate-limiting.md
```diff
@@ -167,58 +184,393 @@ the backstop, not the first line of defense.
...
+          - capacity: 100_000
+            action: deny
+      - window: 24h
+        tiers:
+          - capacity: 800_000
+            inject:
+              headers:
+                X-Token-Day-Tier: warning
+          - capacity: 1_000_000
+            action: deny
```
alexsnaps: How is `inject`, _not_ an `action` here? nitpick, but from a modeling/normalization perspective I think we can do better here by having a "tighter API surface", which then would probably also be more open to extensions in the future, wdyt?

docs/proposals/00121_token-rate-limiting.md
```diff
@@ -167,58 +184,393 @@ the backstop, not the first line of defense.
...
+   estimated and actual cost is logged, and usage past
+   tier capacities is still tracked for observability
+   when no `deny` tier applies. (In future iterations,
+   we may refund the estimate-vs-actual difference back
+   to the budget; for MVP, budgets consume at the
+   estimated rate.)
+
+4. **Cleanup** - If a request is lost (timeout,
+   connection reset, upstream failure), release the
+   reservation after a configurable hold period.
```
alexsnaps: If we want, which I think we do and certainly recommend so, to be able to deal with all failure scenarios, this flow will probably need to be somewhat different:

- Reservations need to be first class citizens in the `admission` phase and have their own timeout
- `Reconciliation` needs to deal with "applying the actual diff" using the reservation, where the "refund" is actually the easiest problem, overshoots are more annoying - as users may want different outcomes.
- There is a note to be added here possibly, about how a reservation behaves when it "overlaps two (or more buckets)". Not a huge problem, but probably clearly explaining what happens would probably be good.
- `Cleanup` here I _think_ is what the proxy does... but the proxy can only do it _if it's still around_ (see first bullet point above)

docs/proposals/00121_token-rate-limiting.md
```diff
@@ -167,58 +184,393 @@ the backstop, not the first line of defense.
...
+| `input_plus_max_tokens` | content size + `max_tokens` | Conservative full-cost |
+| `fixed` | constant per request | Uniform cost model |
+| `model_scaled` | `max_tokens` * model multiplier | Model-aware budgets |
+
+```yaml
+rules:
+  - name: team-alpha
+    estimation:
+      strategy: input_plus_max_tokens
+      multiplier: 1.2
```
alexsnaps, line 425: Unless there is a reason _not_ to support an expression here, I'd stay away from these "explicit YAML" strategies. Now, if on the other hand, the goal is to actually _not_ use an expression then... maybe  ¯\_(ツ)_/¯

docs/proposals/00121_token-rate-limiting.md
jland-redhat: Maybe this is right... but I do think that we want to at least specify an optional "fail open" or "fail closed".

I feel like this was a major piece of missing functionality on limitador.

docs/proposals/00121_token-rate-limiting.md
jland-redhat: Just to get more clarity here would you recommend breaking bservability out into its own proposal separate from Token Rate Limiting?

docs/proposals/00121_token-rate-limiting.md
```diff
@@ -26,64 +28,76 @@ wildly in terms of token cost. This capability lets
...
 - **[M7]** Observability: metrics and tracing
   distinguishing admitted vs. limited requests,
-  including estimated cost, actual cost, and
-  remaining budget. The ability to log tokenomic
-  results distinctly for accounting.
-- **[M7]** Soft limits: usage tiers that modify request
+  including estimated cost, actual cost, and remaining
+  budget. The ability to log tokenomic results
+  distinctly for accounting.
+- **[M8]** Multi-environment observability: aggregate
```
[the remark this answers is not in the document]
jland-redhat: This could be another good reason to break observability out into its own separate proposal

docs/proposals/00121_token-rate-limiting.md
jland-redhat: Yeah  that makes sense but mentioned above if we are not going to have a default rule I think we need to at least have a way to define fail open/close

docs/proposals/00121_token-rate-limiting.md
```diff
@@ -167,58 +184,393 @@ the backstop, not the first line of defense.
...
+within a match block are ANDed; multiple match blocks
+are ORed.
+
+**Budget evaluation:** every `token_budget` on the
+matched rule is evaluated. Deny wins — if any budget
+hits a tier with `action: deny`, the request is
+rejected with 429. Otherwise the request continues and
+inject headers from all currently exceeded soft tiers
+are unioned onto the request. If two budgets inject the
+same header name, the last budget in config order wins;
```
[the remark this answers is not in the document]
jland-redhat: Disagree I think we should design with the full picture in mind. What actually gets implemented in for MVP can include less but we should try to have an idea of what the final shape of the CRD would look like so we don't paint ourselves into a corner.

docs/proposals/00121_token-rate-limiting.md
jland-redhat: Hmm that is a good point.

I originally had inject as an action and it just felt like it started to get very wordy but let me revert to that and get your feed back on how you feel about the original API

docs/proposals/00121_token-rate-limiting.md
jland-redhat: > overshoots are more annoying - as users may want different outcomes

I feel like for MVP the only outcome should be reconciling the count to be correct. But the traffic should be treated the same as if it undershoots

But agree that we should be applying the refund even in the MVP, that seems like it should be a fairly trivial problem.

docs/proposals/00121_token-rate-limiting.md
```diff
@@ -167,58 +184,393 @@ the backstop, not the first line of defense.
...
+Weighted cost at reconciliation:
+
+```
+cost = Σ (tokens_of_type * weight_of_type)
+```
+
+Admission still uses a single estimated cost; typed
+weights apply when the provider reports actual counts.
+
+#### Batch Workloads
```
jland-redhat, line 540: I am honestly still not totally clear on what we are trying to do with batch

docs/proposals/00121_token-rate-limiting.md
jordigilh: Agreed this needs to be explicit rather than left to a default. Worth pointing at prior art already in this repo: the generic per-filter `failure_mode` mechanism (`open`/`closed`, default `closed`) was added in [`praxis#48`](https://github.com/praxis-proxy/praxis/issues/48)/[`praxis#62`](https://github.com/praxis-proxy/praxis/pull/62), with the explicit rationale *"everything by default is closed; some benign filters could default to open."* [`praxis#547`](https://github.com/praxis-proxy/praxis/issues/547) proposes a bespoke `fail_open: true/false` (default `false`) specifically for rate limiting — should `token_rate_limit` just inherit the existing generic mechanism instead of #547 reinventing it? Worth resolving in #547 before this PR implements against it either way.

Aside: I can't verify Limitador/Kuadrant have a fail-open *default* to point to as precedent (their `failureMode` is explicitly configurable, `allow`|`deny`, no documented default) — so #547's own stated motivation (Envoy AI Gateway's availability-first philosophy) doesn't actually match its chosen default. That's consistent with Praxis's own house convention elsewhere (#269, #17, #330 all default fail-closed too), so the default itself is probably fine — the motivation text is what's inconsistent with it.

cc @shaneutt (author of #547) — worth deciding whether to delegate to #48's mechanism or keep #547's bespoke field.

docs/proposals/00121_token-rate-limiting.md
jordigilh: Two data points, both landing on the same answer as @jland-redhat's reply below: the [k8s AI Gateway WG proposal](https://github.com/kubernetes-sigs/wg-ai-gateway/pull/60) author (`yanavlasov`) states overshoot is accepted as an inherent property of any response-cost rate limiter ("enforcement happens on request path, debit on response path"). [Envoy AI Gateway](https://aigateway.envoyproxy.io/docs/0.2/api/) ships exactly this pattern (`apply_on_stream_done`) with no special overshoot handling.

This is already scoped as its own issue: [`praxis-proxy/ai#120`](https://github.com/praxis-proxy/ai/issues/120) ("Response-cost deduction"), independently citing the same mechanism.

@jland-redhat, endorsing your proposed resolution above — reconcile to the correct count, treat overshoot the same as undershoot, always apply the refund even in MVP. Suggest closing this open question with that and pointing implementation at #120 rather than re-specifying reconciliation mechanics here.

docs/proposals/00121_token-rate-limiting.md
```diff
@@ -167,58 +184,393 @@ the backstop, not the first line of defense.
...
+```
+
+Weighted cost at reconciliation:
+
+```
+cost = Σ (tokens_of_type * weight_of_type)
+```
+
+Admission still uses a single estimated cost; typed
+weights apply when the provider reports actual counts.
```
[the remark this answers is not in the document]
jordigilh: This asymmetry is a symptom of a design conflict worth surfacing explicitly — cc @shaneutt @jland-redhat:

[`praxis-proxy/ai#122`](https://github.com/praxis-proxy/ai/issues/122), an existing open sub-issue of this same #121 epic, specifies *"independent buckets... maintained per token type"* — i.e. separate budgets per type, not a weighted composite. [Envoy AI Gateway's shipping implementation](https://github.com/envoyproxy/ai-gateway/blob/main/examples/token_ratelimit/token_ratelimit.yaml) does the same. Both sidestep this exact bug class, since no cross-type weight conversion is needed at either admission or reconciliation.

@jland-redhat — you asked essentially this same question ("Would each just be a different rule?") on the [k8s AI Gateway WG rate-limiting proposal](https://github.com/kubernetes-sigs/wg-ai-gateway/pull/60); `yanavlasov`'s answer there was a custom cost expression per rule rather than a hardcoded per-type enum, which is closer to #658's weighted-composite approach than to #122's per-type-independent-budgets.

@shaneutt since you authored both #658 and #122 — worth explicitly reconciling: is #122 superseded by #658 's weighted-composite model, or should #658 adopt #122 's per-type-independent-budget design? Either is defensible, but right now the epic has two open, unreconciled answers to the same question.

docs/proposals/00121_token-rate-limiting.md
```diff
@@ -26,64 +28,76 @@ wildly in terms of token cost. This capability lets
...
+- **[M5]** Flexible bucket keys: quotas keyed by
+  request infromation. Headers, model identity, or
+  compound keys so different clients and models get
+  independent budgets. (TBD - this one may need to be
+  teased out more - possibly using CEL note there is an
+  upstream effor around  this found
+  [here](https://github.com/kubernetes-sigs/wg-ai-gateway/pull/57))
+- **[M6]** Hard deny with 429 when a budget is
+  exhausted, with standard rate limit response headers
+  (`Retry-After`, `X-RateLimit-*`).
```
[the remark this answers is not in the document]
jordigilh: Praxis's existing `RateLimitFilter` already computes this today (`praxis/filter/src/builtins/http/traffic_management/rate_limit/limiter.rs`, `rate_limit_headers`):

```rust
let retry_secs = if remaining < 1.0 {
    ((1.0 - remaining) / self.rate).ceil().max(1.0) as u64
} else {
    0
};
```

i.e. `retry_after = ceil(deficit / refill_rate)`, floored at 1 second, covered by `rejection_includes_retry_after_header` in `rate_limit_burst.rs`.

@shaneutt — recommend token rate limiting reuse the identical formula with `deficit` in tokens and `rate` in tokens/sec, for behavioral consistency between the two limiters and to avoid re-deriving a formula that already has test coverage.

docs/proposals/00121_token-rate-limiting.md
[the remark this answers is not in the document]
jordigilh: @shaneutt @jland-redhat — agreed, and here's concrete evidence for splitting it out: M8 alone touches **four** open, uncoordinated issues on distributed rate-limit state — [`praxis#155`](https://github.com/praxis-proxy/praxis/issues/155) (peer-replicated in-memory counters), [`praxis#551`](https://github.com/praxis-proxy/praxis/issues/551) (sliding window, "Valkey backend" compatible), [`ai#126`](https://github.com/praxis-proxy/ai/issues/126) (KV-store-registry + Valkey, a sub-issue of this same epic), and [`praxis#929`](https://github.com/praxis-proxy/praxis/issues/929) (opened Aug 7, peer-replication or Valkey for the existing request-count limiter) — with only #929→#155 cross-linked.

ai#126 looks like the most direct dependency since it's already scoped to this epic, but `KvStoreRegistry`'s `KvBackend` trait (`praxis/core/src/kv/mod.rs`) is in-memory-only today with no atomic-increment primitive, so it's not a drop-in yet either.

That's real, independent design surface (state backend + consolidation across 4 issues) — enough on its own to justify a separate proposal rather than an M8 subsection here.

docs/proposals/00121_token-rate-limiting.md
jordigilh: Small addition if this does get split out: worth including a structured "why was this limited" breakdown (which rule/window/tier triggered a given 429), similar to [Stripe's rate-limit headers](https://docs.stripe.com/rate-limits) which disambiguate `global-rate` / `endpoint-rate` / `global-concurrency` / `endpoint-concurrency`. Agreed it shouldn't live in TRLP itself per your point above — just flagging it as a requirement for whatever the separate accounting/observability design ends up covering, reusing TRLP's bucket-keying logic as you suggest.

docs/proposals/00121_token-rate-limiting.md
```diff
@@ -167,58 +184,393 @@ the backstop, not the first line of defense.
...
-requests that never complete. What conditions qualify
-(timeout, connection reset, upstream 5xx)? How long
-should a reservation be held before it is considered lost?
-This needs to be accounted for in architecture, but
-potentially out-of-scope for MVP.
+What conditions qualify a request as lost (timeout,
+connection reset, upstream 5xx)? How long should a
+reservation be held before it is considered lost?
 
 [#210]: https://github.com/praxis-proxy/ai/issues/210
```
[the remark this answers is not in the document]
jordigilh: +1, still unaddressed — `[#210]` is defined but never referenced. Either link it where relevant (likely the Token Type Capture section) or remove it.

docs/proposals/00121_token-rate-limiting.md
```diff
@@ -97,61 +111,64 @@ estimation, what to do with unused tokens, etc.
...
+  window.
 - Input message tokenization: fully tokenize request
   content before admission to use a real input token
-  count rather than relying on estimations, `max_tokens`
-  as a proxy, etc.
+  count rather than relying on estimations,
+  `max_tokens` as a proxy, etc.
+- Token Rate Limiting over a static window
 
 ## Why?
```
jordigilh, line 149: Suggest consolidating references into a dedicated Prior Art section rather than scattered inline links:

- [Kuadrant/Limitador](https://docs.kuadrant.io/dev/kuadrant-operator/doc/overviews/rate-limiting/) — configurable `failureMode` (allow/deny), Envoy RLS protocol
- [Envoy AI Gateway](https://aigateway.envoyproxy.io/docs/0.2/api/) — the most directly comparable prior art: `BackendTrafficPolicy` cost model, `apply_on_stream_done` reconciliation, per-token-type budget rules
- [NGINX Gateway Fabric `RateLimitPolicy`](https://github.com/nginx/nginx-gateway-fabric/blob/main/docs/proposals/rate-limiting.md)
- [Stripe's layered rate limiter](https://docs.stripe.com/rate-limits) — request-rate vs. concurrency separation, reason-coded 429s
- [IETF `draft-ietf-httpapi-ratelimit-headers`](https://datatracker.ietf.org/doc/html/draft-ietf-httpapi-ratelimit-headers-11) — modern successor to `X-RateLimit-*`
- [Kubernetes AI Gateway WG rate-limiting API](https://github.com/kubernetes-sigs/wg-ai-gateway/pull/60) — already linked inline elsewhere in this doc; same reviewers as this PR are active there

cc @shaneutt, FYI — purely organizational, not blocking.

docs/proposals/00121_token-rate-limiting.md
```diff
@@ -26,64 +28,76 @@ wildly in terms of token cost. This capability lets
...
+**Should have (same effort if capacity):**
+
+- **[S1]** Soft limits: usage tiers that modify request
   headers instead of rejecting, enabling downstream
-  systems (e.g. llm-d `InferenceObjective`) to
-  degrade gracefully as a client approaches its
-  budget.
-- **[M8]** Batch workload awareness: separate quota
+  systems (e.g. llm-d `InferenceObjective`) to degrade
+  gracefully as a client approaches its budget.
```
jordigilh, line 95: A couple of sibling issues under the #121 epic overlap with this section and aren't cross-linked yet:

- [`praxis#856`](https://github.com/praxis-proxy/praxis/issues/856) (soft rate limiting) already covers S1 as its own issue, and already cross-links #547/#549 itself.
- [`praxis#549`](https://github.com/praxis-proxy/praxis/issues/549) (quota exhaustion failover — reroute to an alternate backend on 429, also modeled on Envoy AI Gateway) isn't mentioned anywhere in this doc.

@jland-redhat (author of #856) @shaneutt (author of #549) — worth at least a Non-Goals mention if either is intentionally out of scope for #121.

docs/proposals/00121_token-rate-limiting.md
```diff
@@ -167,58 +184,443 @@ the backstop, not the first line of defense.
...
+Anthropic field names diverge, and even within one
+provider the shape varies by API and feature (Chat
+Completions vs Responses; cache, reasoning, audio,
+thinking). See OpenAI
+[prompt caching / usage details](https://developers.openai.com/api/docs/guides/prompt-caching)
+and Anthropic
+[prompt caching usage](https://docs.anthropic.com/en/docs/build-with-claude/prompt-caching)
+(plus Messages `usage` /
+`output_tokens_details.thinking_tokens`).
+
```
jordigilh, line 467: Slightly worried about `token_type_capture` turning into a second, independent place we parse provider responses for token data. We've actually hit this exact bug class before — #351 was a real double-count where two code paths in the Anthropic transform disagreed on whether `input_tokens` already included cached tokens. If this mapping reads straight off the response body instead of `token_count`'s existing `FilterContext` keys, we're setting up for the same failure mode again, just in a new spot. Could `token_count` just emit the extra typed fields (`cached_input`, `reasoning`, etc.) itself and have this consume those instead of re-parsing?

docs/proposals/00121_token-rate-limiting.md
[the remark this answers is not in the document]
jordigilh: Piling on to what you two already said above about splitting observability out — I'd go further and say S3 should go with it, maybe even more so. M8 at least has a pile of existing issues floating around it (messy, but something). Metering has none — we'd be defining Praxis's entire billing/chargeback data model for the first time, inside a rate-limiting proposal, for something the doc itself already calls "independent of rate-limit counters." Feels like its own doc, not a subsection here.

docs/proposals/00121_token-rate-limiting.md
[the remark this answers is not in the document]
jordigilh: On the CEL note for M5 — we've actually got a spike open for this already, #232 (which superseded #191). @shaneutt, you closed #191 in favor of it a while back so this probably isn't news, but #232 's proposed syntax (`header("x-admin") == "true"`) is basically the same shape as what's sketched here for bucket keys. Might be worth pulling from that instead of the k8s WG PR — would rather not end up with two separate CEL integrations in the codebase for the same "evaluate an expression against request attributes" problem.

jordigilh: Small one — there's already an issue for this exact bit: #123, tracks per-IP/per-header/per-model/composite bucket keys for token rate limiting. Might be worth a link since it's basically M5 spelled out as its own ticket.

docs/proposals/00121_token-rate-limiting.md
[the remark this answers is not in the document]
jordigilh: One more on header naming — #124 already specs this out as `X-RateLimit-Limit-Tokens` / `X-RateLimit-Remaining-Tokens` / `X-RateLimit-Reset` (distinct from the plain `X-RateLimit-*` names). Worth calling out explicitly here: the existing `rate_limit` filter uses the non-suffixed names (`X-RateLimit-Limit`, `X-RateLimit-Remaining`, `X-RateLimit-Reset`), and the doc's own Non-Goals say operators can run both `rate_limit` and `token_rate_limit` in the same pipeline. If `token_rate_limit` reuses the same header names instead of  #124 's `-Tokens`-suffixed set, whichever filter runs second clobbers the other's headers.

docs/proposals/00121_token-rate-limiting.md
jordigilh: Worth adding one more data point to the Kuadrant/Limitador entry above, since it resolves a question that could otherwise linger: why build a native filter here instead of delegating to Limitador the way that bullet's link describes?

That was spiked directly in praxis-proxy/ai#463 ("Spike: Native kuadrant-filter Integration in Praxis"). @didierofrivia's PoC got a native `kuadrant-filter` integration working end-to-end against real Authorino/Limitador, but at real cost (4x the code of wasm-shim, manual reimplementation of Envoy's async digest-loop). @shaneutt then closed the loop directly on that issue:

> Yes, this is the long-term plan: we want to implement policies using our native policy engine (CPEX is being donated here) and adapt the Kuadrant APIs and control-plane to it.

So this proposal's native-filter approach is the intended direction, not a fork from #127 ("Kuadrant integration for token rate limiting policy") — worth citing #463 's conclusion in the Prior Art section so a future reader doesn't wonder why this doesn't just delegate to Limitador. Side note: #463 still has an open, unanswered ask from @shaneutt (Aug 5) to spin the spike's conclusions into follow-up tasks — #127 likely needs rescoping once that happens ("adapt Kuadrant CRDs onto the native engine" rather than "delegate to Limitador"), but not a blocker for this PR.

docs/proposals/00121_token-rate-limiting.md
jordigilh: Following up on my own comment above — dug into this a bit more since M5 is marked TBD. Beyond #123, there are two more overlapping specs for the same "derive a bucket key from a header/expression" primitive:

- #129 ("Per-header rate limit bucket keys") — same idea, but scoped to the *existing* `rate_limit` filter, not the new one this PR builds.
- #301 ("Epic: Multi-Tenancy Support") — lists #129 as its own "Phase 1", and separately lists this epic (#121) under "Related Issues" for per-tenant token quotas.

And checked the current code: today's `rate_limit` filter has zero key-extraction abstraction to build on — `RateLimitMode` is a closed `Global`/`PerIp` enum with `PerIp` hardcoded straight to `DashMap<IpAddr, TokenBucket>` (`praxis/filter/src/builtins/http/traffic_management/rate_limit/config.rs:25-31`). So #129 and #123/M5 would each be building a bucket-key mechanism from scratch, independently, for two sibling filters, with overlapping requirements (#129 explicitly lists "fall back to global bucket if header absent" and "multiple key dimensions" — the same shape M5 wants). Worth deciding whether bucket-key derivation should be a shared helper/trait consumed by both `rate_limit` and `token_rate_limit`, rather than diverging behavior across three separate implementations (#123, #129, and M5's own CEL note).

docs/proposals/00121_token-rate-limiting.md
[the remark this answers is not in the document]
jordigilh: Actually, scratch what I said above about no sibling issue existing — there is one, I just missed it. #577 already builds almost this exact thing: balance check before the request goes out, usage report after it comes back, reading off the same `token_count` metadata. It reports to an external billing service instead of tracking things internally, but it's clearly the same problem. It's linked from #656 (the actual parent epic here), so it's not even that far removed.

If anything this makes the "split it out" case stronger, not weaker — now it's not just "undefined scope," it's "two different shapes of the same thing being built in parallel without talking to each other."

docs/proposals/00121_token-rate-limiting.md
```diff
@@ -26,64 +28,76 @@ wildly in terms of token cost. This capability lets
...
+  request information. Headers, model identity, or
+  compound keys so different clients and models get
+  independent budgets. (TBD - this one may need to be
+  teased out more - possibly using CEL note there is an
+  upstream effort around this found
+  [here](https://github.com/kubernetes-sigs/wg-ai-gateway/pull/57))
+- **[M6]** Hard deny with 429 when a budget is
+  exhausted, with standard rate limit response headers
+  (`Retry-After`, `X-RateLimit-*`).
 - **[M7]** Observability: metrics and tracing
```
jordigilh, line 76: Small thing, but the epic description for #121 calls out "hierarchical quotas — org > team > user" as one of the things this is supposed to deliver, and I don't see that anywhere in here anymore. M1's rules are flat — one condition, one `default` fallback, no nesting.

Might be intentional, might just be an MVP cut, not sure. Only reason I'm flagging it: #191 is already doing a version of this (org → team → user → agent) over on the Grid side with CRDT-backed budgets, and it explicitly mentions a "gateway-local" fallback for single-site setups — which sounds like it'd land right here if it ever happens. Not saying build it now, just maybe worth a one-liner in Non-Goals so it doesn't look like it got forgotten.

docs/proposals/00121_token-rate-limiting.md
usize: Inheriting our existing failure mode semantics seems like the nicest outcome here. If there's a problem because e.g., we want to send back a particular error code and response in a way that's not supported now it's probably worth just extending the common mechanism if possible?

docs/proposals/00121_token-rate-limiting.md
```diff
@@ -26,64 +28,76 @@ wildly in terms of token cost. This capability lets
...
-reported by the provider after the response completes.
-The estimation method must be configurable because
-different deployments have different cost models.
+The system requires reservation-based admission:
+requests are admitted by reserving an estimated cost up
+front, then the reservation is reconciled against
+actual usage reported by the provider after the
+response completes. The estimation method must be
+configurable because different deployments have
+different cost models.
```
usize, line 37: Re: reconciliation

Another thing to note is that OpenAI requires a flag to be set if we want token counts from streaming responses. We force that to be set in apis/src/openai/translation/chat_completions.rs but I think we should verify that it's always set anywhere this could happen. 

We should also call this out in the proposal I think, along with what happens if we don't get a token response from streaming since streaming is different enough that spelling it out might help e.g., a coding agent reading through the proposal here. :D

[The docs](https://developers.openai.com/api/reference/resources/chat/subresources/completions/streaming-events) also point out that if a stream is interrupted we won't get a token count at all from streaming responses so we should mention that in the proposal and suggest what we might do
there.. Maybe fire a warning and bump token counts based on an estimate as per the estimate reconciliation mentioned here.

([Anthropic's docs](https://docs.anthropic.com/en/docs/build-with-claude/streaming) mention the same problem, so whatever we do should probably be provider-agnostic.)

jordigilh: Dug into this a bit -- you're right that it's not always set, and I opened #770 to track it since it's bigger than just this proposal.

`map_stream_options` in `apis/src/openai/translation/chat_completions.rs` does force `include_usage: true`, but it's only wired into the Responses -> Chat Completions translation path. For a plain Chat Completions passthrough -- which is exactly what `token-counting.yaml` and `token-rate-limit.yaml` model -- nothing sets it. `token_count` itself never touches `stream_options`, it's pure response-side parsing. So any streaming deployment shaped like those two examples gets zero usage data unless the client sets the flag itself.

Anthropic's fine as-is, for what it's worth -- their streaming API sends usage in `message_delta`/`message_stop` unconditionally, no opt-in needed, so there's nothing to inject there.

Agreed the proposal should call this out explicitly rather than assume `token.total` is always available by the time reconciliation runs. On the interrupted-stream case: I'd lean toward "no usage metadata is a safe no-op, not an error" rather than firing a warning + estimate-based bump -- that's actually already how reconciliation is designed to behave in the token_rate_limit walking-skeleton I put together to validate this end-to-end (reservation just stands as final if `token.total` never shows up, no phantom charge either way). Happy to be argued out of that if there's a reason estimate-bumping is worth the extra complexity, but "silently correct, never wrong" seemed like the safer default for a first cut.

jordigilh:
One more gap surfaced by digging into the "windows are sliding" language added on line 231 in the latest revision: that's not just an implementation detail, it names a primitive Praxis doesn't have yet, and it isn't tracked anywhere in this proposal or its sibling issues.

- ** praxis praxis-proxy/ai#789 "Sliding window rate limiting"** is open and unassigned -- it's literally "sliding window algorithm... alongside the existing token bucket... compatible with distributed counters (Valkey backend)". Nothing in it or in this PR currently links to the other.
- That's a different concern from the counters-and-replication blockers already flagged (#155 , ai praxis-proxy/praxis#126) -- those are about how counts survive across replicas, praxis-proxy/ai#789 is about which windowing algorithm computes the count in the first place. A distributed *token-bucket* implementation wouldn't satisfy this proposal's current design; a single-instance *sliding-window* implementation wouldn't either. Both pieces are needed together for this MVP as currently written.

Worth cross-linking praxis-proxy/ai#789 here (and I'll drop a note on praxis-proxy/ai#789 pointing back), so whoever picks it up knows there's now a concrete consumer with real semantics to build against, and this proposal's dependency list reflects reality -- right now praxis-proxy/praxis#155 reads as "the" blocker but it's really one of at least two missing primitives.

(Flagging this as a fact-finding note, not proposing to solve it here -- happy to help scope praxis-proxy/ai#789 further if that'd be useful.)

docs/proposals/00121_token-rate-limiting.md
asaadbalum: Agreed that the estimation surface should accommodate expressions. The current config shape supports this naturally, MVP ships named strategies under the `strategy` key, and a future `expression` key using the existing CEL spike (#232) can be added to the same `estimation` object without a breaking change. Worth designing the implementation with that extension point in mind.

docs/proposals/00121_token-rate-limiting.md
```diff
@@ -167,58 +184,443 @@ the backstop, not the first line of defense.
...
+
+2. **Inference** - The request is forwarded upstream.
+   The provider performs inference and returns token
+   usage.
+
+3. **Reconciliation** - Actual provider-reported token
+   usage replaces the estimate in the budget. After
+   inference, apply the delta: refund unused reserved
+   capacity when actual cost is below the estimate, or
+   charge additional capacity when actual exceeds it.
```
asaadbalum: Building on @alexsnaps reservation overlap question: the proposal should also address concurrent in-flight reservations *within* a single budget.

When multiple requests from the same rule arrive in a burst, each reserves its conservative estimate. The cumulative reserved cost can significantly exceed what actual usage will be after reconciliation, causing false denials at `deny` tiers that real usage would not have reached.

This is likely an acceptable MVP trade-off (conservative admission protects the budget), but a sentence in the Request Lifecycle section acknowledging that in-flight reservations may cause transient over-restriction until reconciliation completes would prevent operator confusion during bursts.

docs/proposals/00121_token-rate-limiting.md
```diff
@@ -167,58 +184,443 @@ the backstop, not the first line of defense.
...
+inject headers from all currently exceeded soft tiers
+are unioned onto the request. If two budgets inject the
+same header name, the last budget in config order wins;
+prefer distinct header names per window.
+
+**Tiers:** each budget is a capacity ladder over a
+window. Thresholds use defined `capacity`. Every tier
+has an `action` with a `type`:
+
+- `inject` -- continue the request and apply
```
asaadbalum, line 281: The proposal defines tiers as a capacity ladder but doesn't specify validation constraints on the tier list. A few edge cases worth defining:

- Must tiers be declared in ascending capacity order, or does the system sort them internally? If declaration order matters, a misconfigured rule (deny at 100K listed before inject at 80K) would silently skip the soft tier.
- Are duplicate capacities (two tiers at 100K with different actions) an error or last-wins?
- Is a budget with zero tiers valid (track usage without any action)?

A sentence on ordering semantics and a validation-time rejection of invalid tier configurations would prevent a class of silent misconfiguration.

docs/proposals/00121_token-rate-limiting.md
```diff
@@ -167,58 +184,443 @@ the backstop, not the first line of defense.
...
+   budgets therefore reflect actual provider cost, not
+   the admission estimate alone.
+
+4. **Cleanup** - If a request is lost (timeout,
+   connection reset, upstream failure), release the
+   reservation after a configurable hold period.
+
+#### Estimation Strategies
+
+Estimation is pluggable and configured **once per
```
asaadbalum, line 404: The proposal specifies in-memory sliding windows with accumulated usage records, and Praxis supports hot-reload of filter chains. These interact: a config reload rebuilds the pipeline and creates new filter instances, resetting all in-memory state. Accumulated window records are lost, a team at 490K of a 500K hourly budget gets a fresh 500K after any config change.

This applies to the existing `rate_limit` filter too, but token bucket state recovers quickly (refill rate restores the bucket). Sliding window history spanning up to 24 hours of usage is a larger loss.

Worth acknowledging this as a known limitation of the in-memory model, and noting the relationship with proposal 00099 (Stateful Proxy State Management) which addresses state persistence across reloads.

PR #699, opened 2026-08-10
feat(inference): add llmisvc_model_provider_resolver

jland-redhat:
## Summary
- Add `llmisvc_model_provider_resolver`, porting only the LLMISvc / KServe BBR body-rewrite path from IPP’s `model-provider-resolver`.
- Prefer a configurable model header (default `X-Model`, aligned with `model_to_header`), fall back to body `"model"`, and when the value is a publisher ID (`publishers/.../models/<name>`) rewrite the body `"model"` to `<name>` only.
- Leave the routing header untouched so KServe can still route on the publisher ID; stash the original ID in `llmisvc_model_provider_resolver.publisher_id` for metering.
- Includes unit tests, example config, integration coverage, and generated filter docs.

Does **not** port ExternalModel / ExternalProvider resolution, weighted provider selection, Host rewrite, api-format detection, or credential handling.

## Sister PR (merge after this)
- [ ] `praxis-extproc`: fix ExtProc body-mutation path to set `content-length` to the mutated body length (avoids Envoy `mismatch_between_content_length_and_the_length_of_the_mutated_body` / 500).
- Link: https://github.com/opendatahub-io/praxis-extproc/pull/23

Without the ExtProc follow-up, body rewrites that change length will fail in Envoy `BUFFERED` + header `SEND` mode even though this filter’s rewrite is correct.

## Test plan
- [x] Unit tests for rewrite / header preference / body fallback / non-publisher passthrough
- [x] Example config + integration tests
- [x] Validated on local cluster with publisher-ID model request; upstream returned a completion:

```json
{
  "id": "chatcmpl-efb19481-6952-5be9-9572-ebfb8aa9070b",
  "model": "demo/sim-stream",
  "object": "chat.completion",
  "choices": [
    {
      "finish_reason": "stop",
      "message": {
        "role": "assistant",
        "content": "I am fine, how are you today? ..."
      }
    }
  ]
}
```


filters/src/inference/llmisvc_model_provider_resolver.rs
```diff
@@ -0,0 +1,566 @@
...
+fn resolve_model_name(
+    ctx: &HttpFilterContext<'_>,
+    header: &HeaderName,
+    obj: &serde_json::Map<String, serde_json::Value>,
+) -> Option<String> {
+    if let Some(from_header) = header_model_name(ctx, header) {
+        return Some(from_header);
+    }
+
+    obj.get("model")
```
aslakknutsen: Why this if model_to_header ?

filters/src/inference/llmisvc_model_provider_resolver.rs
```diff
@@ -0,0 +1,566 @@
...
+        };
+
+        // Stash the original publisher ID for later metering.
+        ctx.set_metadata(META_PUBLISHER_ID, model_name.as_str());
+
+        if obj.get("model").and_then(serde_json::Value::as_str) == Some(short_name) {
+            return Ok(FilterAction::Continue);
+        }
+
+        obj.insert("model".to_owned(), serde_json::Value::String(short_name.to_owned()));
```
aslakknutsen, line 190: Is there ever a chance that "model" won't be in the request body on a valid request?

jland-redhat: Not if it is following the OpenAI spec I don't believe, and it has to follow that spec if it is a vLLM model which should be the only thing that this filter would catch.

filters/src/inference/llmisvc_model_provider_resolver.rs
jland-redhat: `model_to_header` puts the model in the header.

But the model serving needs the header to be a "canonical id" publishers/<NAMESPACE>/models/<MODEL_NAME>. So the way BBR works in 3.5 is that we have the user put this value in the body "models" value and we move it into the header and replace it with the real <MODEL_NAME>

This resolver is that second part. 


External Models do something similar but replacement will happen based on the ExternalModel CRD.

aslakknutsen: I mean, this seems to "if it's not in the header as provided by model_to_header", it will fall back to getting the value from where model_to_header would/should have gotten the value.

filters/src/inference/llmisvc_model_provider_resolver.rs
aslakknutsen: Ok. I'm hinting at we could pretty easy do a StringBuffer splice to inject the new model value to avoid the full DOM deserialize and serialize of the complete body.  (handling missing model fields is a little bit more tricky, but a replace in a buffer is easy)

filters/src/inference/llmisvc_model_provider_resolver.rs
jland-redhat: Hmm maybe we can sync on this so I can better understand but I would think that is ok right?

It is faster to get this from the header and we fallback to the model if it is not there seems reasonable. But I can just remove the fallback if we think that makes more sense.

jland-redhat: Agreed after sync — dropped the body fallback. This filter now assumes `model_to_header` (or an equivalent) already set the routing header. If the header is absent/empty, we no-op and leave the body alone.

filters/src/inference/llmisvc_model_provider_resolver.rs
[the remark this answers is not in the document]
jland-redhat: Fixed in the latest push: we only rewrite when the body already has a `"model"` key; otherwise no-op (no invented field). Also added a unit test for header publisher ID + body without `"model"`.

filters/src/inference/llmisvc_model_provider_resolver.rs
```diff
@@ -0,0 +1,566 @@
...
+    #[test]
+    fn short_model_name_extracts_after_models() {
+        assert_eq!(
+            llmisvc_short_model_name("publishers/ns/models/granite-3.1-8b"),
+            Some("granite-3.1-8b")
+        );
+        assert_eq!(
+            llmisvc_short_model_name("publishers/ns/models/a/b"),
+            Some("a/b"),
+            "SplitN keeps remainder after first /models/"
```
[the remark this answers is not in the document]
jland-redhat: Fixed — assertion message now says `split_once`.

filters/src/inference/llmisvc_model_provider_resolver.rs
jland-redhat: What you are pointing at: today we serde_json::from_slice the whole request body into a DOM, mutate the "model" field, then re-serialize with replace_json_body. That is correct but pays a full parse/serialize of the JSON for a single string field change.

A StringBuffer-style splice would instead find the existing "model":"..." bytes in the buffered body and overwrite/replace just that string in place (or with a small rebuild of the surrounding bytes), avoiding a full DOM round-trip. That is attractive for hot paths, especially once we know "model" is always present on OpenAI/vLLM chat completions which is all this should be targeting currently.

For this PR I am keeping the DOM path: we already share replace_json_body with other filters, and with the header-required + no-invent-model guards the edge cases are simpler. Happy to follow up with a targeted byte/string splice later if profiling shows the parse/serialize cost matters here.

leseb:
@jland-redhat please rebase

docs/filters/reference.md
```diff
@@ -100,6 +100,7 @@ see the [Praxis core filter reference][core-ref].
 
 | Filter | Description |
 |--------|-------------|
+| [`llmisvc_model_provider_resolver`](llmisvc_model_provider_resolver.md) | Ports the `LLMISvc` / `KServe` BBR body-rewrite branch from IPP's `model-provider-resolver`. |
```
shaneutt: This language about BBR and IPP seems out of place for praxis-ai, we should generalize this.

filters/src/inference/llmisvc_model_provider_resolver.rs
```diff
@@ -0,0 +1,593 @@
...
+
+/// Separator between the publisher path and the short model name.
+const MODELS_SEPARATOR: &str = "/models/";
+
+// -----------------------------------------------------------------------------
+// Config
+// -----------------------------------------------------------------------------
+
+/// Deserialized YAML config for the `LLMISvc` model-provider resolver.
+#[derive(Debug, Deserialize)]
```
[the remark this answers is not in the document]
shaneutt: There are tests for this now.

filters/src/inference/llmisvc_model_provider_resolver.rs
```diff
@@ -0,0 +1,673 @@
...
+/// body `"model"` field to `<name>` only. The routing header is never
+/// modified -- `KServe` routes on the publisher ID.
+///
+/// If the header is absent/empty, or the body has no `"model"` field,
+/// this filter is a no-op (it does not invent a body `"model"`).
+///
+/// Does **not** resolve `ExternalModel` / `ExternalProvider` CRDs, perform
+/// weighted provider selection, rewrite `Host`, or inject credentials.
+///
+/// # YAML configuration
```
shaneutt: We need to re-construct this more generically, not referencing the IPP project (which is more of a downstream concern) and maybe spell BBR out for the audience that would read this at the top.

shaneutt (approved):
I have comments to resolve, but most of them are about language as opposed to structure. Thanks @jland-redhat 👍

PR #726, opened 2026-08-12
feat(compact): reactive compaction with direct input support

crstrn13:
## Summary

Completes the remaining work items from #30:

- **Use `previous_usage` for token estimation** — `should_compact()` checks the stored `usage.total_tokens` from the rehydrated response before falling back to local tiktoken counting
- **Configurable summary prefix** — `summary_prefix` config option (default: `[Previous conversation summary]`) carried through to compaction items and both translation paths
- **Hide compaction items from `input_items` API** — `normalize_input_items()` filters out `{"type": "compaction"}` items so clients never see internal state
- **Persist compaction as a stored response** — after summarization, the compacted response is written back to the response store with the compaction item in its input
- **Explicit compact endpoint** — `POST /v1/responses/{id}/compact` triggers compaction on a previously stored response

## Behavior details

- **Invalid config is rejected up front** — an invalid `compact_threshold` (e.g. below `MIN_COMPACT_THRESHOLD`) returns `400 invalid_request_error` *before* the compactability gate, for both streaming and non-streaming requests, and for direct-input requests
- **Summaries cover the full conversation** — `should_compact()` summarizes the entire rehydrated history including the current turn, so the summarizer has complete context. The current turn is not lost or duplicated: `replace_messages()` preserves it verbatim after the compaction item via a tail split
- **Direct input passes through** — requests with inline input (no rehydration) are not reactively compacted

## Test plan

- [x] `cargo test -p praxis-ai-apis` — unit tests for all work items
- [x] `cargo test -p praxis-ai-filters` — no filter regressions
- [x] `cargo test -p praxis-tests-integration` — example integration tests (mock backends), incl. `compact.rs`
- [x] vLLM functional SDK tests (`test_openai_responses_vllm.py`) — invalid-threshold rejection and over-threshold rehydrated-history compaction, against live vLLM + OGX
- [x] `make lint`, `rustdoc`, unicode-safety, DCO/signoff all green


apis/src/openai/responses/compact/mod.rs
```diff
@@ -280,62 +400,86 @@ impl HttpFilter for CompactFilter {
 // Compaction Logic
 // -----------------------------------------------------------------------------
 
+/// Returns `true` when compaction should proceed.
+fn ensure_compactable_state(ctx: &HttpFilterContext<'_>) -> bool {
```
aslakknutsen, line 418: This is untested.

tests/integration/tests/suite/examples/compact.rs
aslakknutsen, line 217: "AGENTS.md requires functional integration coverage for new capabilities in example configs."

No tests POST to /v1/responses/compact.

apis/src/openai/responses/compact/mod.rs
aslakknutsen: "CompactFilter's struct doc (mod.rs:97-99) and docs/filters/openai_responses_compact.md Configuration Notes (lines 12-13) still state compaction only applies when rehydrate loaded stored history and that single-turn requests skip compaction. The module-level doc (mod.rs:14-20) and direct-input integration test contradict this. Update both stale locations to describe the direct-input path."

aslakknutsen (changes requested):
Outside the PR diff, but https://github.com/praxis-proxy/ai/blob/bc1e5bd9386d07ed7e3f79f2ec115ff4708a8364/apis/src/openai/responses/store/filter.rs#L464

"ResponseStoreFilter::should_init_store_for_request() returns false for POST /v1/responses/compact because the path is not is_responses_create and the classifier does not set has_previous_response_id for sub-resource routes. The store Arc is therefore never registered into ResponseStoreRegistry before CompactFilter::resolve_store_and_tenant() runs. On a freshly built pipeline (or after reload), the first explicit compact call fails with "response store not available" even when the backend is healthy. Extend should_init_store_for_request (or equivalent) to initialize the store for POST /v1/responses/compact, mirroring how rehydrate paths are handled."

apis/src/openai/responses/compact/mod.rs
crstrn13: should be added.

tests/integration/tests/suite/examples/compact.rs
crstrn13: added.

apis/src/openai/responses/compact/mod.rs
crstrn13: regenerated docs.

apis/src/openai/responses/compact/mod.rs
```diff
@@ -485,36 +761,108 @@ fn parse_summarization_response(body: &[u8]) -> Result<String, String> {
...
+/// (no rehydration). Preserving `input` would duplicate the entire
+/// conversation after the summary. Result: `[compaction_item]`.
 fn replace_messages(state: &mut ResponsesState, compaction_item: Value) {
-    let mut new_messages = Vec::with_capacity(state.input.len() + 1);
-    new_messages.push(compaction_item);
-    new_messages.extend(state.input.iter().cloned());
-    state.persisted_messages = new_messages.clone();
+    let direct_input = !state.history_rehydrated;
+    let new_messages = if direct_input {
+        vec![compaction_item]
```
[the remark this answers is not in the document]
crstrn13: I decided to take path a instead due to simplicity. If a user wants compact without rehydration, it can use the explicit `POST /v1/responses/compact`

leseb:
@crstrn13 please rebase

leseb (changes requested):
P1 — Explicit compact rejects conforming requests. [parse_compact_request_body](https://github.com/praxis-proxy/ai/blob/b4f65903614cc683f6015331e1150045c373d174/apis/src/openai/responses/compact/mod.rs#L575-L597) requires nonstandard response_id. The checked-in API contract requires model and supports input/previous_response_id. Standard {model,input} requests therefore return 400.

P1 — Response violates CompactResource. [The response builder](https://github.com/praxis-proxy/ai/blob/b4f65903614cc683f6015331e1150045c373d174/apis/src/openai/responses/compact/mod.rs#L647-L684) returns object: "response" without usage; the contract requires object: "response.compaction" and usage.

P2 — Fail-open becomes fail-closed. [Explicit compaction](https://github.com/praxis-proxy/ai/blob/b4f65903614cc683f6015331e1150045c373d174/apis/src/openai/responses/compact/mod.rs#L240-L269) converts the Ok(None) produced by on_failure: open into a hard-coded 502.

P2 — Reactive compaction creates unreachable rows. [apply_compaction](https://github.com/praxis-proxy/ai/blob/b4f65903614cc683f6015331e1150045c373d174/apis/src/openai/responses/compact/mod.rs#L285-L312) generates and persists a response ID, then discards it without exposing or linking it. Every compaction leaves an orphan record despite the normal response subsequently persisting the compacted history.

leseb (changes requested):
last one [on_failure: open](https://github.com/praxis-proxy/ai/blob/e02ffc476b131c5a78bc0e218bd4750bbe5107b3/apis/src/openai/responses/compact/mod.rs#L694-L700) returns raw input messages inside CompactResource.output. Those lack required output fields such as type, id, and status, so typed OpenAI clients may reject the response.

leseb (changes requested):
final: 

Persisted response.compaction objects have no status, but rehydration requires status == "completed". Referencing the returned ID therefore fails as status 'unknown', despite the new code claiming follow-up continuity.

Accept response.compaction records during rehydration and add an actual follow-up test.

leseb (approved):
Thanks for your patience!

aslakknutsen (approved):

Issue #249, opened 2026-07-02
Epic: Inference Provider Fallback


shaneutt:
When an inference provider fails or otherwise can not provide an adequate response we need semantics to describe conditions to identify the fault, and logic to correct it by failing over to another provider.

Triumph1701:
The failover semantics should distinguish provider selection from retry policy. A useful contract is to classify the failure (transport, timeout, 429 with Retry-After, auth, or deterministic 4xx), record the attempted provider/model, and only substitute a model when the request is still safe to replay. Then expose whether the final result came from the primary or a fallback so callers can make idempotency and quality decisions without treating every provider error as interchangeable.

Issue #109, opened 2026-05-15
Failover with model substitution


shaneutt:
On failure or low confidence, retry against a more capable model. This extends the existing failover mechanism with model-tier escalation.

## Requirements

- Ordered model tier configuration (e.g. small -> medium -> large)
- Retry with model field substitution in request body
- Configurable escalation triggers: HTTP errors, timeout, rate limiting
- Per-tier retry budgets to prevent cascade
- Integration with circuit breaker state

## Acceptance Criteria

- Integration test: request to small model fails, retried on medium model successfully
- Model field in request body correctly rewritten on escalation
- Retry budget exhaustion stops escalation
- Circuit breaker open on a tier skips to next tier

## Observability (from praxis-proxy/ai#134)

- Failover events exposed to access logging and metrics
- Log which provider/model was attempted, why it failed, and which fallback was selected
- Metric counter for failover events by provider and trigger type

PR #730, opened 2026-08-13
Document and Test Provider Fallback Capabilities via Iterative Request Router

usize:
## Summary

Provider fallbacks, along with model rewriting and api translation, are now feasible via the iterative request router. Because this functionality exists as an emergent property of our existing filters and filter pipeline, we should document it. These changes document it via a new integration test and example config.

## Related issue

Ref: #249, #109

## Validation

- [ x] Unit tests
- [ x] Integration or functional tests
- [ x] `make lint`

## Checklist

- [ x] I reviewed every changed line and can explain the change.
- [x ] New capabilities include an example config and functional example test.
- [ x] User-facing behavior and generated documentation are updated.
- [ x] Performance-sensitive changes include appropriate benchmark or load-test evidence.
- [ x] Commits are signed and include a `Signed-off-by` trailer.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3205,7 +3205,8 @@ dependencies = [
 [[package]]
 name = "praxis-proxy"
 version = "0.5.2"
-source = "git+https://github.com/praxis-proxy/praxis.git?tag=v0.5.2#af54edcd24de01e4dac7da59a809bf88bd246fe5"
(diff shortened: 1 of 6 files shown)


leseb:
@usize lint fails

leseb (approved):
It's a good example thanks! However, the impl has some holes like streaming support we should reject for now since IRR does not support it yet.

PR #731, opened 2026-08-13
feat(routing): add provider load balancing

nerdalert:
## Summary

  This PR adds local provider selection policies to the AI gateway's intelligent_route filter.

The authenticated provider-routing boundary from praxis-proxy/ai#386 is now part of the base branch.

```
  Client
    |
    v
  Consumer Gateway
    |
    v
  intelligent_route
    |
    +--> Provider Gateway A
    +--> Provider Gateway B
    `--> Provider Gateway C
```

  Grid remains the asynchronous control plane. It determines provider eligibility, admission state, ordering, and selection-group membership, then publishes that information in the routing overlay.

  The AI gateway consumes the accepted overlay from memory. When the overlay is loaded, it validates the group metadata and builds an immutable index for request-time selection.

  The supported selection modes are:

  - `deterministic`: select the first viable provider, preserving existing behavior.
  - `roundRobin`: rotate evenly among viable providers in the active group.
  - `random`: select uniformly among viable providers in the active group.

  Session affinity is evaluated before new-request selection. A valid existing binding continues to use its provider when admission permits it. New, unbound requests use the configured selection mode.

  The request path does not call Grid, Kubernetes, the operator, overlay-sync, ConfigMaps, EPP, Prometheus, or a remote scoring service.

  ## Responsibility Boundary

  This change keeps control-plane and request-time responsibilities separate:

```
  Grid
    -> discovers providers
    -> evaluates health, freshness, and admission
    -> applies routing and scoring policy
    -> assigns selection groups
    -> publishes an immutable overlay

  Praxis AI intelligent_route
    -> reads the accepted in-memory snapshot
    -> resolves session affinity
    -> identifies the best viable group
    -> applies the configured local selection mode
    -> selects a provider cluster

  Praxis core load_balancer
    -> selects an endpoint within that provider cluster
```

  intelligent_route does not recompute scores, geography, health, or group membership.

  ## Overlay Contract

  The routing overlay may include an optional selection policy and producer-assigned groups:

```
  selection_policy:
    mode: roundRobin

  candidates:
    - cluster: provider-a
      selection_group: 0
    - cluster: provider-b
      selection_group: 0
    - cluster: provider-c
      selection_group: 0
```

  Group 0 is preferred. If a group contains no candidate eligible for a new request, selection proceeds to the next group.

  Provider-level retries after a forwarding failure are not introduced by this PR. Praxis core may retry endpoints within the selected cluster, but retrying a different Grid provider cluster remains follow-up work.

  ## Compatibility

  Existing overlays without `selection_policy` or `selection_group` retain the deterministic, ordered-first behavior.

  An upgraded deployment therefore does not begin balancing traffic unless its overlay explicitly enables a non-deterministic selection mode.

  Unknown selection modes and malformed group metadata fail validation. Invalid reloads retain the last-known-good snapshot, and in-flight requests continue using the immutable snapshot with which they started.

  The selection mode is not inferred from:

  - scoring policy;
  - candidate score or rank;
  - provider count;
  - the presence of selection groups;
  - queue-depth metrics;
  - KV-cache metrics.

  Scores and ranks do not implicitly become traffic weights. Weighted selection is a separate future extension.

  ## Snapshot And Hot-Path Behavior

  Selection-group indexes are constructed when a routing snapshot is loaded, not while processing requests.

  Round-robin counters are:

  - local to each gateway process;
  - scoped to a capability and selection group;
  - stored with the immutable routing snapshot;
  - concurrency-safe;
  - preserved when an overlay rewrite has the same semantic revision;
  - replaced when a genuinely changed snapshot is accepted.

  The request path performs bounded in-memory lookups and one local selection operation. It does not parse configuration, rebuild groups, scan unrelated capabilities, scrape metrics, or make control-plane calls.

  This provides balanced behavior per gateway. It does not attempt to coordinate one global round-robin counter across a gateway fleet.

## Security And Tenancy

  The selector only chooses among candidates already present in the accepted routing overlay and eligible under their published admission state. It does not discover providers, expand authorization, resolve credentials, or override Grid's trust and policy decisions.

  Selection groups must not cross security boundaries unintentionally. Grid remains responsible for ensuring that candidates grouped together are valid for the same capability and applicable policy scope. Praxis AI validates the group structure but does not reinterpret tenancy, locality, credential, or provider trust rules.

  Provider credentials and authorization data are not used as selection inputs or exposed through selection metadata. The authenticated provider-routing boundary remains owned by praxis-proxy/ai#386 (https://github.com/praxis-proxy/ai/pull/386).

  ## Operational Observability

  Successful route decisions expose bounded in-process metadata for:

  - selection mode;
  - selection group;
  - selected provider site, cluster, and stable ID;
  - admission state;
  - rank and selection tier when present;
  - serving overlay revision;
  - session-affinity creation, reuse, or failover state.

  This evidence supports logs and tracing without adding a request-time
  control-plane dependency.

  Provider IDs, session identifiers, request IDs, and other high-cardinality values should not become unbounded metric labels. Aggregate metrics should use controlled dimensions. Rich per-request routing details are better represented as bounded trace attributes when OpenTelemetry support is enabled separately.

  ## Multiple Consumer Gateways

  This design also works when several consumer or edge gateways accept traffic. Each gateway loads the same Grid overlay and independently applies the configured selection mode:

```
                      +--> Consumer Gateway East --+
  Client / GTM / DNS -+                            +--> Providers A, B, C
                      +--> Consumer Gateway West --+
```

  No consumer gateway coordinates with another gateway during request handling. Each process maintains its own round-robin counters and session-affinity state. This preserves the local, low-latency request path and avoids introducing a shared database or remote selection service.

  With enough new requests, each gateway should independently distribute traffic  across the active provider group. The combined distribution will generally  approach the expected split, but a strict global sequence or exact fleet-wide ratio is not guaranteed.

  Limitations include:

  - Round-robin counters are process-local, not synchronized across gateways.
  - A gateway restart or genuinely changed overlay snapshot resets that gateway's selection state.
  - Uneven DNS, GTM, or load-balancer traffic across consumer gateways can produce uneven aggregate provider traffic.
  - Small request samples may not produce an exact fleet-wide split.
  - Session affinity intentionally overrides selection distribution for established  sessions.
  - Long-lived or unevenly sized requests can create unequal backend load even when request counts are equal.
  - The selector balances provider selections, not tokens, latency, concurrency, or inference cost.
  - Provider health and admission changes reach gateways asynchronously through new Grid overlays; AI does not probe providers on the request path.
  - This PR does not add cross-gateway counter coordination, provider-level retries, metric-derived weights, or globally exact traffic allocation.
  - External DNS, Anycast, or GTM remains responsible for distributing incoming clients across consumer gateways. Grid does not provide that entry-point balancing.

  ## Functional Proof

  The focused
  provider-traffic demo (https://github.com/praxis-proxy/demos/tree/main/demos/grid-provider-traffic) exercises the intended topology:

```
  One consumer gateway
    |
    +--> Provider Gateway A --> backend A
    +--> Provider Gateway B --> backend B
    `--> Provider Gateway C --> backend C
```

  The clean runtime validation demonstrated:

  - one consumer gateway receiving the complete measured workload;
  - three provider gateways advertising the same capability;
  - all three candidates accepted in one active selection group;
  - 60 serial requests without session-affinity headers;
  - HTTP 200 for all 60 requests;
  - exact 20/20/20 provider distribution;
  - a repeating three-provider round-robin sequence;
  - an unchanged serving overlay during the measurement window;
  - no gateway restart during selection;
  - complete environment teardown.

  The demo proves equal provider selection from a stable snapshot. It does not claim to prove weighted routing, provider-level retry, lower-group runtime failover, or fleet-global round-robin coordination.

  Other Grid demos cover separate concerns:

  - llm-d pool metrics (https://github.com/praxis-proxy/demos/tree/main/demos/grid-llmd-pool-metrics) demonstrates asynchronous EPP metric scoring and rank changes.

  - Grid GLB (https://github.com/praxis-proxy/demos/tree/main/demos/grid-glb-demo) demonstrates entry-point, edge, and provider routing across a larger topology.

  OpenTelemetry instrumentation and the tracing UI are separate work and are not modified by this PR.

  ## Dependency

  The authenticated provider-routing work from praxis-proxy/ai#386 (https://github.com/praxis-proxy/ai/pull/386) is merged and included in this branch's base. This PR remains one isolated feature commit above current main.

  ## Follow-Up Work

  This PR establishes equal local selection within producer-defined groups.
 
 **Separate follow-up work includes:**

  - provider-level retry within the selected group;
  - explicit lower-group fallback and recovery runtime proof;
  - metric-derived traffic_weight publication by Grid;
  - weighted request or new-session selection in Praxis AI;
  - normalization across providers with different capacities;
  - affinity behavior when weights or membership change;
  - protection against noisy metrics and excessive overlay churn;
  - throughput and latency benchmarks;
  - multi-consumer runtime validation;
  - bounded aggregate selection metrics;
  - globally coordinated selection, if a concrete requirement emerges.

  Weighted routing must remain explicit. Scores, ranks, queue depth, and KV-cache pressure should not silently become traffic percentages.

  ## Related Issue

  Related to
  praxis-proxy/grid#31.

  The research spike includes the focused provider-traffic demo narrative, runtime evidence, and a video showing one consumer gateway distributing requests across three provider gateways.

  This PR implements the AI request-time selection portion of the spike. It does not close the Grid-side overlay producer work or the remaining retry, fallback, and weighted-selection research.

  ## Validation

  - [x] Unit tests
  - [x] make lint

  Validated with:

  cargo +nightly fmt --all -- --check
  cargo test -p praxis-ai-filters
  cargo test -p praxis-ai-filters --no-default-features
  cargo test -p praxis-ai-filters --all-features
  cargo clippy --workspace --all-targets -- -D warnings
  make lint
  make test
  RUSTDOCFLAGS="-D warnings" cargo doc --workspace --no-deps
  git diff --check

  Focused coverage includes:

  - deterministic selection;
  - random selection with deterministic ordinal coverage;
  - round-robin distribution;
  - concurrent round-robin selection;
  - counter wraparound;
  - counter isolation across capabilities;
  - multiple and interleaved capabilities;
  - contiguous group validation;
  - mixed grouped and ungrouped rejection;
  - admission-aware group fallback;
  - overlays without selection metadata;
  - semantically unchanged reloads;
  - changed-snapshot selection state;

  ## Checklist

  - [x] New capabilities include an example configuration and functional demo.
  - [ ] Performance-sensitive changes include benchmark or load-test evidence.
  - [x] Commits are signed and include a Signed-off-by trailer.

  The runtime demo validates functional distribution. It is not presented as a throughput or latency benchmark.

  ## Breaking Changes

  None intended.

  Existing overlays retain deterministic behavior. New deployments can opt into roundRobin or random through an explicit selection policy. This PR does not change scoring semantics, add weighted routing, or modify OpenTelemetry instrumentation.

/cc @jordigilh


filters/src/routing/overlay.rs
```diff
@@ -488,10 +542,16 @@ impl RouteSnapshot {
 
         descriptor::validate_local_site(&doc.local_site)?;
         let candidates = overlay_to_candidates(&doc)?;
+        let group_index = group_index::build(&candidates)?;
         let generated_at = doc.generated_at.map(|s| Arc::from(s.as_str()));
+        let selection_mode = doc
+            .selection_policy
+            .map_or(PickerPolicy::Deterministic, |policy| policy.mode);
```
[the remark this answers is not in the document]
nerdalert: Addressed in commit 638fac0. RouteSnapshot now emits a warning for non-deterministic selection policies when no candidates carry selection_group. The existing deterministic fallback remains unchanged, but the misconfiguration is now visible during snapshot construction.

filters/src/routing/overlay.rs
```diff
@@ -990,12 +1057,26 @@ pub(crate) fn read_overlay_bounded(path: &Path) -> Result<Vec<u8>, std::io::Erro
...
+    content: &[u8],
+    snapshot: &ArcSwap<RouteSnapshot>,
+    expected_scope: Option<&ExpectedOverlayScope>,
+) -> bool {
     let new_hash: [u8; 32] = Sha256::digest(content).into();
-    let unchanged = new_hash == snapshot.load().content_hash;
+    let current = snapshot.load();
+    let unchanged = new_hash == current.content_hash
+        || (current.semantic_revision.is_some()
+            && RouteSnapshot::from_overlay_with_scope(content, expected_scope)
```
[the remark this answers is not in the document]
nerdalert: Addressed in commit 638fac0. Reload handling now parses and validates a changed overlay once, compares its semantic revision to the current snapshot, and reuses that validated RouteSnapshot when the revision changed. This preserves the last-known-good behavior and avoids the previous double parse.

filters/src/routing/overlay.rs
```diff
@@ -219,6 +220,42 @@ pub(crate) struct OverlayDocument {
...
+            Self::Deterministic => "deterministic",
+            Self::RoundRobin => "round_robin",
+            Self::Random => "random",
+        }
+    }
+}
+
+/// Wire wrapper kept extensible independently from routing and scoring policy.
+#[derive(Clone, Copy, Debug, Deserialize)]
+#[serde(rename_all = "camelCase", deny_unknown_fields)]
```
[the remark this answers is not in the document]
nerdalert: Thanks for catching this. The current branch keeps `SelectionPolicy` forward-compatible by accepting additive fields, and `selection_policy_ignores_unknown_forward_compatible_fields` now protects that contract. Unknown `mode` values and malformed known fields are still rejected.

jordigilh (commented):
A few unit-test coverage gaps I noticed while going through the selection-group/policy logic, plus a couple of open questions. Nice work keeping this backward-compatible — overlays without `selection_policy`/`selection_group` provably retain the old deterministic-first-viable behavior (`select_legacy` in `picker.rs` is exactly the pre-PR path), so existing deployments won't start load-balancing silently on upgrade.

Open question, not a blocker: for a session that's already affinity-bound, if its pinned candidate later drops out of admission, is that covered by an existing (pre-this-PR) affinity-fallback test, or does the group/policy addition touch that path at all?

filters/src/routing/overlay.rs
```diff
@@ -517,6 +581,16 @@ impl RouteSnapshot {
     }
 }
 
+/// Warn when a non-deterministic policy has no grouped candidates to select.
+fn warn_if_selection_policy_has_no_groups(selection_mode: PickerPolicy, group_index: &GroupIndex) {
+    if selection_mode != PickerPolicy::Deterministic && group_index.is_empty() {
+        tracing::warn!(
```
jordigilh, line 588: Following up on the earlier thread here (addressed in 638fac0) — the warning now exists, but I don't see a test that actually exercises `warn_if_selection_policy_has_no_groups` (e.g. constructing a snapshot with `selectionPolicy.mode: roundRobin` and zero grouped candidates, then asserting the warning fires). Given this is the main operator-facing signal for that misconfiguration, seems worth locking down directly.

filters/src/routing/overlay.rs
```diff
@@ -614,6 +688,9 @@ fn compute_semantic_digest(overlay_value: &serde_json::Value) -> Result<String,
     if let Some(network) = overlay_value.get("network") {
         semantic.insert("network".to_owned(), network.clone());
     }
+    if let Some(selection_policy) = overlay_value.get("selection_policy") {
+        semantic.insert("selection_policy".to_owned(), selection_policy.clone());
```
jordigilh, line 694: This adds `selection_policy` as digest input, but I don't see a test that proves changing *only* `selection_policy` (leaving candidates/local_site/network untouched) actually changes the computed digest. Since `handle_overlay_reload` treats a matching digest as "nothing to do," a field that's silently missing from the digest would mean a policy change (e.g. `deterministic` → `random`) could slip through as a no-op reload. Worth a test like `digest_changes_when_only_selection_policy_changes` that holds everything else constant and asserts the digest differs.

filters/src/routing/overlay.rs
```diff
@@ -942,11 +1021,23 @@ fn handle_overlay_reload(
...
+    if content_hash == current.content_hash {
+        tracing::debug!("intelligent_route: overlay snapshot is unchanged");
         return;
     }
 
-    apply_overlay(path, &content, snapshot, expected_scope);
+    match RouteSnapshot::from_overlay_with_scope(&content, expected_scope) {
+        Ok(new_snap) => {
+            if is_semantically_unchanged(&current, &new_snap) {
+                tracing::debug!("intelligent_route: overlay snapshot is semantically unchanged");
```
jordigilh, line 1036: The `is_semantically_unchanged` short-circuit branch in `handle_overlay_reload` isn't hit by any test with a `selection_policy`-bearing overlay. Given the digest-coverage question above, I'd want to see this exercised directly — e.g. reload with two overlays that differ only in whitespace/timestamp but have identical semantic content, and assert the short-circuit path is taken (snapshot not swapped).

filters/src/routing/picker.rs
```diff
@@ -0,0 +1,311 @@
...
+    name: &str,
+    policy: PickerPolicy,
+) -> Option<(&'a RouteCandidate, Option<u32>)> {
+    if let Some(capability_groups) = groups.get(&kind).and_then(|by_name| by_name.get(name)) {
+        for group in capability_groups {
+            if let Some(candidate) = select_from_group(candidates, group, policy) {
+                return Some((candidate, Some(group.number)));
+            }
+        }
+        return None;
```
jordigilh, line 30: When `capability_groups` exists for the requested capability but every group in it is inadmissible, this returns `None` — no test covers that path. It's meaningfully different from `unavailable_best_group_falls_through` (which has one draining group and one healthy fallback group) since here *all* groups are exhausted. Is `None` the intended contract for full exhaustion, or should this fall back to something? Either way, seems worth a dedicated test.

filters/src/routing/picker.rs
```diff
@@ -0,0 +1,311 @@
...
+        .get(ordinal)
+        .and_then(|&index| candidates.get(index))
+}
+
+/// Resolve a policy to an index inside a non-empty selection group.
+fn choose_ordinal(policy: PickerPolicy, len: usize, counter: &AtomicUsize) -> usize {
+    match policy {
+        PickerPolicy::Deterministic => 0,
+        PickerPolicy::RoundRobin => counter.fetch_add(1, Ordering::Relaxed) % len,
+        PickerPolicy::Random => rand::rng().random_range(0..len),
```
jordigilh, line 77: The `Random` arm of `choose_ordinal` isn't exercised by any test. `random_stays_inside_the_best_group` (despite the name) doesn't actually call `choose_ordinal`/`select_candidate` with `PickerPolicy::Random` — it calls `select_from_group_at` directly with fixed ordinals 0 and 1, which only proves group membership, not that the `Random` policy dispatch produces a bounded, varying selection. Suggest a test that drives `select_candidate` with `PickerPolicy::Random` a few hundred times and asserts every result stays inside `0..len` (and ideally visits more than one member).

filters/src/routing/intelligent_route.rs
```diff
@@ -902,39 +926,15 @@ fn evict_expired(affinity: &SessionAffinity) {
...
-            continue;
-        }
-        if !is_admitted_for_new_request(c.admission_state) {
-            continue;
-        }
-        return Some(c);
+/// Record bounded request metadata without changing forwarding headers.
+fn record_selection_metadata(ctx: &mut HttpFilterContext<'_>, group: Option<u32>, mode: PickerPolicy) {
+    if let Some(group) = group {
+        ctx.set_metadata(ROUTE_SELECTION_GROUP, group.to_string());
```
jordigilh, line 935: Neither the new-request selection path nor the reused-session path has a test that reaches the `Some(group)` arm here — i.e. nothing asserts `ROUTE_SELECTION_GROUP`/`ROUTE_SELECTION_MODE` actually get written to request metadata for a grouped candidate. This is the observability hook for the PR's headline feature (which group/mode served a given request), so it seems worth a direct test rather than relying on it being hit incidentally.

filters/src/routing/overlay.rs
nerdalert: Addressed in 37a4d5bd. The helper now returns whether the warning branch was taken, and the focused test covers both non-deterministic/no-group and deterministic/no-group inputs while retaining the actual structured warning.

filters/src/routing/overlay.rs
nerdalert: Addressed in 37a4d5bd.  holds candidates, local site, and network constant and verifies that changing only  changes the digest.

filters/src/routing/overlay.rs
nerdalert: Addressed in 37a4d5bd. The test now drives  with a grouped round-robin envelope whose only changes are timestamps, then uses  to prove the serving snapshot and its picker state were not replaced.

filters/src/routing/picker.rs
nerdalert: Addressed in 37a4d5bd. Full exhaustion intentionally returns ;  now covers an  active group plus an  fallback group.

filters/src/routing/picker.rs
nerdalert: Addressed in 37a4d5bd. The random test now calls the production  path repeatedly and verifies every result remains in group 0 and never reaches the fallback group. It avoids a probabilistic assertion that could make CI flaky.

filters/src/routing/intelligent_route.rs
nerdalert: Addressed in 37a4d5bd. The new test verifies  and  on the initial selection and again when the same session reuses its affinity binding.

nerdalert:
Thanks for the affinity question as well. The group/policy change does not alter binding reuse: an existing binding may continue to use a `new_and_existing` or `existing_only` candidate, while an excluded or missing candidate invalidates that binding and runs normal new-request selection. The existing `existing_session_can_use_existing_only`, `failover_skips_existing_only_candidate`, and `missing_candidate_after_reload_fails_over` tests cover those boundaries; the new grouped-metadata test also exercises the affinity-reused path under `roundRobin`.

leseb (approved):

PR #736, opened 2026-08-13
fix(server): maintain Prometheus recorder state

### LINKED ISSUE #734: task: Implement Prometheus recorder upkeep in Praxis core
## Objective

Implement Prometheus recorder lifecycle management in `praxis-proxy/praxis` core so every application using the shared admin and metrics integration receives scrape-independent recorder upkeep.

## Required behavior

When the Praxis admin endpoint is enabled:

- install the Prometheus recorder before startup instrumentation begins;
- install exactly one process-global recorder;
- use the same recorder for `/metrics` rendering and upkeep;
- retain the recorder through an opaque core-owned handle;
- register exactly one Pingora-managed background service;
- wait five seconds before the first upkeep pass;
- run `PrometheusHandle::run_upkeep()` through `tokio::task::spawn_blocking`;
- prevent upkeep passes from overlapping or accumulating;
- observe Pingora shutdown while a pass is running;
- stop scheduling new passes during shutdown;
- avoid routine success logging and privacy-sensitive logging.

When the admin endpoint is disabled, the recorder and upkeep service must not be installed.

Applications must not install a second recorder, retain the raw exporter handle, or schedule their own upkeep task.

## Scope

This change addresses Prometheus histogram recorder upkeep when `/metrics` is not scraped. It drains buffered histogram observations and publishes pending metric descriptions on a predictable schedule.

The change must not add request-path work or alter:

- metric names or labels;
- histogram buckets or quantiles;
- the `/metrics` route or response format;
- health and readiness endpoints;
- existing admin endpoint behavior;
- proxy request processing.

Idle-series expiration is a separate concern. `run_upkeep()` does not itself provide complete idle-series expiration, so do not add unrelated idle-timeout configuration unless the pinned exporter supports it through a documented public API and the scope is explicitly expanded.

## Implementation guidance

Use the existing Praxis recorder installation and Pingora service-registration paths. Refactor them so recorder acquisition occurs once and the endpoint and upkeep service share the same handle.

Use a Pingora `BackgroundService`, not an unmanaged recurring `tokio::spawn`.

Run each synchronous upkeep pass through `spawn_blocking`. Await the pass before scheduling another one, and continue observing shutdown while the pass is running. A blocking closure that has already started cannot be forcibly cancelled; the scheduler should stop scheduling additional work and return.

## Validation

Add deterministic focused tests covering:

- delayed first execution;
- no immediate upkeep at startup;
- missed intervals do not queue catch-up passes;
- shutdown before the first pass;
- shutdown while a blocking pass is running;
- no overlapping upkeep passes;
- exactly one recorder/upkeep service registration;
- shared recorder use by `/metrics` and upkeep;
- unchanged behavior when the admin endpoint is disabled.

Use paused Tokio time and synchronization primitives rather than real five-second sleeps or bounded yield loops.

Run the normal Praxis formatting, Clippy, test, and documentation gates.

## AI integration

Praxis AI should inherit this behavior by consuming the Praxis release containing the core implementation. No AI-local recorder installation, exporter dependency, raw handle lookup, or background upkeep service is required.

nerdalert:
## Summary

  Ensure the Prometheus recorder performs periodic upkeep even when `/metrics` is not being scraped. Praxis AI installs the Prometheus recorder when the admin endpoint is configured. `PrometheusBuilder::install_recorder()` exposes the recorder and its handle, but it does not start the exporter's upkeep task. Without an external scrape, buffered  histogram samples may remain pending and accumulate over time. 

This change registers a Pingora-managed background service that calls `PrometheusHandle::run_upkeep()` every five seconds. The service starts only when the admin endpoint is configured, reuses the existing recorder handle, does not install another recorder or exporter, waits five seconds before its first upkeep pass, operates independently of `/metrics` scrape traffic, exits through Pingora's graceful-shutdown signal, and does not log metric values, requests, or credentials. The existing /metrics endpoint and recorder installation behavior remain unchanged.

  ## Why a Pingora background service?

  An unmanaged `tokio::spawn` task would not participate in the server lifecycle and could continue running during shutdown. Using Pingora's `BackgroundService` integrates upkeep
  with the existing runtime and provides an explicit ShutdownWatch. The service stops when shutdown begins and does not require a separate runtime or worker thread.

  ## Scope

  This change addresses Prometheus histogram recorder upkeep when /metrics is not scraped. It drains the recorder's buffered histogram observations on a predictable schedule. It does not expire inactive metric series. Stale dynamic-label series still require Praxis to expose and configure the Prometheus recorder's recency or `idle_timeout` behavior. That should be handled as a separate Praxis capability.

  ## Implementation

  The server registers `PrometheusUpkeepService` alongside the existing admin endpoint when `config.admin.address` is present. The service retains a clone of the public `PrometheusHandle` and performs upkeep after a five-second interval, while listening for Pingora's shutdown signal. The interval provides regular buffer maintenance without  coupling upkeep frequency to Prometheus's configured scrape interval.

## Alternatives Considered

  - Call upkeep during `/metrics` scrapes: simpler, but fails when metrics are not being scraped, which is the condition this fixes.
  - Use `tokio::spawn:` works mechanically, but creates an unmanaged task without Pingora’s graceful-shutdown lifecycle.

## Perf/Risk

It adds a small amount of background work, but no per-request hot-path work.

  Every five seconds, one task calls PrometheusHandle::run_upkeep(). The cost depends on the number of histogram buffers and metric series, but it is generally modest and is work the recorder already expects to perform. It does not inspect requests, block proxy workers, create another exporter, or run once per request.

  The practical risk is only unusually high metric cardinality, where any recorder maintenance becomes more expensive. This change should reduce the greater risk of buffered histogram state accumulating indefinitely when /metrics is not scraped. Benchmarking or observing upkeep duration would still be appropriate before claiming zero overhead.

  ## Validation

  - [x] cargo +nightly fmt --all -- --check
  - [x] git diff --check
  - [x] make lint
  - [x] cargo test -p praxis-ai-proxy
  - [x] make test

  make test completed with 2,524 passing tests, 27 ignored tests, and one pre-existing SQLite database is locked failure. The affected test passed when run independently.

  ## Breaking changes

  None. The behavior is limited to deployments with the admin endpoint enabled, and the existing Prometheus endpoint and metric format are unchanged.

Closes #734

### DIFF
--- a/server/src/server.rs
+++ b/server/src/server.rs
@@ -6,8 +6,14 @@
 use std::{
     path::PathBuf,
     sync::{Arc, Mutex},
+    time::Duration,
 };
 
+use async_trait::async_trait;
+use pingora_core::{
+    server::ShutdownWatch,
+    services::background::{BackgroundService, background_service},
+};
 use praxis_core::{
     PingoraServerRuntime,
     config::{Config, ProtocolKind},
@@ -19,6 +25,33 @@ use tokio_util::sync::CancellationToken;
 use tracing::info;
 
 use crate::pipelines::resolve_pipelines;
+
+/// Prometheus recorder maintenance interval.
+///
+/// `PrometheusBuilder::install_recorder` does not spawn the exporter upkeep
+/// task. Keep upkeep independent of `/metrics` scrape traffic so histogram
+/// sample buffers are drained even when the admin endpoint is not scraped.
+/// This does not expire stale metric series; that requires recorder
+/// recency/idle-timeout configuration in Praxis.
+const PROMETHEUS_UPKEEP_INTERVAL: Duration = Duration::from_secs(5);
+
+/// Pingora-managed Prometheus recorder maintenance.
+struct PrometheusUpkeepService {
+    /// Handle for the recorder installed by the Praxis admin service.
+    handle: metrics_exporter_prometheus::PrometheusHandle,
+}
+
+#[async_trait]
+impl BackgroundService for PrometheusUpkeepService {
+    async fn start(&self, mut shutdown: ShutdownWatch) {
+        loop {
+            tokio::select! {
+                _ = shutdown.changed() => break,
+                () = tokio::time::sleep(PROMETHEUS_UPKEEP_INTERVAL) => self.handle.run_upkeep(),
+            }
+        }
+    }
+}
 
 // -----------------------------------------------------------------------------
 // Config Path Resolution
@@ -232,6 +265,12 @@ fn register_admin_endpoints(
     kv_stores: &praxis_core::kv::KvStoreRegistry,
 ) {
     if let Some(admin_addr) = &config.admin.address {
+        let upkeep = PrometheusUpkeepService {
+            handle: praxis_protocol::http::pingora::metrics::install_prometheus_recorder().clone(),
+        };
+        server
+            .server_mut()
+            .add_service(background_service("Prometheus upkeep", upkeep));
         praxis_protocol::http::pingora::health::add_admin_endpoints_to_pingora_server(
             server.server_mut(),
             admin_addr,
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3183,8 +3183,10 @@ dependencies = [
 name = "praxis-ai-proxy"
 version = "0.2.0"
 dependencies = [
+ "async-trait",
  "cargo_metadata",
  "clap",
+ "metrics-exporter-prometheus",
  "nix",
  "notify",
  "praxis-ai-apis",
@@ -3193,6 +3195,7 @@ dependencies = [
  "praxis-proxy-core",
  "praxis-proxy-filter",
  "praxis-proxy-protocol",
+ "quixotic-plecostomus-core",
  "serde",
  "tempfile",
  "tikv-jemallocator",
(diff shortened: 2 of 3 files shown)


aslakknutsen:
@nerdalert We could use this in praxis as well if you feel like it :)

server/src/server.rs
```diff
@@ -232,6 +265,12 @@ fn register_admin_endpoints(
     kv_stores: &praxis_core::kv::KvStoreRegistry,
 ) {
     if let Some(admin_addr) = &config.admin.address {
+        let upkeep = PrometheusUpkeepService {
```
aslakknutsen, line 268: "register_admin_endpoints calls install_prometheus_recorder to obtain the upkeep handle, then add_admin_endpoints_to_pingora_server calls the same function again. OnceLock makes this safe but redundant; registering the upkeep service after add_admin_endpoints would still clone the same handle with a single install site."

server/src/server.rs
```diff
@@ -20,6 +26,33 @@ use tracing::info;
...
+    handle: metrics_exporter_prometheus::PrometheusHandle,
+}
+
+#[async_trait]
+impl BackgroundService for PrometheusUpkeepService {
+    async fn start(&self, mut shutdown: ShutdownWatch) {
+        loop {
+            tokio::select! {
+                _ = shutdown.changed() => break,
+                () = tokio::time::sleep(PROMETHEUS_UPKEEP_INTERVAL) => self.handle.run_upkeep(),
```
aslakknutsen, line 50: "run_upkeep is invoked synchronously inside the async select loop. If upkeep becomes expensive under high metric cardinality, shutdown.changed() is not polled until the current pass finishes. Consider bounding work, documenting expected cardinality limits, or running upkeep in spawn_blocking with cancellation awareness if shutdown latency matters."

nerdalert:
> @nerdalert We could use this in praxis as well if you feel like it :)

@aslakknutsen  Good idea! I’m going to move this into Praxis core, since core already owns the Prometheus recorder and /metrics endpoint. That lets us use one recorder handle, run upkeep without blocking the async worker, and give the fix to every Praxis-based application.

 I’ll close this PR as superseded . AI will pick up the fix through release.

nerdalert:
Superseded by https://github.com/praxis-proxy/praxis/pull/977

PR #739, opened 2026-08-14
ci(store): add PostgreSQL CI coverage for response store

leseb:
## Summary

Add CI workflows that exercise the response store with a real PostgreSQL
backend, covering unit tests, Rust integration tests, and end-to-end vLLM
SDK integration tests.

- New `postgres.yaml` workflow runs the 27 ignored store unit tests (service
  container with `DATABASE_URL`) and the 2 ignored Rust integration tests
  (which spawn their own container via `start_postgres()`).
- The vLLM integration workflow gains a `vllm-responses-postgres` job that
  reruns the full Python SDK test suite with `STORE_BACKEND=postgres`.
- Fixes two pre-existing bugs: mixed-case `ThreadId` in table name suffixes
  broke PostgreSQL schema validation, and missing `ssl_mode: disable` caused
  TLS handshake failures against the plaintext test container.
- Adds `test-postgres-unit` and `test-postgres-integration` Makefile targets.

## Related issue

N/A — no existing issue for PostgreSQL CI coverage.

## Validation

- [x] Unit tests
- [x] `make lint`

## Checklist

- [ ] New capabilities include an example config and functional example test.

## Breaking changes

None.


franciscojavierarceo (dismissed):
the new PostgreSQL vLLM lane is useful coverage, but its workflow path filters do not include the SDK test file that the lane executes. A future change to tests/integration/sdk/openai/test_openai_responses_vllm.py or its STORE_BACKEND setup will not trigger this PostgreSQL lane, so the coverage can silently go stale. we should add tests/integration/sdk/openai/** to the push, pull_request, and merge_group path filters, including the merge-group git diff allowlist.

leseb:
> the new PostgreSQL vLLM lane is useful coverage, but its workflow path filters do not include the SDK test file that the lane executes. A future change to tests/integration/sdk/openai/test_openai_responses_vllm.py or its STORE_BACKEND setup will not trigger this PostgreSQL lane, so the coverage can silently go stale. we should add tests/integration/sdk/openai/** to the push, pull_request, and merge_group path filters, including the merge-group git diff allowlist.

@franciscojavierarceo the vllm-integration.yaml workflow already includes tests/integration/sdk/openai/** in all three trigger sections. The vllm-responses-postgres job inherits those workflow-level triggers, it doesn’t have separate path filters.

tests/integration/sdk/openai/test_openai_responses_vllm.py
```diff
@@ -85,17 +87,31 @@ def _ogx_endpoint() -> str:
     return f"{host}:{port}"
 
 
+def _patch_store_backend(config: str, db_path: str) -> str:
+    if STORE_BACKEND == "postgres" and STORE_DATABASE_URL:
```
aslakknutsen: Do we need STORE_BACKEND and STORE_DATABASE_URL ? It kinda silently goes for sqlite if URL is missing. Strictly speaking we could use the URL to detect the DATABASE type couldn't we?

leseb: yes, we could detect the backend based on the url, but i feel it's clearer to have the STORE_BACKEND, wdyt?

.github/workflows/vllm-integration.yaml
```diff
@@ -178,3 +184,117 @@ jobs:
           if [ -f /tmp/ogx.pid ]; then
             kill "$(cat /tmp/ogx.pid)" 2>/dev/null || true
           fi
+
+  # ----------------------------------------------------------------------------
+  # Responses API integration tests with PostgreSQL store backend
+  # ----------------------------------------------------------------------------
+
+  vllm-responses-postgres:
```
aslakknutsen, line 192: Ok enough for this PR, but this is basically a copy of what's above repeating the same setup/clean up. Could we at least make a tracker issue to move this to a matrix job or a composition action or something?

tests/integration/sdk/openai/test_openai_responses_vllm.py
aslakknutsen: not sure.. 

STORE_BACKEND=postgres
STORE_DATABASE_URL=sqlite://xxx

I would assume the URL I'm configuring this with has more value in my intent then what this other variable that I need to keep in sync has...

.github/workflows/vllm-integration.yaml
leseb: https://github.com/praxis-proxy/ai/issues/749

tests/integration/sdk/openai/test_openai_responses_vllm.py
leseb: ok i'll make that change!

leseb: 7baa79ef

aslakknutsen (approved):

PR #750, opened 2026-08-14
docs(proposals): add semantic classification & routing proposal (MVP 3)

jordigilh:
## What?

Splits MVP 3 (semantic routing) scoping out of `ai#74` into its own proposal doc, per this repo's `docs/proposals/` convention. `What?`/`Why?` only, per the first-PR rule — `How?` follows in a follow-up PR once the two blocking decisions on `ai#715` are confirmed.

## Why?

`ai#715` promoted semantic routing to its own scoped issue with two open design questions (embedding backend, compliance-tier gating) blocking any real estimate. A revised recommendation grounded in external precedent (vLLM Semantic Router) is posted on the issue, awaiting @cnuland's confirmation. This proposal captures the scoping so it's tracked in the same place as this repo's other design work, not just in issue comments.

Closes: none yet (tracks `ai#715`)


docs/proposals/00715_semantic-classification-routing.md
```diff
@@ -0,0 +1,131 @@
+---
+issue: https://github.com/praxis-proxy/ai/issues/715
+discussion: https://github.com/praxis-proxy/ai/issues/715
```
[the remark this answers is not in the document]
jordigilh: Fixed in 4c929cc (superseding an earlier unsigned push, 84b089d, to satisfy this repo's signed-commits-check): discussion now points to https://github.com/praxis-proxy/ai/pull/750, matching the 00354/00484 convention.

cnuland:
@jordigilh @shaneutt 
Praxis should own the callout/integration, routing policy, stickiness, fallback, and final enforcement. llm-d-sc should own model/tokenizer loading, batching, inference, calibration, abstention, caching/optimization (let it be stateful), and the classifier runtime itself.

Because of that, I don’t think “embedding backend selection” should be a blocking Praxis design decision anymore. Embedding similarity may be one classifier technique, but we should not make it the semantic-routing architecture. The current direction for llm-d-sc is a pluggable runtime, initially likely Rust Candle transformer from huggingface, with the ability to experiment with other backends and classifier approaches later. Praxis should consume the signal contract without caring whether that signal came from a fine-tuned encoder, embeddings, another optimized runtime, or eventually something larger.

I’d also avoid making current-turn-only classification part of the contract. Turn-by-turn evaluation is important for domain and sensitivity, but that is different from saying the classifier can only inspect the current prompt. Some signals will need conversational context. The API should be able to receive sufficient context for correctness while allowing llm-d-sc to cache tokenization/features/session-derived information for performance. If that state is lost and sufficient context is unavailable, the classifier should abstain and Praxis should apply its fallback/stickiness behavior rather than guessing.

On the compliance point, I agree strongly with the concern, but I’d frame the initial architecture as local classification inside the trusted inference boundary rather than routing every prompt to an external embeding provider. This needs to be done with things like redaction and egress guardrails if done properly (not something that should be exclusively on praxis). If we eventually support a remote classification backend, Praxis/policy should be able to prevent payloads from being sent to a backend that is not permitted for that data.

docs/proposals/00715_semantic-classification-routing.md
```diff
@@ -0,0 +1,269 @@
...
+stakeholders:
+  - cnuland
+  - shaneutt
+---
+
+# Semantic Classification & Routing
+
+## What?
+
+Add a new `semantic_classify` filter that classifies an inference request's
```
shaneutt: I think having a separate filter for this makes sense. I think `semantic_classifier` perhaps?

docs/proposals/00715_semantic-classification-routing.md
```diff
@@ -0,0 +1,269 @@
...
+**task type** (e.g. code, math, creative, general) and **complexity**
+(easy/hard) from the prompt content itself, and promotes each as an
+independent fact (not fused into a single combined capability) into the
+same capability-kind seam `intelligent_route` already uses for
```
shaneutt, line 28: 👍

docs/proposals/00715_semantic-classification-routing.md
```diff
@@ -0,0 +1,269 @@
...
+backend's concern, not this filter's: v1's default backend uses embedding
+similarity against operator-defined seed-utterance routes, but the
+contract this filter consumes — task/complexity facts in — is
+technique-agnostic, so a future backend using a different method needs no
+interface change.
+
+This closes MVP 3 ("semantic routing"), the one Praxis MVP with zero
+existing code today, of the 3 the customer named (alongside
+model-based/app-based multi-cluster routing and intelligent/Grid-aware
+routing).
```
shaneutt: This kind of thing isn't really needed in a proposal, you can reference issues and discussions above if needed. Also references to customers/users should not be in these upstream proposals.

```suggestion
```

docs/proposals/00715_semantic-classification-routing.md
```diff
@@ -0,0 +1,269 @@
...
+    confidence threshold. Protects against misrouting on ambiguous prompts.
+  - `skipped_by_policy` — the compliance-tier gate blocked the request
+    before the classifier ever ran.
+  - `unavailable` — the classifier itself could not run (backend down,
+    timed out, errored).
+
+  `intelligent_route` needs to tell these apart: "we confidently decided
+  not to classify this prompt" is a materially different signal than "the
+  classification system is broken," and only the latter should be treated
+  as a health/availability concern.
```
shaneutt: Using the `intelligent_router` means that a classification will result in signals used by a filter, or a scorer. That sounds right but I'm wondering if there are other cases we need to be considering. Would classification be used in guardrails? Nothing that needs to hold this up, but possibly something worth notes on.

docs/proposals/00715_semantic-classification-routing.md
```diff
@@ -0,0 +1,269 @@
...
+- On `unavailable`, semantic routing **fails closed**: it routes on
+  existing structured signals (model header, `mcp.method`) only, the same
+  way Grid already fails closed elsewhere — e.g. `grid#7`'s GridSite
+  eligibility gate excludes a peer when no matching site is found, rather
+  than guessing. It never fails open to an alternate classification path,
+  since that would reopen the exact compliance side-channel this proposal
+  exists to close.
```
shaneutt:
```suggestion
- On `unavailable`, semantic routing **fails closed** by default: it routes on
  existing structured signals (model header, `mcp.method`) only, the same
  way Grid already fails closed elsewhere — e.g. `grid#7`'s GridSite
  eligibility gate excludes a peer when no matching site is found, rather
  than guessing. It never fails open to an alternate classification path,
  since that would reopen the exact compliance side-channel this proposal
  exists to close.
```

There's a use case for semantic classification as purely route _optimization_, not the ultimate routing _decision_. It seems we would need to make the failure-mode optional?

docs/proposals/00715_semantic-classification-routing.md
```diff
@@ -0,0 +1,269 @@
...
+  (`skipped_by_policy`, fail-closed not fail-open). **The tier itself must
+  come from trusted policy context — declared tenant/session/app metadata
+  evaluated upstream of this filter — never from a client-supplied header
+  or derived from the request content itself.** A content-derived tier is
+  circular: it would require classifying the request to decide whether
+  classifying it is safe. Exactly which component sets the tier and how
+  it's carried internally is a `How?` decision; this proposal only fixes
+  the trust boundary.
+- Wire the classification results into `intelligent_route` as new
+  capability kinds, reusing the existing candidate-selection seam.
```
shaneutt, line 111: These are a bit verbose, can we distill them down into more concise goals?

docs/proposals/00715_semantic-classification-routing.md
```diff
@@ -0,0 +1,269 @@
...
+
+### Non-Goals
+
+- Semantic caching (a related but separate feature; not in scope here).
+- Jailbreak / prompt-injection / general content-safety detection — that's
+  `ai_guardrails`'s domain. This filter only classifies task type and
+  complexity for routing purposes, not safety.
```
shaneutt: Do we at least want to leave the door open for the `semantic_classifier` filter to send signals to guardrails later?

docs/proposals/00715_semantic-classification-routing.md
```diff
@@ -0,0 +1,269 @@
...
+  a candidate v2 follow-up pending resolution of the accuracy question in
+  Open Questions below, not settled v1 scope. v1's embedding/cosine-
+  similarity path has no instruction-following LLM call, so it isn't
+  exposed to prompt injection targeting an evaluator's own judgment; if the
+  LLM-escalation path above is ever built, it would need the same
+  untrusted-content isolation framing tracked in
+  [ai#754](https://github.com/praxis-proxy/ai/issues/754).
+- Response-side classification (classifying model *output*, not the
+  request). Request-side only, matching how `intelligent_route` already
+  operates.
```
shaneutt: Can we make these a bit more concise?

docs/proposals/00715_semantic-classification-routing.md
```diff
@@ -0,0 +1,269 @@
...
+
+## Why?
+
+### Motivation
+
+Semantic routing is one of the 3 MVPs the customer named, but today it has
```
shaneutt: Nothing related to customers/users with any specificity should be in these upstream proposals.


```suggestion
Semantic routing today has
```

docs/proposals/00715_semantic-classification-routing.md
```diff
@@ -0,0 +1,269 @@
...
+zero coverage anywhere in `praxis`, `praxis-ai`, or `praxis-grid` — confirmed
+by a code-search sweep across all three repos returning zero
+embedding/classification code; only structured-field matching (model
+header, `mcp.method`) exists. It was tracked only as a stretch-goal bullet
+inside the `ai#74` epic until being promoted to its own scoped issue,
+[ai#715](https://github.com/praxis-proxy/ai/issues/715), specifically
+because bundling it inside a larger epic left it with no independent
+estimate and two unresolved decisions blocking any estimate from firming
+up:
```
shaneutt: I would expect the motivation to be more about what value proposition the feature has, not a statement of user desire, or a link to issues that ask for it.

docs/proposals/00715_semantic-classification-routing.md
```diff
@@ -0,0 +1,269 @@
...
+   remote), covered in Open Questions below.
+2. **An unresolved compliance/security question**, not just a timeline
+   risk. Prior research (`@cnuland`, Jul 31) found that routing on shared
+   context across models with different compliance postures — e.g. a local
+   model cleared for PII vs. an external model that isn't — risks leaking
+   sensitive data across an egress boundary *through the classification
+   call itself*, independent of which model ultimately serves the request.
+   The same research found that turn-level embeddings (not
+   whole-conversation) are also needed to avoid semantic dilution as
+   context grows.
```
shaneutt: Can we get this a bit more concise?

docs/proposals/00715_semantic-classification-routing.md
```diff
@@ -0,0 +1,269 @@
...
+classification specifically to avoid the same side-channel risk, and its
+multi-turn handling is turn-scoped by default. A revised recommendation
+grounded in this precedent — defaulting to local classification for the
+routing decision rather than an external API — was posted on
+[ai#715](https://github.com/praxis-proxy/ai/issues/715#issuecomment-5293888816)
+and confirmed by `@cnuland` on `ai#750` ("I'd frame the initial
+architecture as local classification inside the trusted inference
+boundary rather than routing every prompt to an external embedding
+provider"). The `How?` section of this proposal will follow once the open
+questions below are resolved, per this repo's proposal convention.
```
shaneutt: Why is this in here?

docs/proposals/00715_semantic-classification-routing.md
```diff
@@ -0,0 +1,269 @@
...
+   caching per the Goals above), published benchmarks for the underlying
+   model put single-sentence CPU inference around 10–30ms — small next to
+   typical LLM-based guardrail-provider latencies (several hundred ms to
+   ~1s per published NeMo Guardrails benchmarks), so no separate
+   parallel-execution optimization against `ai_guardrails` is proposed for
+   v1. This should be revisited only if a future classification backend
+   (the technique-agnostic contract permits e.g. an LLM-based classifier)
+   brings classify latency into the same order of magnitude as guardrails.
+
+### User Stories
```
shaneutt, line 246: Good user stories 👍

docs/proposals/00715_semantic-classification-routing.md
```diff
@@ -0,0 +1,269 @@
...
+  of technique — by implementing one interface, not by modifying the
+  filter's core logic.
+- As an operator, I want requests that don't confidently match any
+  configured route to pass through unclassified rather than be misrouted on
+  a low-confidence guess.
+- As an SRE/operator, I want to distinguish "the classifier confidently
+  chose not to classify this request" from "the classification system is
+  broken" so that only the latter pages anyone or affects availability
+  SLOs, and so that a broken classifier degrades to structured routing
+  rather than silently guessing or falling back to an external service.
```
shaneutt:
```suggestion
- As an SRE/operator, I want semantic routing failure modes that
  enable me to decide how to proceed when a classifier fails.
```

shaneutt (changes requested):
The motivation and goals here don't read like they're coming from the perspective of the value prop we want for the upstream project, they read more like they come from a product management perspective. Let's make this more about the value users of Praxis will get.

docs/proposals/00715_semantic-classification-routing.md
jordigilh: Good catch — you're right that this needs to be optional in effect, but we landed on not adding a new knob to `semantic_classifier` itself, since that would duplicate (and could contradict) `intelligent_route`'s own fallback-candidate mechanism.

Instead: on `unavailable`, `semantic_classifier` sets no task/complexity facts but does promote the outcome itself (e.g. `X-Classification-Outcome: unavailable`) via the same header mechanism it already uses for task/complexity. That gets you the optionality without a new field — configure a fallback candidate on `intelligent_route` for "classification as pure optimization," or don't for "classification as the routing decision," the same pattern `router` already uses for catch-all routes, now matchable explicitly via `conditions.headers` instead of only inferable from whether a candidate happens to exist.

This also matches how two production systems handle the same shape of problem: Envoy's `ext_proc` (the mechanism vLLM Semantic Router itself runs on) exposes it as an explicit [`failure_mode_allow` toggle](https://github.com/envoyproxy/envoy/blob/main/api/envoy/extensions/filters/http/ext_proc/v3/ext_proc.proto), and [vLLM Semantic Router](https://github.com/vllm-project/semantic-router/blob/29aba60e/website/docs/proposals/production-stack-integration.md) calls its own unavailable-bypass a deliberate "fail-safe design," not an accident of routing-table shape.

Updated in the Goals section of the latest revision (53cf067).

docs/proposals/00715_semantic-classification-routing.md
jordigilh: Applied, worded to point at `intelligent_route`'s fallback-candidate config specifically — stays consistent with the answer above on making the failure mode optional.

docs/proposals/00715_semantic-classification-routing.md
jordigilh: Done — renamed to `semantic_classifier` throughout.

docs/proposals/00715_semantic-classification-routing.md
jordigilh: Removed — the `What?` section no longer references the customer/MVP framing.

docs/proposals/00715_semantic-classification-routing.md
jordigilh: Yes — added a note under the jailbreak/prompt-injection Non-Goal: `ai_guardrails` can already scope its own policy off this filter's `X-Task-Type`/`X-Complexity` headers today via `conditions.headers`, no interface change needed on either side.

docs/proposals/00715_semantic-classification-routing.md
jordigilh: Fair, and I tried, but every other round of feedback (yours on guardrails/optionality, the `unavailable` fail-mode discussion) added detail rather than removed it. Net Goals length didn't shrink. Open to a concrete second pass if you still want it trimmed, but I'd rather not silently drop the reasoning that answered other review threads.

docs/proposals/00715_semantic-classification-routing.md
jordigilh: Same answer as the sibling thread above — `ai_guardrails` reading this filter's headers is already possible today, called out explicitly in Non-Goals now.

docs/proposals/00715_semantic-classification-routing.md
jordigilh: Same tension as the Goals conciseness thread — answering the guardrails door-open question added a paragraph here rather than trimming one. Let me know if you'd rather I move that explanation to a footnote/Open Question instead of inline.

docs/proposals/00715_semantic-classification-routing.md
jordigilh: Reworded the opener to lead with the technical gap (`intelligent_route` matches only structured fields, confirmed via code-search) instead of "customer named" — same intent as your suggestion, slightly different wording.

docs/proposals/00715_semantic-classification-routing.md
jordigilh: Reworked — Motivation now opens on the concrete capability gap and cites the vLLM Semantic Router precedent, no "users asked for this" framing or issue-link-as-justification left in the opening paragraph.

docs/proposals/00715_semantic-classification-routing.md
jordigilh: Trimmed both items — dropped the "reframed this on ai#750" narration and the "prior research found" framing, kept the substance. Can cut further if it's still too long.

docs/proposals/00715_semantic-classification-routing.md
jordigilh: You're right to flag this — the trailing sentence ("The `How?` section... per this repo's proposal convention") is process metadata, not motivation, and it's redundant with the `graduation_criteria` frontmatter that already says the same thing. I'll cut it from Motivation.

jordigilh:
Thanks for the reframing — checking back against the current revision, all four points are already folded in:

> Praxis should own the callout/integration, routing policy, stickiness, fallback, and final enforcement. llm-d-sc should own model/tokenizer loading, batching, inference, calibration, abstention, caching/optimization (let it be stateful), and the classifier runtime itself.

Reflected in Open Question 1: the classifier runs as a separate sidecar service specifically so Praxis doesn't own anchors, thresholds, calibration, or model versioning — that ML lifecycle stays on the classifier side.

> I don't think "embedding backend selection" should be a blocking Praxis design decision anymore... Praxis should consume the signal contract without caring whether that signal came from a fine-tuned encoder, embeddings, another optimized runtime, or eventually something larger.

Motivation item 1 now reads "~~No embedding backend chosen~~ — resolved, not blocking" — the filter consumes a technique-agnostic task/complexity signal regardless of backend.

> I'd also avoid making current-turn-only classification part of the contract... The API should be able to receive sufficient context for correctness while allowing llm-d-sc to cache tokenization/features/session-derived information for performance. If that state is lost... the classifier should abstain and Praxis should apply its fallback/stickiness behavior rather than guessing.

The signal contract is explicitly not restricted to current-turn-only — that's called out as a v1 *implementation default* (for cost), not an interface limit. The four-state outcome (`classified`/`unclassified`/`skipped_by_policy`/`unavailable`) is the abstention mechanism you're describing: if the classifier can't produce a confident answer, it reports that rather than guessing, and `intelligent_route`'s fallback-candidate config handles the "what happens next" you're describing as Praxis's job.

> I'd frame the initial architecture as local classification inside the trusted inference boundary rather than routing every prompt to an external embedding provider.

Quoted directly in Motivation now, with this framing credited back to this comment.

Let me know if any of these landed differently than you intended, particularly the ownership-split framing in Open Question 1 — that's the one I extrapolated the most rather than quoting directly.

Issue #758, opened 2026-08-17
Epic: Standalone AI Gateway MVP ("single-user ready")


usize:
### Summary

Ship a standalone Praxis AI Gateway container image that is immediately useful. Think, the way nginx is immediately useful as an HTTP server.

"Single-user ready" means one user but that user typically runs several apps and agents. The image therefore must include LiteLLM-style local API keys and per-key token ceilings as experimental filters.

What it does *not* mean: multi-tenant identity, distributed quota state, or billing-grade accounting. Those are tracked in epics (#78, #121, #301) and are explicitly out of scope here.

The "single-user-ready" image is also a vehicle for demonstrating Praxis's composable filter pipeline. Experimental filters ship from [praxis-proxy/experimental] and are documented as example implementations.

[praxis-proxy/experimental]: https://github.com/praxis-proxy/experimental
[`ghcr.io/praxis-proxy/ai`]: https://github.com/praxis-proxy/ai/pkgs/container/ai

## API Translation Current Limitations and Mitigation

Praxis translates in one client direction today: **Anthropic Messages
clients → OpenAI-format backends**. The reverse OpenAI-format clients → native Anthropic
backends does not exist, and is deliberately not in this MVP.

see: #363  

The most common use case the gap would serve -- "point your OpenAI SDK
at Claude" -- is already covered by [Anthropic's OpenAI compatibility
endpoint].

We should document this as a temporary mitigation.  Temporary because Anthropic positions the compatibility layer for testing and model comparison rather than long-term production use.

[Anthropic's OpenAI compatibility endpoint]: https://platform.claude.com/docs/en/api/openai-sdk

## Where Work Lands

Lands in `praxis-proxy/experimental`:

- Interim filters superseded by an owned production design (`api_key_auth` → ai#707/ai#708; `token_ceiling` → epic ai#121)
- Filters with pre-alpha dependencies (`switchyard_route`)
- The standalone image build and quickstart config

Lands in `praxis-proxy/ai` via existing issues, normal review:

- Bugfixes to shipped filters (guardrails: ai#753, ai#755, ai#673)
- Capabilities with an accepted epic home: quota-exhaustion detection under epic ai#249; translation gaps ai#296–298, ai#385 under epic ai#363

## Best-Effort / Stretch

Upstream issues that improve the image but are not strictly required:

- **[existing]** ai#298 — translate `tool_choice` for Chat
  Completions (highest-value of the translation gaps).
[...]

PR #727, opened 2026-08-12
Feat/http callout

gdozortsev:
## Summary

Continuation of https://github.com/praxis-proxy/ai/pull/484, refactored to use SubRequestConnector for routing. This allows us to make arbitrary API calls, passing selected fields from payloads and to parse any results. It's particularly useful for integrating with third party guardrails services.

This is currently blocked by: https://github.com/praxis-proxy/praxis/issues/964

PR #484, opened 2026-07-22
Feat/http callout

usize:
An HTTP Callout filter, utilizing core's reqwest based callout client.

This allows us to make arbitrary API calls, passing selected fields from payloads and to parse any results. It's particularly useful for integrating with third party guardrails services.

PR #760, opened 2026-08-17
feat(callout): add http_callout filter — supersedes #727

usize:
## Summary

Adds a generic `http_callout` HTTP filter: makes an outbound call during request
processing via core's `SubRequestConnector`, optionally reshapes the forwarded body
with JSONPath field mapping, extracts values from the JSON response into the
`FilterResultSet` for `branch_chains` evaluation, and can inject callout response
headers into the upstream request. Supports fail-open/fail-closed, per-target timeout,
circuit breaking, and callout-depth loop prevention.

This supersedes #727 (itself a continuation of #484) and was written by
@gdozortsev — her two commits are preserved as-authored; I've rebased them onto
main, added sign-offs, and stacked conflict/lint/scope fixes and extra hardening
on top. Thanks Gaby!

## Scope: Lakera Guard now, LlamaGuard later

The shipped example (`examples/configs/lakera-guard.yaml` + functional integration
tests) screens requests through Lakera Guard, which returns an exact boolean —
this works with today's exact-equality `on_result` matching and is the guardrails
integration required by the AI Gateway MVP (#758, success criterion 4; consumed by
the Track C quickstart).

A LlamaGuard example is deferred: its verdict format (`"unsafe\nS02"`) needs the
richer `on_result` matching proposed in praxis-proxy/praxis#964. This PR is NOT
blocked by praxis#964.

## Testing

- 73 unit tests: config validation (incl. nested `deny_unknown_fields`), SSRF
  checks, env-var expansion, `JSONPath` extraction coercion/sanitization, body
  shaping (incl. non-JSON raw-forward fallback), failure modes, circuit breaker,
  depth limit, phase handling, and header forwarding/injection (incl. absence).
- Functional integration tests for the example config: flagged → 403, clean → 200,
  and GET bypasses the POST-scoped callout.
- `make lint`, `make test`, `make doc`, `make audit` green locally.

Refs: #727, #484, #758; praxis-proxy/praxis#964


jordigilh:
Nice test coverage on this one (74 filter-level cases + `wiremock` integration). One security gap worth a look before merge, plus two minor notes.

**SSRF / DNS-rebinding:** `validate_callout_url` only checks the URL's literal host for private/loopback IPs — it warns, doesn't reject, and there's a test (`validate_url_warns_on_private_ip`) confirming that's deliberate. But `resolve_peer()` does a fresh DNS lookup per request with no check on the *resolved* address. A domain-based target (the normal case) that resolves to `169.254.169.254` or an internal service at request time bypasses the check entirely — untested. Given callout targets are usually domains, this is the more realistic SSRF vector than the literal-IP case that's currently guarded.

Minor, non-blocking:
- `max_body_bytes` also bounds the *response* read from the callout (via `SubRequestClient::execute`), but `build_subrequest_client` uses `SubRequestClient::new()` rather than `with_max_response_bytes`, so it silently clamps to the 64 MiB client-wide ceiling regardless of a higher configured value — worth a doc note or a `with_max_response_bytes` call so config matches behavior.
- `inject_headers` values from the callout response skip the same sanitization (`sanitize_string`) that JSONPath-extracted body values get — same untrusted origin, two trust levels.

senanz (approved):

senanz:
LGTM

filters/src/callout/extract.rs
```diff
@@ -0,0 +1,517 @@
...
+///
+/// Extracted values come from an untrusted third-party callout response
+/// and feed [`FilterResultSet`] entries, branch matching, and logs, so
+/// this is a defense-in-depth pass against control-character/log
+/// injection. It:
+///
+/// - strips leading control characters and whitespace (e.g. a leading `"\n\n"`),
+/// - keeps characters up to the first embedded control character and drops everything after it (multi-line verdicts
+///   such as `"unsafe\nS02"` collapse to their first line, `"unsafe"`),
+/// - drops `/` and `\` from the retained text,
```
aslakknutsen: Are they? 

split_at_first.. says /// `/` and `\` — are preserved verbatim so legitimate extracted values

docs/filters/http_callout.md
```diff
@@ -0,0 +1,38 @@
...
+| `target.timeout` | Duration | no | Request timeout (e.g. `"2s"`, `"500ms"`). |
+| `target.headers` | HeaderEntry[] | no | Static headers to send with every callout. |
+| `target.headers[].name` | string | yes | Header name. |
+| `target.headers[].value` | string | yes | Header value. Supports `${VAR}` env-var expansion. |
+| `target.forward_headers` | string[] | no | Headers to copy from the downstream request. |
+| `target.body` | object<string, string> | no | Reshape the downstream request body for the callout. Each key becomes a field in the callout JSON body; each value is a `JSONPath` expression evaluated against the downstream body. When set, only the listed fields are sent — the downstream body goes to upstream untouched. When absent, the downstream body is forwarded verbatim. |
+| `request` | RequestConfig | no | Request phase and body forwarding options. |
```
aslakknutsen, line 27: @shaneutt @alexsnaps Wouldn't it make sense to treat a callout as a type of sub/new filter_chain. That would allow to configure any of the other filters on a callout call and have a thin callout filter instead of a fat one trying to implement all the other filters...

usize: What sort of composition did you have in mind there?

We can indeed do matching on our results here and use branch chains to jump to a new filter pipeline. And we could add support for passing values on to downstream filters. 

Generic HTTP configuration here is necessary. JSONPath is where I start to see your point. We could have a separate JSONPath filter that consumes the output of a callout etc... though, at the same time, for most of the use cases here having JSONPath as a simple config option makes life much easier than composing with another filter and having to reason about the chain.

filters/src/callout/extract.rs
usize: Indeed, this is an old comment.

`test_sanitize_string_preserves_slashes` contradicts the comment here. Fixing. Thank you.

shaneutt (approved):
I understand some of the comments here, but I don't think we need to hold this up any longer.

@usize please ensure we have a build flag around this (bucketing under the standard `experimental` flag is fine) since it's work in progress, and so we can sort of draw a line around it and let it soak/iterate, and then LGTM.

Create follow-up issues (as needed) for unresolved comments.

docs/filters/http_callout.md
aslakknutsen: I was thinking conceptually to treat the Callout Request similar to the Proxy Incoming Request. Where target is the "router/cluster" endpoint hardcoded. The Callout response being the Proxy Response.

The filter it self could in theory just reference filter_chains by name like we do for Listeners. 

I'm just thinking what happens when you want the callout;
* signed
* compressed
* auth injected
etc etc..

aslakknutsen (approved):

Issue #765, opened 2026-08-18
security: cargo audit/deny fail on main — RUSTSEC-2026-0258 (h2 0.4.15 unbounded empty DATA frames DoS)


jordigilh:
## Summary

`cargo audit` / `cargo deny check` currently fail on `main` due to **RUSTSEC-2026-0258** in `h2` 0.4.15: unbounded empty DATA frames can be used to exhaust memory/CPU, a denial-of-service risk for a proxy that handles streaming responses.

## Evidence

- `Cargo.lock` pins `h2` at version `0.4.15`:

```1832:1835:Cargo.lock
name = "h2"
version = "0.4.15"
source = "registry+https://github.com/rust-lang/crates.io-index"
checksum = "6cb093c84e8bd9b188d4c4a8cb6579fc016968d14c99882163cd3ff402a4f155"
```

- `deny.toml`'s `[advisories]` ignore list does not exempt RUSTSEC-2026-0258, so `cargo deny check` treats it as blocking:

```1:10:deny.toml
[advisories]
unmaintained = "all"
maximum-db-staleness = "P30D"
ignore = [
    # Pingora transitive dependencies; not under our control.
    # derivative 2.2.0: unmaintained
    "RUSTSEC-2024-0388",
    # rustls-pemfile 2.2.0: unmaintained (via pingora-rustls)
    "RUSTSEC-2025-0134",
]
```

- `.github/workflows/supply-chain.yaml` runs `cargo audit` (`security-audit` job) and `cargo deny check` (`dependency-check` job) via the shared `praxis-proxy/conventions/.github/actions/supply-chain-audit` action on every PR/push/merge to `main`, so this currently blocks that job on `main`.

(Evidence fetched from `upstream/main` at commit `333f93d5e3aef60a86897412da617012fc20267e`, since file line numbers may drift over time.)

## Risk

`h2` is the HTTP/2 implementation underlying this proxy's streaming response path. Unbounded empty DATA frames let a malicious/misbehaving peer force unbounded buffering or CPU spend per stream, which is a direct denial-of-service exposure for a gateway that proxies long-lived, streaming (SSE/chunked) AI responses.

## Suggested fix

`cargo update -p h2` to `0.4.16`, which satisfies the existing `^0.4.15` requirement already implied by dependents' `Cargo.toml` constraints, so this should be a `Cargo.lock`-only change with no `Cargo.toml` edits required.

[...]

PR #767, opened 2026-08-18
lint: add diff-scoped extended-lint heuristic checks

### LINKED ISSUE #766: lint: add diff-scoped extended-lint heuristic checks (TODOs, commented-out code, comment slop, literal repetition)
## Summary

Add a diff-scoped "extended lint" script that catches low-effort-code smells Clippy can't check structurally — leftover TODO/FIXME markers, commented-out code, comments that just narrate the next line, and repeated literals that should be named constants — and wire it into `make lint` / CI alongside the existing `cargo xtask lint-*` checks.

## Motivation

This repo's `[workspace.lints]` tables already deny the machine-checkable half of low-quality-code patterns: `unwrap_used`, `expect_used`, `panic`, `todo`, `unimplemented`, `dead_code`, `missing_docs`, `print_stdout`/`print_stderr`, `dbg_macro`, and more. But two common low-effort-code tells are structurally out of Clippy's reach:

- **Comment content** — a leftover `// TODO: fix this later`, or a comment that's just commented-out code, or a comment that narrates "what" a line does instead of explaining *why* (this repo's own `CONTRIBUTING.md` already calls out narrating comments as a convention to avoid).
- **Diff-local repetition** — the same numeric/string literal typed 3+ times in a single PR's added lines instead of being hoisted into a named constant.

Six Praxis repos are formalizing a shared, org-owned version of this idea; this issue tracks the `praxis-proxy/ai` instance.

## Proposed implementation

A small, dependency-free `scripts/extended-lint.py` that:

1. Runs `git diff --unified=0 <diff-base> -- '*.rs'` and only inspects **added** lines, so pre-existing code is never relitigated.
2. **BLOCKs** (non-zero exit) on:
   - Leftover `TODO`/`FIXME`/`XXX`/`HACK` markers in comments
   - Commented-out code (heuristic pattern match on `let`/`fn`/`if`/`for`/`match`/`return`/call-expression-shaped comments)
3. **WARNs** (printed, does not fail) on:
   - "Narrating" comments (`// increment the counter`, `// returns the result`, etc.) — prefer self-explanatory code or a doc comment on *why*
   - The same literal repeated 3+ times in a file's added lines without a corresponding `const`/`static` declaration
   - Weak/generic identifier names (`tmp`, `foo`, `val`, ...) introduced by a new binding
   - A newly added `#[allow(clippy::...)]`/`#[expect(clippy::...)]` suppression, as a prompt to double-check it has a real justification
4. Resolves its diff base the same way `.github/workflows/unicode-safety.yaml` already does: the PR's base SHA on `pull_request` events, or `git merge-base origin/main HEAD` otherwise — no new CI mechanism invented.
5. Is wired into `make lint` (alongside `cargo xtask lint-deps`/`lint-separators`/`lint-markdown-links`/etc.) so it runs automatically in the existing `tests.yaml` lint job.

See the implementation PR for the full script and CI wiring.

**Severity:** N/A (tooling/process improvement)

jordigilh:
## Summary

Adds a `cargo xtask lint-extended` subcommand, a diff-scoped heuristic checker for low-effort-code smells that Clippy can't catch structurally — comment *content* and diff-local *repetition* — and wires it into `make lint` and CI.

Closes #766

This was originally implemented as a Python script (`scripts/extended-lint.py`); Python is the wrong choice of language for tooling in this Rust-native codebase, so it's now a proper Rust `xtask` subcommand instead, matching this repo's existing `lint-deps`/`lint-separators`/`lint-filter-docs`/`lint-example-tests`/`lint-markdown-links` xtask modules in CLI dispatch (clap derive enum), error handling (`eprintln!` + `std::process::exit`, no custom `Result` type — this crate's own convention), and code structure.

## What's covered

Only lines **added or changed** versus the diff base are scanned, so pre-existing code is never relitigated:

- **BLOCK:** leftover `TODO`/`FIXME`/`XXX`/`HACK` markers, commented-out code
- **WARN:** narrating "what" comments, repeated literals without a named constant, weak/generic identifier names, new clippy suppressions

## CI wiring

- `make lint` now runs `cargo xtask lint-extended` in place of the former `python3 scripts/extended-lint.py`, alongside the existing `cargo xtask lint-deps`/`lint-separators`/`lint-filter-docs`/`lint-example-tests`/`lint-markdown-links` checks, so it runs automatically in `tests.yaml`'s existing `lint` job.
- The diff base is resolved the same way `.github/workflows/unicode-safety.yaml` already resolves its base ref (PR base SHA on `pull_request` events, `git merge-base origin/main HEAD` otherwise) — no new mechanism invented, just exported as `EXTENDED_LINT_BASE`, which `cargo xtask lint-extended` falls back to (after a `--base` CLI flag) when resolving its diff base.
- `lint` job's checkout still uses `fetch-depth: 0` (mirroring `unicode-safety.yaml`), since the diff needs history beyond the default shallow clone.
- `CONTRIBUTING.md`'s "Rules, Practices & Lints" section now points at `cargo xtask lint-extended` instead of the removed script.

## Testing

- `cargo clippy --workspace --all-targets -- -D warnings`, `cargo +nightly fmt --all -- --check`, and `cargo test -p xtask` all pass, including 8 new unit tests for the `lint_extended` module.
- Sanity-tested locally with a throwaway commit containing a deliberate `// TODO: fix this` line — confirmed `cargo xtask lint-extended` correctly BLOCKs (exit 1) and reports the file:line. That test commit is not included in this PR.

## Context

This formalizes a clean, org-owned version of an ad hoc personal script idea across several Praxis repos; this PR is the `praxis-proxy/ai` instance.

### DIFF
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -40,6 +40,8 @@ jobs:
       contents: read
     steps:
       - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
+        with:
+          fetch-depth: 0
 
       - name: Install actionlint
         uses: ./.github/actions/install-actionlint
@@ -54,6 +56,18 @@ jobs:
 
       - name: Install cargo-machete
         run: cargo install cargo-machete --locked
+
+      - name: Resolve extended-lint diff base
+        shell: bash
+        run: |
+          set -euo pipefail
+          if [ "${{ github.event_name }}" = "pull_request" ]; then
+            BASE_SHA="${{ github.event.pull_request.base.sha }}"
+          else
+            git fetch origin main
+            BASE_SHA="$(git merge-base origin/main HEAD)"
+          fi
+          echo "EXTENDED_LINT_BASE=${BASE_SHA}" >> "$GITHUB_ENV"
 
       - name: Lint
         run: make lint
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -132,6 +132,16 @@ Security is enforced at the lint level. See lints in
 [Cargo.toml]:Cargo.toml
 [getting-started.md]:docs/developing/getting-started.md
 
+### Extended Lint
+
+`make lint` also runs `cargo xtask lint-extended`, a diff-scoped
+heuristic check for low-effort-code smells that Clippy can't catch
+structurally: leftover `TODO`/`FIXME`/`HACK` markers, commented-out
+code, "narrating" comments that just restate the next line, and
+literals repeated 3+ times in a PR's added lines instead of being
+hoisted to a named constant. It only scans lines added or changed
+versus the diff base, so pre-existing code is never relitigated.
+
 ### Lint Suppression Policy
 
 Use `#[expect(...)]` instead of `#[allow(...)]`. The
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -6074,6 +6074,7 @@ dependencies = [
  "praxis-proxy-core",
  "praxis-test-utils",
  "quote",
+ "regex",
  "reqwest 0.13.4",
  "serde",
  "serde_json",
--- a/Makefile
+++ b/Makefile
@@ -110,6 +110,7 @@ lint:
 	cargo xtask lint-filter-docs
 	cargo xtask lint-example-tests
 	cargo xtask lint-markdown-links
+	cargo xtask lint-extended
 	cargo xtask sync-example-readme
 	cargo xtask sync-inference-readme
 	cargo xtask sync-responses-readme
--- a/xtask/Cargo.toml
+++ b/xtask/Cargo.toml
@@ -21,6 +21,7 @@ praxis-ai = { path = "../server", package = "praxis-ai-proxy" }
 praxis-ai-apis = { workspace = true }
 praxis-core = { workspace = true }
 praxis-test-utils = { workspace = true }
+regex = { workspace = true }
 reqwest = { workspace = true, features = ["blocking"] }
 serde = { workspace = true }
 serde_json = { workspace = true }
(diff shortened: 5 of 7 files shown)


jordigilh:
Note: `security-audit` and `dependency-check` are failing here due to the pre-existing RUSTSEC-2026-0258 (`h2` 0.4.15) vulnerability on `main`, tracked separately in #765 — unrelated to this PR's diff (only `scripts/extended-lint.py` + CI wiring + a `CONTRIBUTING.md` mention). The `lint` job (which now runs the new script) passes.

jordigilh:
Closing this out without merging.

After digging into what actually exists in the Rust ecosystem for this class of check, most of the "extended lint" heuristics either have no reliable off-the-shelf equivalent in any language's tooling (commented-out-code detection, "narrating" comment detection), or only partial overlap that does not justify maintaining bespoke regex/diff-scoped Rust across six repos (e.g. TODO markers are trivially a one-line `git diff | rg` in CI rather than a dedicated tool; new-clippy-suppression tracking is already better covered by denying `clippy::allow_attributes` outright rather than diff-scoping it).

Decision: we do not want to own and maintain custom heuristic lint logic per-repo. If this capability is worth having, the better path is contributing it upstream to a maintained linter (clippy itself, or a `dylint`/`ast-grep` rule set) rather than a bespoke `xtask` subcommand we would need to keep in sync across the org. Not pursuing this further for now.

jordigilh:
Follow-up on the reasoning above: this PR is a live example of exactly that friction. Its `lint` CI check is still failing as of closing (https://github.com/praxis-proxy/ai/actions/runs/32174983883/job/95834699914), blocked on the tool's own source lines -- the regex definitions and test fixtures in `xtask/src/lint_extended.rs` literally contain `TODO`/`FIXME`, and unlike the other five repos' ports, this one never added the self-exclusion pathspec needed to stop it blocking on itself. The fix itself is small, but it is exactly the kind of one-off follow-up someone would need to notice and chase down before this could be safely turned on -- reinforcing the decision to close rather than merge.

PR #769, opened 2026-08-18
feat(filter): add jwt_auth filter with JWKS validation

### LINKED ISSUE #708: feat(filter): add JWT/OIDC authentication filter with JWKS validation
## Summary

Add a filter that validates JWT bearer tokens against a JWKS endpoint (Keycloak, any OIDC provider), extracts configured claims into `filter_metadata`, and strips the token header before upstream forwarding.

## Motivation

Enterprise deployments use SSO/OIDC for identity — users authenticate via Keycloak or another IdP and receive JWTs. The gateway needs to:

1. **Validate** the JWT signature against the issuer's JWKS endpoint
2. **Verify** standard claims (issuer, audience, expiration)
3. **Extract** configured claims (username, groups, subscription) into `filter_metadata`
4. **Strip** the token header so JWTs don't leak to upstream LLM providers

This is the SSO/enterprise counterpart to API key validation (#707) — both write the same `filter_metadata` contract, so downstream filters work unchanged regardless of which auth mechanism authenticated the request.

## Key design points

- **JWKS caching** with TTL-based refresh and unknown-kid-triggered refresh (with cooldown to prevent stampede)
- **Configurable claim extraction** — operator maps JWT claims to metadata keys (e.g. `preferred_username` → `username`, `groups` → `groups`)
- **Azure AD compatibility** — infers algorithm from JWK key type when the `alg` field is missing (common with Azure AD)
- **Writes to `filter_metadata`** — same contract as `api_key_auth` (#707): `username`, `groups`, `subscription`
- **Strips token header** before upstream send — prevents JWT leakage to providers
- **Supports both `Authorization: Bearer` and raw header** (e.g. `x-api-key`) token formats

## Relationship to other filters

- Alternative to `api_key_auth` (#707) — different auth mechanism, same metadata output
- Works with `credential_injection` (core builtin) for provider key injection
- Works with `identity_header_guard` (#698) for identity header capture
- Works with `external_metering` (#577) which reads identity from metadata

## Scope

- Configurable JWKS URL, issuer, audience, claim mappings, token header
- JWKS cache with TTL refresh + cooldown on failure
- ~500 lines of filter code
- Unit tests with pre-generated RSA keypair + wiremock JWKS server (valid/expired/bad-sig/wrong-issuer/unknown-kid/cache-refresh)
- Integration test, example config
- Working implementation with tests available on my fork (`yossiovadia/ai`, branch `feat/dogfood-gateway`)

## Non-goals

- No token issuance or refresh — that's the IdP's responsibility
- No RBAC/authorization beyond claim extraction — filters downstream decide policy
- No OIDC discovery (`.well-known/openid-configuration`) — JWKS URL is configured directly

yossiovadia:
## What

Adds a `jwt_auth` filter that validates JWT bearer tokens against a JWKS endpoint and injects verified claims as request headers.

- Fetches and caches JWKS from a configured URL, keyed by `kid`
- Validates signature, expiry, and issuer
- Maps configured claims to request headers and filter metadata
- Strips the bearer token before forwarding upstream
- Supports RS256/384/512 and ES256 (via `jsonwebtoken` with `rust_crypto`)

## Why / relationship to #708

This is the temporary bridge discussed in **#708**. As noted there, JWT validation is not the long-term target architecture — the core CPEX engine is the intended home. But this filter is fully self-contained and unblocks JWT/OIDC-fronted deployments today. Per the discussion, no other filter should take a dependency on it, so it can be cleanly retired once CPEX covers this path.

Closes #708

## Tests

- 11 unit tests: valid token passes and writes metadata, bearer-prefix extraction, token stripped from upstream request, expired / wrong-issuer / unknown-kid / garbage / missing tokens rejected, config validation (empty JWKS URL, empty claim headers, unknown fields)
- 4 integration tests (`tests/integration/.../examples/jwt_auth.rs`): config parses, valid token allowed + stripped, missing/garbage tokens rejected
- Example config (`examples/configs/jwt-auth.yaml`), filter docs (`docs/filters/jwt_auth.md`), reference table entry

`cargo test`, `cargo clippy`, and `cargo +nightly fmt --check` all pass locally.

## Notes for reviewers

- **Test key material:** `filters/src/jwt_auth/test_fixtures/rsa_private.pem` is a throwaway RSA keypair generated solely to sign tokens in unit tests (`include_str!`, clearly labeled test-only). It signs nothing real and protects nothing — happy to switch to in-memory keygen if you'd prefer no committed key.
- **Cargo.lock:** adds only `jsonwebtoken`'s `rust_crypto` transitive crypto crates (rsa, ecdsa, ed25519, p256/p384, der, spki, etc.). `jsonwebtoken 10.4.0` was already in the lock via the CPEX subtree; the diff is add-only (no packages removed).


yossiovadia:
Pushed a round of security hardening from a self-review:

- TLS certificates are now **verified by default** on the JWKS fetch, with an opt-in `insecure_skip_tls_verify` escape hatch (logs a warning) for self-signed in-cluster IdPs; non-https/non-loopback JWKS URLs warn at startup.
- Renamed `claim_headers` → `claim_metadata` to match actual behavior — claims are written to `filter_metadata`, never injected as upstream request headers.
- Single-flight JWKS refresh (no stampede on cache miss) and bounded reads (Content-Length + running chunk cap).
- Added forged-signature rejection and audience match/mismatch tests.

One CI caveat: `cargo xtask generate-filter-docs` / `sync-example-readme` couldn't run in my local env (pre-existing compile break in a vendored dep on the pinned toolchain), so I hand-synced the generated docs. If `lint-filter-docs` flags formatting, it's a one-command regen.

examples/configs/jwt-auth.yaml
```diff
@@ -0,0 +1,50 @@
...
+  - name: gateway
+    address: "127.0.0.1:8080"
+    filter_chains:
+      - authenticated
+
+filter_chains:
+  - name: authenticated
+    filters:
+      - filter: jwt_auth
+        jwks_url: "http://127.0.0.1:8280/realms/ai-gateway/protocol/openid-connect/certs"
```
rikatz, line 36: I would probably accept this configuration just if a flag is also set to allow http instead of https. Overall, IMO the intention of the user of not using https for something like a jwks endpoint should be a very conscious decision

docs/filters/jwt_auth.md
```diff
@@ -0,0 +1,28 @@
...
+Validates JWT bearer tokens against a JWKS endpoint and writes verified claims to `filter_metadata`.
+
+## Configuration
+
+| Field | Type | Required | Description |
+|-------|------|---------|-------------|
+| `jwks_url` | string | yes | URL of the JWKS endpoint (JSON Web Key Set). The filter fetches public keys from here to verify JWT signatures. This channel is the filter's root of trust — it should be `https://` in any deployment where the proxy-to-`IdP` path is not already authenticated. A plain `http://` URL (other than loopback) is accepted but logs a warning at startup. |
+| `issuer` | string | no | Expected `iss` (issuer) claim. If set, tokens from other issuers are rejected. |
+| `audience` | string | no | Expected `aud` (audience) claim. If set, tokens not intended for this audience are rejected. |
+| `claim_metadata` | object<string, string> | no | Maps JWT claim names to `filter_metadata` keys. The filter extracts these claims from verified tokens and writes them to `filter_metadata` for downstream filters (e.g. `external_metering`). Claims are intentionally NOT injected as upstream request headers — see the filter docs for the trusted-channel rationale. |
```
rikatz, line 15: maybe not for now, but the metadata can be an arbitrary set of object<string, object>

As an example, Kubernetes JWT token for a SA can be decoded into:

```json
{
  "aud": [
    "https://kubernetes.default.svc.cluster.local"
  ],
  "exp": 1787326909,
  "iat": 1787323309,
  "iss": "https://kubernetes.default.svc.cluster.local",
  "jti": "170964fe-f262-45f7-be60-594c62e70c7a",
  "kubernetes.io": {
    "namespace": "default",
    "serviceaccount": {
      "name": "mygateway",
      "uid": "48c32fc6-a3c8-4e14-ad09-6219c381d4b4"
    }
  },
  "nbf": 1787323309,
  "sub": "system:serviceaccount:default:mygateway"
}
```

I am not sure if it would make sense to support some more arbitrary structure, or something like a "jsonpath'ish" for these metadata

docs/filters/jwt_auth.md
```diff
@@ -0,0 +1,28 @@
...
+
+# `jwt_auth`
+
+Validates JWT bearer tokens against a JWKS endpoint and writes verified claims to `filter_metadata`.
+
+## Configuration
+
+| Field | Type | Required | Description |
+|-------|------|---------|-------------|
+| `jwks_url` | string | yes | URL of the JWKS endpoint (JSON Web Key Set). The filter fetches public keys from here to verify JWT signatures. This channel is the filter's root of trust — it should be `https://` in any deployment where the proxy-to-`IdP` path is not already authenticated. A plain `http://` URL (other than loopback) is accepted but logs a warning at startup. |
```
rikatz, line 12: we will not now (or never) support symmetric JWT tokens, right? like a shared signing key. I will try to take a look if there is some existing architectural proposal for it, but though would be good to check

filters/src/jwt_auth/test_fixtures/rsa_private.pem
```diff
@@ -0,0 +1,28 @@
+-----BEGIN PRIVATE KEY-----
```
rikatz, line 1: this is very nit, but a fixture private key may cause lots of alarms on code scanners. Just consider maybe if it is possible to generate this during the test and cleanup after the test

filters/src/jwt_auth/config.rs
```diff
@@ -0,0 +1,88 @@
...
+// -----------------------------------------------------------------------------
+// Validation
+// -----------------------------------------------------------------------------
+
+/// Validate a [`JwtAuthConfig`], returning an error on missing required fields.
+pub(super) fn validate_config(config: &JwtAuthConfig) -> Result<(), String> {
+    if config.jwks_url.is_empty() {
+        return Err("jwt_auth: jwks_url must not be empty".into());
+    }
+    if config.claim_metadata.is_empty() {
```
rikatz, line 84: is this really required? I mean, a inner filter could use just the subject for authorization

filters/src/jwt_auth/jwks.rs
```diff
@@ -0,0 +1,274 @@
...
+        } else if !is_authenticated_url(&url) {
+            warn!(
+                %url,
+                "jwt_auth: JWKS URL is not https:// and not loopback — key delivery is \
+                 unauthenticated and vulnerable to MITM key substitution"
+            );
+        }
+
+        let client = reqwest::Client::builder()
+            .timeout(Duration::from_secs(10))
```
rikatz, line 106: maybe make this configurable on the filter?

filters/src/jwt_auth/jwks.rs
```diff
@@ -0,0 +1,274 @@
...
+    /// If a refresh fails (e.g. the `IdP` is unreachable), the last
+    /// good cache is served past its TTL rather than failing closed —
+    /// a stale-if-error availability tradeoff. This means a revoked
+    /// key can keep validating for as long as the `IdP` stays down;
+    /// tokens still expire independently via `exp`.
+    pub(super) async fn get_key(&self, kid: &str) -> Option<(DecodingKey, Vec<Algorithm>)> {
+        // Fast path: key is cached and TTL hasn't expired.
+        {
+            let keys = self.keys.read().await;
+            let ttl_ok = keys.last_refresh.is_some_and(|t| t.elapsed() < DEFAULT_TTL);
```
rikatz, line 143: nit (possible followup)

I think it makes sense to have the TTL also configurable. Overall:
* An insecure environment, public you may want to have the TTL lowered, in case you need to revoke the keypair signing the JWT tokens you can react fast
* A safe airgapped cluster could probably use of some bigger refresh time for the JWKS

filters/src/jwt_auth/jwks.rs
```diff
@@ -0,0 +1,274 @@
...
+            if ttl_ok && let Some(entry) = keys.by_kid.get(kid) {
+                return Some(entry.clone());
+            }
+        }
+
+        // Cooldown gate (lock-free): if we attempted a refresh very
```
rikatz: I was wondering if the cooldown for failures should be somehow an exponential backoff until you hit the max REFRESH_COOLDOWN. 

my concern here is that some unavailability on the jwks server will block requests for 30s, while those can be transitive. This could cause an overall unavailability of your authentication workflow.

yossiovadia: Good call — implemented in f2c78d8. A *successful* refresh keeps the 30s anti-stampede cooldown, but a *failed* one now backs off starting at 1s and doubling per consecutive failure, capped at 30s. So a transient JWKS blip recovers in ~1s instead of blocking auth for a full 30s, while a sustained outage still isn't hammered. Added a test that proves recovery after the short backoff (would fail under the old flat cooldown).

filters/src/jwt_auth/jwks.rs
```diff
@@ -0,0 +1,310 @@
...
+                keys.by_kid.get(kid).cloned()
+            },
+            Err(e) => {
+                warn!("JWKS refresh failed: {e}");
+                // Grow the backoff (capped) so a transient failure is
+                // retried quickly but a sustained outage isn't hammered.
+                let mut keys = self.keys.write().await;
+                keys.consecutive_failures = keys.consecutive_failures.saturating_add(1);
+                let shift = (keys.consecutive_failures - 1).min(5);
+                keys.retry_after = FAILURE_BACKOFF_INITIAL.saturating_mul(1 << shift).min(REFRESH_COOLDOWN);
```
rikatz, line 188: not sure if this was just added or if I missed it, but resolves my comment on the backoff, thanks!

filters/src/jwt_auth/mod.rs
```diff
@@ -0,0 +1,304 @@
...
+            issuer: config.issuer,
+            audience: config.audience,
+            claim_metadata,
+            token_header: config.token_header.to_lowercase(),
+        }))
+    }
+
+    /// Extract the JWT from the configured header.
+    ///
+    /// Handles both `Authorization: Bearer <token>` and raw
```
rikatz, line 141: more like a consistency concern: would it make sense to not make user assumptions, but instead explicit choices?

As an example, per [RFC-6750](https://datatracker.ietf.org/doc/html/rfc6750#section-2.1) the authorization header is always of format `"Bearer" 1*SP b64token` so a header like "x-api-key: bearer xpto" doesn't really match this expectations.

I would put on a different way:
* If the token_header is named "Authorization" you expect its value to be "Bearer <1 space only> value"
* If the token is something else, you get its value directly

In case some other format is required I would consider it a followup.

filters/src/jwt_auth/mod.rs
```diff
@@ -0,0 +1,304 @@
...
+                    other => other.to_string(),
+                };
+
+                ctx.set_metadata(metadata_key.clone(), metadata_value);
+            }
+        }
+
+        debug!(
+            username = claims
+                .get("preferred_username")
```
rikatz, line 281: this is some info from previous debug, or left intentionally?

shaneutt (changes requested):
(some ongoing discussion)

araujof (commented):
Thanks for this PR, @yossiovadia ! 

I [commented](https://github.com/praxis-proxy/ai/issues/708#issuecomment-5667951287) on the related issue. I think that we could descope this PR to integrating the Policy filter `AuthenticatedIdentity` with Praxis AI consumers such as `external_metering`, instead of adding a separate JWT/JWKS filter.

cc: @leseb @shaneutt

leseb:
> Thanks for this PR, @yossiovadia !
> 
> I [commented](https://github.com/praxis-proxy/ai/issues/708#issuecomment-5667951287) on the related issue. I think that we could descope this PR to integrating the Policy filter `AuthenticatedIdentity` with Praxis AI consumers such as `external_metering`, instead of adding a separate JWT/JWKS filter.
> 
> cc: @leseb @shaneutt

I fully agree, @yossiovadia please take that route instead. I'm closing this, feel free to re-open if you have concerns. Thanks!

Issue #129, opened 2026-05-15
Per-header rate limit bucket keys


shaneutt:
Extend the rate_limit filter to key token buckets by arbitrary header value, enabling per-tenant and per-user rate limiting.

## Requirements

- Configure a header name as the bucket key source
- Each unique header value gets its own token bucket
- Fall back to global bucket if header absent (configurable)
- Support multiple key dimensions (e.g. tenant + API key)
- Bucket eviction for inactive keys to bound memory

## Implementation Notes

- Extend existing rate_limit filter config with `bucket_key_header` field
- Use concurrent map for bucket storage
- Memory-bounded: configurable max distinct keys

Parent epic: praxis-proxy/ai#301
Related: praxis-proxy/ai#121

leseb:
Duplicate of #123, which owns per-header and composite token-rate-limit bucket keying and has the active implementation discussion.

PR #773, opened 2026-08-18
[DRAFT] token_rate_limit: per-app token budgets via bucket_key_header (demo vehicle, not a merge candidate)

jordigilh:
## Status: draft / not for merge

This is a **spike/demo branch**, not a merge candidate. It exists to give a
concrete, runnable example of per-app token budgeting while [ai#658](https://github.com/praxis-proxy/ai/pull/658)
(the actual Token Rate Limiting proposal) is still under active review with
open design questions. Opening as draft so the code is available to try out
without implying it's ready to land.

## What this does

Implements a `token_rate_limit` HTTP filter with reservation-based admission
(reserve an estimated token cost at request time, reconcile against actual
provider-reported usage after the response completes) and, on top of that,
[ai#129](https://github.com/praxis-proxy/ai/issues/129)'s `bucket_key_header`:
an optional config field that keys independent buckets off a request header
value (e.g. one budget per app/tenant), falling back to a single shared
bucket when the header is configured but absent on a given request.

See `examples/configs/token-rate-limit-per-app.yaml` for a runnable example,
and `filters/src/token_rate_limit/mod.rs`'s module doc for the full list of
what's implemented vs. deliberately deferred (composite/CEL keys, configurable
estimation, token-type-aware accounting, observability, metering).

Covered by unit tests (`filters/src/token_rate_limit/`) and an integration
test exercising the full HTTP round trip against the example config
(`tests/integration/tests/suite/examples/token_rate_limit.rs`), including a
scenario with multiple independently-budgeted apps sharing one filter
instance.

## Important: algorithm divergence from ai#658's current design doc

This needs to be called out plainly, since it's the main thing I'd want a
maintainer's read on before anyone treats this as more than a demo:

**What's likely reusable regardless of outcome:** the per-key bucket
*architecture* in `state.rs` -- resolve a key from a configurable header,
look it up in a keyed map, fall back to a shared bucket when the header's
absent, evict idle keys -- is orthogonal to which rate-limiting algorithm
sits behind each key. Whichever algorithm `ai#658` lands on, it will still
need "one independent rate-limit state per app/tenant, keyed by a header,
with a fallback and eviction," and this shape is meant to be a reasonable
starting point for that.

**What's a known, deliberate divergence:** the algorithm actually
implemented here is a **token bucket** (`rate`/`burst`/`estimate_tokens`
config), reusing the pattern from praxis's existing `rate_limit` filter.
`ai#658`'s current design doc specifies a **sliding-window** model instead.
I did not implement sliding window here because:

- The core primitive doesn't exist yet -- [praxis#551](https://github.com/praxis-proxy/ai/issues/789)
  ("Sliding window rate limiting") is still open in the core `praxis` repo.
  Implementing it for real would mean building that primitive from scratch
  as part of this branch, not reusing an existing building block.
- The sliding-window design itself isn't settled -- it currently only
  appears in the proposal doc on `ai#658`'s own PR branch, and that PR still
  has open, unresolved review threads.

To be direct: this is **not** claiming compliance with `ai#658`'s design.
The algorithm and config shape here will very likely need to be rebuilt
once the token-bucket-vs-sliding-window question is settled on `ai#658`.
I'm not trying to preempt that decision -- just to unblock a working demo
in the meantime.

## Ask

No expectation this merges as-is. If it's useful as something to point demo
tooling at while `ai#658` review continues, great -- otherwise happy to let
it sit as reference or close it once `ai#658` lands with its own approach.


jordigilh:
Closing this draft — moving the demo material to praxis-proxy/experimental instead. The code stays available on this branch (jordigilh:jordigilh/token-rate-limit-per-app-budgets) for anyone who wants to build it from source; no expectation of merging it here while ai#658's design is still settling.

PR #782, opened 2026-08-19
fix(token_usage): recover from SSE overflow instead of silently dropping usage

### LINKED ISSUE #674: Token accounting silently stops after fixed JSON or SSE capture limits are exceeded
## Clawpatch finding

- **Severity:** high
- **Category:** data-loss
- **Confidence:** high
- **Triage:** confirmed-bug
- **Feature:** `feat_custom_ai_anthropic_guardrails_prompt_token`
- **Finding ID:** `fnd_sig-feat-custom-ai-anthropic-gua_35238e0300`
- **Run ID:** `20260810T073440-229564`
- **Prior-run overlap:** None detected in preserved prior .clawpatch snapshot.

## Evidence

- `filters/src/token_usage/count.rs:49-53` — JSON capture is fixed at 1 MiB and SSE scratch state at 64 KiB.
- `filters/src/token_usage/count.rs:312-316` (`handle_json_body`) — On JSON overflow all token_count working metadata, including extraction mode, is cleared.
- `filters/src/token_usage/count.rs:352-365` (`handle_sse_body`) — On SSE scanner overflow accumulated counts are finalized and all working metadata is cleared.
- `filters/src/token_usage/count/tests.rs:595-618` (`json_overflow_sets_nothing`) — The test codifies that an oversized JSON response produces no token metadata.
- `filters/src/token_usage/count/tests.rs:541-564` (`sse_overflow_finalizes_partial_counts`) — The test verifies overflow clears scanner state after only the usage seen before overflow.

## Description

Usage fields are commonly located at the end of a non-streaming JSON response or in a terminal SSE event. A JSON response larger than 1 MiB clears the mode before the trailing usage object can be parsed. Likewise, one SSE event or incomplete line exceeding 64 KiB clears `token_count.mode`, so all later events—including the provider's terminal usage event—are ignored. Large generated text or tool arguments can therefore cause missing or stale counts, which can undercount billing and monitoring data without any explicit failure signal.

## Reproduction

For OpenAI SSE, send a successful event stream whose first `data:` event exceeds 65,536 bytes, followed by a normal terminal event containing `usage`. The overflow path clears `token_count.*`; the terminal event is ignored and no `token.input`, `token.output`, or `token.total` metadata is produced. For JSON, return a valid response larger than 1 MiB with `usage` near the end and observe the same missing metadata.

## Recommendation

Do not disable extraction silently on overflow. For SSE, recover at the next event boundary and continue scanning for terminal usage events without retaining the oversized payload. For JSON, use a configurable bounded strategy that can still locate usage or publish an explicit overflow/unavailable result that billing consumers cannot mistake for zero usage. If accurate accounting is mandatory, fail closed rather than silently dropping counts.

## Suggested regression test

Send an oversized non-usage SSE event followed by a valid terminal usage event and assert the terminal counts are extracted. Add a valid JSON response exceeding the capture threshold with usage at the end and assert either accurate counts or an explicit, documented overflow failure signal.

---
Reported from scheduled Clawpatch Codex/xhigh review after rebasing `/Users/leseb/Documents/AI/ai` to `origin/main` (`1cc0a47`). The requested historical path `filter/src/builtins/http/ai/` is stale in this checkout; the review mapped it to the current AI implementation under `apis/src` and `filters/src` plus mapped filter-core context files. Please validate before assigning or starting implementation.

mkoushni:
## Summary

Fixes #674. Token accounting stopped permanently after a single oversized SSE event or JSON response, silently clearing all working state — including any terminal usage event still to come — with no signal that data was lost.

- The shared SSE scanner (`filters/src/agentic/a2a/sse.rs`, used by both `token_usage` and the A2A task-route capture filter) now discards only the oversized event and resumes scanning at the next event boundary, instead of aborting the whole stream. A terminal usage event arriving after an oversized one is now captured correctly.
- For JSON responses, and for the residual SSE case where the usage event itself is the one that overflows, a `token.status = overflow` metadata key and `Praxis-Token-Status` response header are now set explicitly, so billing/monitoring consumers cannot mistake missing counts for a genuine zero-usage response.
- `max_body_bytes` / `max_scratch_bytes` are now configurable per `token_count` filter instance instead of fixed at 1 MiB / 64 KiB.

## Changes

- `filters/src/agentic/a2a/sse.rs`: shared scanner recovers at the next blank-line event boundary instead of returning an abort signal; adds `SkipPhase` state tracking (discarded bytes are never buffered).
- `filters/src/agentic/a2a/mod.rs`: task-route SSE capture adapted to the new scanner contract — capture no longer disables itself after one oversized event.
- `filters/src/token_usage/count.rs`: `handle_sse_body` no longer aborts on overflow; `handle_json_body` and `finalize_streaming_counts` emit an explicit overflow status instead of silence; capture limits are now filter config fields.
- `filters/src/token_usage/headers.rs`: injects `Praxis-Token-Status` when set.
- Regression tests added matching the issue's reproduction and suggested tests (oversized non-usage SSE event followed by a valid terminal usage event; oversized JSON with usage near the end).

## Test plan

- [x] `cargo test --workspace` — all tests pass (1080 filters/apis lib tests, plus proxy/server/main suites)
- [x] `cargo clippy --workspace --all-targets` — clean
- [x] `cargo fmt --check` — clean for changed files
- [x] New regression tests cover: oversized SSE event followed by terminal usage event; oversized JSON with usage near the end; explicit overflow status distinguishable from zero; configurable limits


filters/src/agentic/a2a/sse.rs
```diff
@@ -40,37 +40,64 @@ pub(crate) struct SseScanState {
...
+    /// Not discarding; bytes are buffered normally.
+    #[default]
+    NotSkipping,
+    /// Discarding; the line since the last newline has been empty so
+    /// far. A newline seen now means a blank line — the boundary.
+    LineEmptySoFar,
+    /// Discarding; the line since the last newline has had at least
+    /// one byte. A newline seen now just ends that (non-blank) line.
+    LineHasContent,
 }
```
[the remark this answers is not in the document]
mkoushni: Fixed in f68a60ce. `SkipPhase::as_str` / `from_metadata_str` now live on the type in `sse.rs`, and both A2A and `token_count` use those helpers.

filters/src/token_usage/count.rs
```diff
@@ -101,6 +110,26 @@ const HEADER_BEDROCK_OUTPUT: &str = "x-amzn-bedrock-output-token-count";
 struct TokenCountConfig {
     /// AI provider whose response format to parse.
     provider: ProviderKind,
+
+    /// Maximum bytes to buffer for a non-streaming JSON response before
+    /// giving up on locating its usage field.
+    #[serde(default = "default_max_body_bytes")]
+    max_body_bytes: usize,
```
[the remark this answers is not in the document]
mkoushni: Fixed in f68a60ce. `from_config` now uses the shared `validate_max_body_bytes` helper (rejects 0 and the 64 MiB ceiling) and rejects `max_scratch_bytes: 0` the same way, matching other filters.

filters/src/token_usage/count.rs
```diff
@@ -437,9 +502,18 @@ fn merge_accumulated_count(ctx: &mut HttpFilterContext<'_>, key: &str, value: u6
 }
 
 /// Write accumulated streaming counts to the well-known metadata keys.
+///
+/// If no counts were ever accumulated but an oversized event was
+/// discarded during the stream, the terminal usage event itself may
+/// have been the one dropped — an explicit overflow status is recorded
+/// so this is distinguishable from a genuine zero-usage response.
 fn finalize_streaming_counts(ctx: &mut HttpFilterContext<'_>) {
```
aslakknutsen: "When Anthropic/Bedrock partial counts were accumulated and the final complete-usage SSE event exceeds `max_scratch_bytes`, `finalize_streaming_counts` emits partial totals without setting `token.status = overflow` even though `token_count.sse_dropped` was recorded. This may be acceptable if partial maxima are authoritative, but the residual case called out in the PR description (terminal usage event is the one that overflows) is only covered for the no-count path (`sse_overflow_with_no_usage_sets_overflow_status`), not for partial-then-drop."

filters/src/token_usage/count.rs
```diff
@@ -176,8 +215,16 @@ impl TokenCountFilter {
     /// Returns [`FilterError`] if the YAML config is invalid.
     pub fn from_config(config: &serde_yaml::Value) -> Result<Box<dyn HttpFilter>, FilterError> {
         let cfg: TokenCountConfig = parse_filter_config("token_count", config)?;
+        validate_max_body_bytes("token_count", cfg.max_body_bytes)?;
+        if cfg.max_scratch_bytes == 0 {
```
aslakknutsen: "`max_body_bytes` uses `validate_max_body_bytes`, which enforces >0 and ≤64 MiB. `max_scratch_bytes` only rejects zero, so an operator could configure a very large per-event scratch limit and allow substantial buffering before the overflow path triggers on a single SSE event. Defaults remain 64 KiB; misconfiguration risk is operator-side. Consider reusing the shared validator or a dedicated ceiling consistent with other filters."

filters/src/token_usage/count.rs
mkoushni: Fixed in e6077a1e. The scanner now tracks whether the *tail* of the stream was a drop vs a recovered payload. Partial Anthropic/Bedrock counts followed by an oversized terminal event emit those maxima *and* `token.status = overflow`. A usage event that arrives after a dropped content event stays authoritative (no overflow), matching the #674 recovery case. Covered by `sse_partial_then_oversized_terminal_event_sets_overflow_status`.

filters/src/token_usage/count.rs
mkoushni: Fixed in 7bf1e258. `max_scratch_bytes` now has a dedicated validator (same shape as `anthropic_stream_events` / `openai_stream_events`): reject 0, and cap at `MAX_JSON_BODY_BYTES` (64 MiB). Not reused `validate_max_body_bytes` because its error string hardcodes `'max_body_bytes'`.

aslakknutsen (approved):

PR #784, opened 2026-08-19
docs(proposals): How? for Google Vertex AI translation (#779)

mkoushni:
## Summary

Adds the **How?** section for [#779](https://github.com/praxis-proxy/ai/issues/779), one of the four per-provider sub-tasks split out of [#762](https://github.com/praxis-proxy/ai/issues/762). Builds on the accepted What?/Why? in `docs/proposals/00762_api-translation.md`, including its resolution that Vertex shares the SSE decoder family via a forced `?alt=sse`.

- Two new filters (`vertex_translate`, `vertex_stream_events`) mirroring the Bedrock design in [#783](https://github.com/praxis-proxy/ai/pull/783).
- `project`/`location` are operator-config-only and never read from the request at all (stronger than "strip"); `model` is independently allowlisted before path construction, same pattern as Bedrock.
- Reuses the existing `apis/src/openai/sse::SseFrameParser` + `HttpFilterContext::insert_filter_state` pattern (already used by `apis/src/openai/responses/stream_events/`) for the byte-level SSE reassembly, instead of a new parser.
- Reuses the existing `credential_inject` bearer-token seam for the upstream OAuth2 access token, with the shorter (~1h) GCP token lifetime explicitly flagged as a graduation criterion.
- Closes with an explicit acceptance-criteria → design mapping table.

## Test plan

- [ ] Maintainer review of design direction (this PR is design-only, no code)
- [ ] Stakeholder review of graduation criteria (fixture manifest, OAuth2 token rotation strategy)

PR #783, opened 2026-08-19
docs(proposals): How? for AWS Bedrock translation (#778)

mkoushni:
## Summary

Adds the **How?** section for [#778](https://github.com/praxis-proxy/ai/issues/778), one of the four per-provider sub-tasks split out of [#762](https://github.com/praxis-proxy/ai/issues/762). Builds on the accepted What?/Why? in `docs/proposals/00762_api-translation.md`.

- Two new filters (`bedrock_translate`, `bedrock_stream_events`) mirroring the existing `anthropic_to_openai`/`anthropic_stream_events` structure.
- Model-in-path handled via the framework's existing pre-read (`buffered_request_body`) mechanism — no second routing hop needed.
- Provider identity kept structural (fixed by filter-chain config); model independently allowlisted before path construction.
- Reuses `aws-smithy-eventstream` for binary event-stream framing, `apis/src/anthropic/to_openai/*` for the InvokeModel Anthropic-family body, and the existing `credential_inject` bearer-token seam for upstream auth (no SigV4 signing needed, per Bedrock API keys).
- Closes with an explicit acceptance-criteria → design mapping table.

## Test plan

- [ ] Maintainer review of design direction (this PR is design-only, no code)
- [ ] Stakeholder review of graduation criteria (fixture manifest, InvokeModel model-family scope, credential rotation strategy)

Issue #780, opened 2026-08-19
API translation: Azure OpenAI end-to-end inference


szedan-rh:
Sub-task of #762 (which is a sub-task of #114, under Epic #363).

## Goal

Enable real inference through Praxis against **Azure OpenAI**. A workload sends
an OpenAI-format request (e.g. *"What is the capital of France?"*) and Praxis
rewrites it to the Azure OpenAI deployment endpoint, forwards it, and normalizes
the response back to OpenAI format — the caller receives the answer without any
Azure-specific client code.

## Provider-specific notes

- **API:** Azure OpenAI deployment endpoint
  (`.../openai/deployments/{deployment}/chat/completions`) with a required
  `api-version` query parameter — deployment-name endpoint rewriting.
- **Streaming framing:** SSE.
- **Credentials:** `api-key` header (and optionally AAD bearer in
  `Authorization`). All consumer-supplied credential material must be stripped.

## Acceptance criteria

- [ ] Request transform OpenAI → Azure OpenAI deployment endpoint (path rewrite
      + `api-version`), fixture-backed.
- [ ] Response transform Azure OpenAI → OpenAI, fixture-backed.
- [ ] Error-response transform, fixture-backed.
- [ ] Streaming transform across arbitrary chunk boundaries using the SSE
      decoder; event framing and ordering preserved.
- [ ] Provider identity derived exclusively from the trusted route result;
      requests referencing an unrecognized/missing provider rejected before
      translation.
- [ ] Consumer-supplied credentials stripped (`api-key` header, `Authorization`
      bearer), verified by a fixture.
- [ ] Deterministic, documented header and body mutation ordering.
- [ ] Translation runs only after the caller/model authorization stage.
- [ ] Fixtures carry provenance (Azure OpenAI API version, schema version,
      source of truth) and are scanned for real credential material before merge.
- [ ] **End-to-end:** an OpenAI-format prompt returns a correct answer through
      Praxis against a real Azure OpenAI endpoint or a strict Azure protocol
      simulator (the "capital of France" test).
- [ ] Example config in `examples/configs/` plus a functional integration test
      for that config (per repo AGENTS.md test requirements).

Issue #762, opened 2026-08-18
API translation: fulfill translator acceptance criteria from praxis-extproc#6 (Bedrock, Vertex AI, Azure, Cohere)


mkoushni:
Sub-task of #114.

## Context

Originally scoped as the translator portion of opendatahub-io/praxis-extproc#6 ("Implement API translation and provider credential generation"). Migrated to this repo per the discussion on opendatahub-io/praxis-extproc#31, and re-scoped here to this repo's pre-existing #114 (Bedrock, Vertex AI, Azure, Cohere) since Anthropic <-> OpenAI translation is already covered by #103 / #638.

Design proposal: `docs/proposals/6_api-translation.md` (praxis-proxy/ai#761).

## Acceptance criteria

Adapted from opendatahub-io/praxis-extproc#6's required behavior/acceptance criteria, scoped to #114's provider set:

- [ ] Request and response transformations for AWS Bedrock (Converse + InvokeModel), Google Vertex AI (Gemini), Azure OpenAI, and Cohere, each backed by request/response/error/streaming fixtures.
- [ ] Correct handling of buffered responses and streaming responses across arbitrary chunk boundaries, using transport-specific decoders (SSE for Azure/Cohere, Bedrock's `application/vnd.amazon.eventstream` binary framing, Vertex AI's own event-stream framing) rather than a single SSE path.
- [ ] Deterministic, documented header and body mutation ordering.
- [ ] Transformed requests are accepted by representative provider endpoints or strict protocol simulators.
- [ ] Streaming output preserves event framing and ordering.
- [ ] Provider identity is derived exclusively from the trusted route result in stream state; requests referencing an unrecognized or missing provider are rejected before translation.
- [ ] Consumer-supplied provider credentials are stripped across every credential-bearing location (headers, SigV4 query params, URI path tokens, body fields) for each in-scope provider, verified by a fixture.
- [ ] Translation runs only after the required caller/model authorization stage has completed.
- [ ] Fixtures include provenance (provider API version, schema version, source of truth) and are scanned for real credential material before merge.

## Implementation tracking

One sub-task per in-scope provider, each carrying provider-specific acceptance
criteria and an end-to-end inference goal (send an OpenAI-format prompt, get the
answer back through Praxis):

- [ ] #778 — AWS Bedrock (Converse + InvokeModel)
- [ ] #779 — Google Vertex AI (Gemini)
- [ ] #780 — Azure OpenAI
- [ ] #781 — Cohere

[...]

PR #785, opened 2026-08-19
docs(proposals): How? for Azure OpenAI translation (#780)

mkoushni:
## Summary

Adds the **How?** section for [#780](https://github.com/praxis-proxy/ai/issues/780), one of the four per-provider sub-tasks split out of [#762](https://github.com/praxis-proxy/ai/issues/762). Builds on the accepted What?/Why? in `docs/proposals/00762_api-translation.md`.

- Because Azure OpenAI is deliberately wire-compatible with OpenAI's own Chat Completions schema, the design scopes body "translation" honestly small: strip the redundant top-level `model` field on request, identity pass-through on response/error (fixture-verified, not hand-waved), with a defensive fallback for non-conforming upstream errors.
- `model` resolves through an operator-owned `deployments` map (not a flat allowlist, since Azure deployment names are arbitrary) before path construction — same model-before-routing pre-read mechanism as the Bedrock ([#783](https://github.com/praxis-proxy/ai/pull/783)) and Vertex ([#784](https://github.com/praxis-proxy/ai/pull/784)) proposals.
- Identifies and closes a real gap: `credential_inject` only emits `Authorization: Bearer <token>` today, but Azure's primary auth is a raw `api-key` header. Proposes a small, optional, fully backward-compatible `header`/`value_prefix` addition to `credential_inject` (`filters` crate) rather than a second credential filter — flagged as its own graduation criterion since it's shared infra.
- `azure_stream_events` is deliberately read-only (no re-encoding) — it only tracks SSE frame boundaries to detect truncated streams (missing terminal event), since the chunk payloads themselves need no remapping.
- Closes with an explicit acceptance-criteria → design mapping table.

## Test plan

- [ ] Maintainer review of design direction (this PR is design-only, no code)
- [ ] Stakeholder review of graduation criteria, especially the `credential_inject` generalization

### DIFF
--- a/docs/proposals/00780_azure-translation.md
+++ b/docs/proposals/00780_azure-translation.md
@@ -1,0 +1,387 @@
+---
+issue: https://github.com/praxis-proxy/ai/issues/780
+discussion: >-
+  Sub-task of praxis-proxy/ai#762 (itself a sub-task of praxis-proxy/ai#114,
+  under Epic praxis-proxy/ai#363). The What?/Why? for provider translation in
+  general — including Azure OpenAI as one of the four in-scope providers —
+  was already discussed and accepted in
+  docs/proposals/00762_api-translation.md. This proposal narrows that
+  accepted scope to Azure OpenAI specifically and adds the How?, per
+  praxis-proxy/ai#762 being split into one per-provider sub-task
+  (praxis-proxy/ai#778-#781).
+status: proposed
+authors:
+  - mkoushni
+graduation_criteria:
+  - "credential_inject header-name/prefix generalization reviewed by its maintainers (shared filters/src code, not apis-crate-local)"
+  - "Azure request/response/error/stream fixture manifest reviewed by stakeholders"
+stakeholders:
+  - # TODO: add relevant maintainer/domain-expert handles before merge
+---
+
+# API Translation: Azure OpenAI
+
+## What?
+
+Add an Azure OpenAI-specific translation stage to the Praxis filter pipeline
+that rewrites OpenAI Chat Completions-shaped traffic to Azure's deployment
+endpoint (path + required `api-version` query parameter) and back. This is
+the Azure-scoped instance of the general translation stage already accepted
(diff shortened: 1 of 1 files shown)


szedan-rh:
LGTM

docs/proposals/00780_azure-translation.md
```diff
@@ -0,0 +1,387 @@
...
+  praxis-proxy/ai#762 being split into one per-provider sub-task
+  (praxis-proxy/ai#778-#781).
+status: proposed
+authors:
+  - mkoushni
+graduation_criteria:
+  - "credential_inject header-name/prefix generalization reviewed by its maintainers (shared filters/src code, not apis-crate-local)"
+  - "Azure request/response/error/stream fixture manifest reviewed by stakeholders"
+stakeholders:
+  - # TODO: add relevant maintainer/domain-expert handles before merge
```
alexsnaps: This again

alexsnaps (changes requested):
Nothing I can say here. 
Seems like a solid start to implement this.

docs/proposals/00780_azure-translation.md
szedan-rh: Same as before,
Shane,
Alex,
Aslak,
Senan

franciscojavierarceo:
the example chains use `credential_inject` with nested `secret_ref`/`strategy` configuration, but the repository currently provides `credential_injection` with a cluster-scoped configuration (`header`, `value`, `header_prefix`, and `strip_client_credential`). the proposed metadata contract is not implemented by the existing filter, so these examples cannot inject credentials as written. we should either align the design with the existing filter contract or specify the shared-infrastructure changes required before claiming the functional integration test is covered.

franciscojavierarceo:
the Azure design identifies the `api-key` requirement, but the proposed `header`/`value_prefix` extension is attached to the nonexistent `credential_inject` filter and does not explain how credentials are selected for the routed Azure cluster. we should define this against the actual `credential_injection` configuration and show how the Azure `api-key` reaches the upstream request.

Issue #781, opened 2026-08-19
API translation: Cohere end-to-end inference


szedan-rh:
Sub-task of #762 (which is a sub-task of #114, under Epic #363).

## Goal

Enable real inference through Praxis against **Cohere**. A workload sends an
OpenAI-format request (e.g. *"What is the capital of France?"*) and Praxis
translates it to Cohere's native Chat format, forwards it, and normalizes the
response back to OpenAI format — the caller receives the answer without any
Cohere-specific client code.

## Provider-specific notes

- **API:** Cohere Chat (`/v2/chat`, `stream: true` for streaming).
- **Streaming framing:** SSE.
- **Credentials:** Bearer token in the `Authorization` header. All
  consumer-supplied credential material must be stripped.

## Acceptance criteria

- [ ] Request transform OpenAI → Cohere Chat, fixture-backed.
- [ ] Response transform Cohere Chat → OpenAI, fixture-backed.
- [ ] Error-response transform, fixture-backed.
- [ ] Streaming transform across arbitrary chunk boundaries using the SSE
      decoder; event framing and ordering preserved.
- [ ] Provider identity derived exclusively from the trusted route result;
      requests referencing an unrecognized/missing provider rejected before
      translation.
- [ ] Consumer-supplied credentials stripped (`Authorization` bearer), verified
      by a fixture.
- [ ] Deterministic, documented header and body mutation ordering.
- [ ] Translation runs only after the caller/model authorization stage.
- [ ] Fixtures carry provenance (Cohere API version, schema version, source of
      truth) and are scanned for real credential material before merge.
- [ ] **End-to-end:** an OpenAI-format prompt returns a correct answer through
      Praxis against a real Cohere endpoint or a strict Cohere protocol
      simulator (the "capital of France" test).
- [ ] Example config in `examples/configs/` plus a functional integration test
      for that config (per repo AGENTS.md test requirements).

PR #786, opened 2026-08-19
docs(proposals): How? for Cohere translation (#781)

mkoushni:
## Summary

Adds the **How?** section for [#781](https://github.com/praxis-proxy/ai/issues/781), the last of the four per-provider sub-tasks split out of [#762](https://github.com/praxis-proxy/ai/issues/762). Builds on the accepted What?/Why? in `docs/proposals/00762_api-translation.md`.

- Cohere is structurally the simplest of the four: fixed `/v2/chat` endpoint, no model-in-path, so no early body pre-read or path-injection allowlist is needed the way Bedrock ([#783](https://github.com/praxis-proxy/ai/pull/783))/Vertex ([#784](https://github.com/praxis-proxy/ai/pull/784))/Azure ([#785](https://github.com/praxis-proxy/ai/pull/785)) need — the design is sized to match, using the existing `path_rewrite` builtin filter for the static path swap instead of new routing code.
- Real field-level request/response mapping is still required (Cohere is not wire-compatible with OpenAI, unlike Azure): `top_p`/`top_k` → `p`/`k`, nested usage counters, `finish_reason` mapping — while noting Cohere v2's `tool_calls` shape and streaming lifecycle were deliberately designed close to OpenAI/Anthropic conventions already in this codebase.
- Reuses the same `SseFrameParser` + `insert_filter_state` streaming pattern as the Vertex proposal, and the same flat-error-body mapping approach as the Bedrock proposal.
- `credential_inject`'s existing bearer-token strategy works unchanged — no shared-infra changes needed here (contrast with Azure's `api-key` gap).
- Closes with an explicit acceptance-criteria → design mapping table.

## Test plan

- [ ] Maintainer review of design direction (this PR is design-only, no code)
- [ ] Stakeholder review of graduation criteria (fixture manifest)

### DIFF
--- a/docs/proposals/00781_cohere-translation.md
+++ b/docs/proposals/00781_cohere-translation.md
@@ -1,0 +1,365 @@
+---
+issue: https://github.com/praxis-proxy/ai/issues/781
+discussion: >-
+  Sub-task of praxis-proxy/ai#762 (itself a sub-task of praxis-proxy/ai#114,
+  under Epic praxis-proxy/ai#363). The What?/Why? for provider translation in
+  general — including Cohere as one of the four in-scope providers — was
+  already discussed and accepted in
+  docs/proposals/00762_api-translation.md. This proposal narrows that
+  accepted scope to Cohere specifically and adds the How?, per
+  praxis-proxy/ai#762 being split into one per-provider sub-task
+  (praxis-proxy/ai#778-#781).
+status: proposed
+authors:
+  - mkoushni
+graduation_criteria:
+  - "Cohere v2 Chat request/response/error/stream fixture manifest reviewed by stakeholders"
+stakeholders:
+  - # TODO: add relevant maintainer/domain-expert handles before merge
+---
+
+# API Translation: Cohere
+
+## What?
+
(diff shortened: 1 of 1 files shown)


szedan-rh:
LGTM

franciscojavierarceo:
the example chains use `credential_inject` with nested `secret_ref`/`strategy` configuration, but the repository currently provides `credential_injection` with a cluster-scoped configuration (`header`, `value`, `header_prefix`, and `strip_client_credential`). the proposed metadata contract is not implemented by the existing filter, so these examples cannot inject credentials as written. we should either align the design with the existing filter contract or specify the shared-infrastructure changes required before claiming the functional integration test is covered.

franciscojavierarceo:
the response mapping only covers `COMPLETE`, `MAX_TOKENS`, `TOOL_CALL`, and `ERROR`, but Cohere also returns `STOP_SEQUENCE` and `TIMEOUT`. without an explicit mapping or fail-closed fallback, valid completions can lose their termination semantics. we should document the handling for every supported Cohere `finish_reason`.

PR #787, opened 2026-08-19
fix(compat): align AI with Praxis main body and branch routing

nerdalert:
## Summary

Align AI with the current Praxis API and terminal-branch semantics.

This PR contains three compatibility corrections:

1. Routing examples and integration tests now use a shared load-balancer filter after provider-selection branches. Praxis requires `rejoin: terminal` branches to produce the final client response; routing-only branches now rejoin the shared load balancer instead.
2. The AI API and filter test context constructors initialize the retry-related fields required by current Praxis main, behind the existing `praxis-main` feature.
3. The server retains the `TracingGuard` returned by current Praxis main's tracing initialization for the lifetime of the process. This resolves the new `unused_must_use` CI error while preserving the tracing provider lifecycle.

The second and third changes are proper compatibility fixes, not workarounds. Current Praxis main requires the additional context fields and a live tracing guard, while released Praxis builds do not. Feature-gating the context fields keeps both dependency versions supported, and retaining the guard is required for tracing to remain active. Neither change alters request routing or quota behavior.

The third issue surfaced only after the earlier compatibility compiler failure was fixed: CI progressed far enough to compile the server binary and then reported that the tracing guard was being dropped immediately. The fix is included in the same compatibility PR because it is part of the same current-Praxis-main build contract.

The three documentation-only changes update stale links left behind when proposals moved to the central `praxis-proxy/enhancements` repository. They restore the repository’s markdown-link lint without changing documentation content or runtime behavior.

## Files changed

- `apis/src/lib.rs`
- `filters/src/lib.rs`
- `server/src/main.rs`
- `README.md`
- `docs/README.md`
- `docs/anthropic-messages-replay-test-plan.md`
- `examples/configs/openai/responses/responses-routing.yaml`
- `examples/configs/openai/responses/tool-routing.yaml`
- `tests/integration/tests/suite/openai_responses_format.rs`
- `tests/integration/tests/suite/openai_tool_parse.rs`
- `tests/integration/tests/suite/responses_routing.rs`

No production routing algorithm or provider-selection implementation is changed. The server change only preserves the tracing guard that Praxis main requires.

## Validation

The branch is rebased onto current AI main and uses released Praxis `v0.5.3` by default.

Passed:

- `cargo +nightly fmt --all -- --check`
- `git diff --check`
- `make lint`
- `cargo test -p praxis-ai-filters` — 1,072 passed; 2 doctests passed, 5 ignored
- `cargo test -p praxis-ai-apis` — 2,530 passed, 27 ignored; 2 doctests passed
- `cargo test -p praxis-ai-proxy` — 59 library tests, 29 binary tests, 1 external-filter integration test, and 2 doctests passed
- Praxis-main API, filter, proxy, and schema compatibility tests previously passed against current Praxis main
- Affected routing tests previously passed

The earlier local Praxis-main integration build reached linking but rust-lld terminated with a system `Bus error`; no Rust compilation or test assertion failure was reported.

## Compatibility details

The following current-Praxis-main fields are initialized only under `#[cfg(feature = "praxis-main")]`:

- attempted endpoints;
- retry policy;
- route retry policy;
- shared cluster retry state;
- retry-state release marker;
- endpoint reselector.

This preserves compatibility with released Praxis while allowing the Praxis-main CI workflow to compile against the newer context shape.

The tracing initialization result is bound to `_tracing_guard` rather than ignored. The underscore-prefixed binding suppresses the unused-variable warning while keeping the guard alive until `main` exits. When the Praxis release containing this API is consumed as the normal dependency, this line remains correct and does not need to be reverted.

## Breaking changes

None intended.

### DIFF
--- a/README.md
+++ b/README.md
@@ -128,7 +128,7 @@ filters and protocol support. Before opening a pull request, please read the
 [development setup](docs/developing/getting-started.md).
 
 For larger changes, open a [feature request] and follow the
-[proposal process](docs/proposals.md) so we can shape the idea together.
+[proposal process](https://github.com/praxis-proxy/enhancements) so we can shape the idea together.
 
 [Open an issue][issues] · [Request a feature][feature request] ·
 [Open a pull request][pull requests]
--- a/apis/src/lib.rs
+++ b/apis/src/lib.rs
@@ -97,6 +97,18 @@ pub(crate) mod test_utils {
             response_headers_modified: false,
             subrequest_client: None,
             subrequest_response_mode: praxis_filter::SubRequestResponseMode::Buffered,
+            #[cfg(feature = "praxis-main")]
+            attempted_endpoints: Vec::new(),
+            #[cfg(feature = "praxis-main")]
+            retry_policy: None,
+            #[cfg(feature = "praxis-main")]
+            route_retry_policy: None,
+            #[cfg(feature = "praxis-main")]
+            cluster_retry_state: None,
+            #[cfg(feature = "praxis-main")]
+            cluster_retry_state_released: false,
+            #[cfg(feature = "praxis-main")]
+            endpoint_reselector: None,
             rewritten_path: None,
             selected_endpoint_index: None,
             time_source: &praxis_core::time::SystemTimeSource,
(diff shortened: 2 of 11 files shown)


nerdalert:
Looks like breakage in will open if I dig it out.

shaneutt (approved):

Issue #746, opened 2026-08-14
refactor(openai): generalize the operation registry and matcher


leseb:
## Description

Generalize the existing OpenAI operation metadata and Conversations route matcher into a shared, family-agnostic registry foundation.

The shared operation specification must describe operation identity and runtime request shape independently from whether Praxis owns the external OpenAPI contract. Family modules should be able to declare operations once and reuse those declarations for runtime matching and conformance checks.

Parent epic: #740

## Scope

- Add an API-family identifier to shared operation metadata.
- Represent HTTP and WebSocket transport qualifiers.
- Represent request bodies independently from `OwnedOperationContract`:
  - no body
  - required or optional JSON
  - required or optional multipart
  - required or optional binary
- Move path-template matching into the shared operation module.
- Match methods exactly and path parameters as one non-empty segment.
- Define deterministic static-before-parameter precedence.
- Apply one documented query-string and trailing-slash policy.
- Preserve zero-allocation matching on the normal request path where practical.
- Keep implementation/conformance metadata separate from the facts used to identify an operation.

## Acceptance Criteria

- Proxied operations can declare request-body shape without claiming contract ownership.
- Conversations operation declarations continue to generate and verify their owned OpenAPI contract.
- The shared matcher handles static paths, parameterized paths, wrong methods, query strings, trailing slashes, and empty or extra segments.
- Duplicate family/operation IDs and method/path keys fail tests.
- Existing Conversations runtime and conformance tests remain green.
- No request or response body is cloned or buffered by the matcher.

## Tests and Documentation

- Boundary tests for every matching rule and ambiguity case
- Registry uniqueness tests
- Updated operation-module documentation explaining identity, request shape, and contract ownership

## Dependencies

None.

## Area

Provider APIs (OpenAI/Anthropic)

cdoern:
WIP: https://github.com/cdoern/ai/tree/feat/openai-operation-body-kind

Issue #744, opened 2026-08-14
feat(ai): add a request-head operation classifier


leseb:
## Description

Add an `ai_operation` filter that classifies supported application operations
from the HTTP request head and publishes a provider-neutral typed match for
downstream filters.

The classifier aggregates protocol-owned registries through the shared matcher.
It does not inspect request bodies or infer the provider that will ultimately
serve the request.

Parent epic: #740

## Scope

- Run classification during `on_request` with no request-body access.
- Match using the provider-neutral operation registry foundation.
- Aggregate registered OpenAI, Anthropic, and future application protocols
  without protocol-specific branches in the classifier.
- Insert a typed `AiOperationMatch` into request extensions.
- Publish stable metadata and filter results:
  - `ai_operation.application_protocol`
  - `ai_operation.operation_id`
- Support configurable proxy-owned routing headers:
  - `x-praxis-ai-application-protocol`
  - `x-praxis-ai-operation`
- Remove or overwrite client-supplied values for configured routing headers.
- Leave unmatched requests unchanged and let routing policy determine fallback
  behavior.
- Register the filter and provide a focused example configuration.
- Document that current `on_request` publication follows `StreamBuffer`
  pre-read; early visibility to body processors is tracked by
  praxis-proxy/praxis#1142.

## Acceptance Criteria

- The filter declares no request-body access or buffering mode.
- Supported method/path/transport combinations publish the expected typed match,
  metadata, results, and configured headers.
- Unsupported methods and paths publish no operation match.
- Adding a protocol registry requires no protocol-specific branch in
  `ai_operation`.
- OpenAI Responses HTTP and WebSocket operations remain distinguishable.
- Client-supplied operation headers cannot influence the authoritative match.
- No `application_provider` is inferred or published from the incoming request.
- Unit tests, functional integration tests, example configuration, generated
  example inventory, and filter documentation are complete.

## Dependencies

- #1050 — provider-neutral operation identity and matcher

## Related Work

- praxis-proxy/praxis#1142 — make this match available before `StreamBuffer`
  pre-read
- praxis-proxy/praxis#1075 — router visibility of pending header mutations

## Area

Provider APIs (OpenAI/Anthropic)

PR #788, opened 2026-08-19
feat(responses): register operations for request-head classification

### LINKED ISSUE #743: feat(responses): register operations for request-head classification
## Description

Declare the supported OpenAI Responses operations in a family-owned registry so their identities come from method, path, and protocol headers rather than body heuristics.

Parent epic: #740

## Scope

- Register the supported Responses operations, including:
  - create response
  - retrieve response
  - delete response
  - cancel response
  - list input items
  - count input tokens
  - compact a conversation
  - Responses WebSocket creation where supported
- Declare request-body shape for every operation.
- Preserve official operation IDs from the pinned specification where applicable.
- Model WebSocket support as an explicit runtime/protocol extension when it is not represented by the pinned HTTP operation.
- Add drift checks for method, path, and operation ID.
- Publish the registry through the central OpenAI operation matcher.

## Acceptance Criteria

- Every supported Responses method/path pair resolves to the correct operation.
- `POST /v1/responses` and `GET /v1/responses` WebSocket are distinguishable without inspecting a body.
- Unsupported methods such as `PUT /v1/responses` do not match a Responses operation.
- Static endpoints such as input-token counting and compaction cannot be consumed as response-ID parameters.
- Bodyless and body-bearing operations report the correct request shape.
- Unit and conformance tests cover all registered operations.

## Dependencies

- #746 — shared operation registry and matcher

## Area

Provider APIs (OpenAI/Anthropic)

cdoern:
Closes #743

## Stacked on #772

The first commit on this branch, `edee244`, is the #746 work under review in #772.
It is included because this registry is the first consumer of it. **Merge #772
first**; this branch will then contain only its own commit. Review just
`e09e42d` here.

## Summary

Responses operation identity came from path string matching in the request
classifier — hand-written prefix and suffix checks per endpoint. This declares
the supported operations in a family-owned registry instead, so identity comes
from method, path, and protocol headers before any payload is read.

Eight operations are registered against the shared registry:

| Operation ID | Method | Path | Body |
| --- | --- | --- | --- |
| `createResponse` | POST | `/responses` | JSON, required |
| `praxis_createResponseWebSocket` | GET (WebSocket) | `/responses` | none |
| `getResponse` | GET | `/responses/{response_id}` | none |
| `deleteResponse` | DELETE | `/responses/{response_id}` | none |
| `cancelResponse` | POST | `/responses/{response_id}/cancel` | none |
| `listInputItems` | GET | `/responses/{response_id}/input_items` | none |
| `Getinputtokencounts` | POST | `/responses/input_tokens` | JSON, required |
| `Compactconversation` | POST | `/responses/compact` | JSON, required |

Operation IDs are the official ones from the pinned specification, reproduced
verbatim — including upstream's casing for the last two.

## Contract ownership

Praxis proxies the Responses contract rather than owning it, so every operation
declares its runtime request-body shape without an owned OpenAPI contract. This
is the first registry to depend on body shape being independent of contract
ownership, which is the change #746 makes.

## WebSocket as a protocol extension

The pinned specification does not represent the Responses WebSocket handshake as
an HTTP operation, so it is declared an explicit Praxis protocol extension and
separated from `POST /v1/responses` by transport rather than by inspecting a
body. Drift checks skip declared extensions, and fail if one later turns up in
the specification.

## Drift check

`cargo xtask check-responses-registry`, wired into `make lint`, compares
registered method, path, and operation ID against the pinned specification.

It projects the specification through the existing semantic YAML reader rather
than parsing it directly, because the pinned document contains an integer
(`components.schemas.CreateChatCompletionRequest.allOf[1].properties.seed.minimum`,
line 32725) that `serde_yaml` rejects.

Verified to fail correctly: renaming `cancelResponse` in the registry produces

```
POST /responses/{response_id}/cancel registers operation ID cancelResponseTypo
but the pinned specification says cancelResponse
```

and exit code 1.

## File placement

The registry lives in `responses/routes/mod.rs` rather than `responses/routes.rs`
because the filter doc generator parses non-anchor files in a category root as
shared types. As a plain file it leaked registry type documentation into four
unrelated filter docs, replacing their descriptions with
"Static metadata for one Responses operation".

## Testing

- `cargo test --workspace` — all suites pass, including 2556 in `praxis-ai-apis`
- `make lint` — passes end to end, including the new drift check
- `make doc` — clean

New coverage: every registered operation resolving from its own template, static
endpoints not consumed as response IDs, identifier paths still capturing the
response ID, create and WebSocket separated without reading a body, unsupported
methods and unknown subresources not matching, body-bearing versus bodyless
shapes, registry key and operation ID uniqueness, and the WebSocket operation
being the only declared protocol extension.

## Not in scope

Switching the classifier and Responses filters over to consume this registry
depends on the `openai_operation` classifier in #744.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -140,6 +140,7 @@ lint:
 	cargo xtask sync-inference-readme
 	cargo xtask sync-responses-readme
 	cargo xtask check-inference
+	cargo xtask check-responses-registry
 
 fmt:
 	cargo +nightly fmt --all
--- a/apis/src/openai/mod.rs
+++ b/apis/src/openai/mod.rs
@@ -30,6 +30,12 @@ pub use operation::{OpenAiHandlingMode, OpenAiOperationSpec, OpenAiRequestBody};
 pub use responses::{
     AgenticLoopFilter, CompactFilter, DocExtractFilter, FileResolveFilter, FileSearchCalloutFilter, McpDispatchFilter,
     McpToolResolveFilter, ModelRewriteFilter, OpenaiResponsesValidateFilter, RehydrateFilter, ResponseStoreFilter,
-    ResponsesFormatFilter, ToolParseFilter, WebSearchFilter, openai_responses_proxy::ResponsesProxyFilter,
-    responses_to_chat_completions::ResponsesToChatCompletionsFilter, stream_events::OpenaiStreamEventsFilter,
+    ResponsesFormatFilter, ToolParseFilter, WebSearchFilter,
+    openai_responses_proxy::ResponsesProxyFilter,
+    responses_to_chat_completions::ResponsesToChatCompletionsFilter,
+    routes::{
+        PROTOCOL_EXTENSION_OPERATION_IDS as RESPONSES_PROTOCOL_EXTENSION_OPERATION_IDS, ResponsesOperation,
+        ResponsesOperationSpec, operation_specs as responses_operation_specs,
+    },
+    stream_events::OpenaiStreamEventsFilter,
 };
--- a/apis/src/openai/responses/mod.rs
+++ b/apis/src/openai/responses/mod.rs
@@ -37,6 +37,12 @@ pub(crate) mod openai_mcp_tool_resolve;
 pub(crate) mod openai_responses_proxy;
 pub(crate) mod openai_tool_parse;
 pub(crate) mod responses_to_chat_completions;
+#[expect(clippy::allow_attributes, reason = "dead_code expect unfulfilled on module")]
+#[allow(
+    dead_code,
+    reason = "the Responses operation registry is consumed by the openai_operation classifier"
+)]
+pub(crate) mod routes;
 pub(crate) mod state;
 pub(crate) mod store;
 pub(crate) mod stream_events;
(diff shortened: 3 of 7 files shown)


leseb (changes requested):
2 findings:

[MAJOR] Request-body optionality disagrees with the pinned specification
routes/mod.rs:162
What: CountInputTokens and CompactConversation declare [required json], while their pinned requestBody definitions omit required, which defaults to false under OpenAPI. The canonical OpenResponses compact definition does likewise. [OpenAPI 3.1](https://spec.openapis.org/oas/v3.1.0), [OpenResponses specification](https://raw.githubusercontent.com/openresponses/openresponses/main/public/openapi/openapi.json)
Impact: The registry publishes incorrect contract metadata that downstream consumers may use.
Fix: Use [optional json], or document and test an explicit provider-contract exception.


[MAJOR] Extension drift check misses upstream operation IDs
responses_registry.rs:37
What: An upstream GET /responses is rejected only if its operation ID equals praxis_createResponseWebSocket. A normal upstream-owned ID therefore returns Agrees.
Impact: Lint misses exactly the documented case where a Praxis extension becomes official.
Fix: Treat any found.is_some() as drift for extensions and add positive/negative unit tests.
This second one frankly sounds very unlikely or could be fixed whether OpenAI actually opens up a new endpoint.

leseb (approved):

Issue #673, opened 2026-08-10
NeMo modified verdicts forward the original sensitive prompt while claiming it was redacted


leseb:
## Clawpatch finding

- **Severity:** high
- **Category:** security
- **Confidence:** high
- **Triage:** confirmed-bug
- **Feature:** `feat_custom_ai_anthropic_guardrails_prompt_token`
- **Finding ID:** `fnd_sig-feat-custom-ai-anthropic-gua_cfe2e666bb`
- **Run ID:** `20260810T073440-229564`
- **Prior-run overlap:** None detected in preserved prior .clawpatch snapshot.

## Evidence

- `filters/src/guardrails/filter.rs:166-169` (`record_verdict`) — GuardResult::Redact { reason, .. } => { ... Ok(FilterAction::Continue) }
- `filters/src/guardrails/tests.rs:343-372` (`on_request_body_modified_writes_filter_results`) — The provider returns redacted content, but the test expects Continue and records status as redacted.

## Description

A NeMo `modified` response contains replacement text intended to remove sensitive content, but `record_verdict` discards `modified_text`, forwards the unchanged request, and publishes the status `redacted`. The included test demonstrates this with an SSN: NeMo returns `my ssn is [REDACTED]`, yet the original `123-45-6789` request is forwarded. This both leaks content the configured guardrail identified as sensitive and can mislead downstream branch logic or auditing into treating the request as sanitized.

## Reproduction

Configure `ai_guardrails` with a NeMo endpoint. Send `{"messages":[{"role":"user","content":"my ssn is 123-45-6789"}]}` and have NeMo return `{"status":"modified","content":"my ssn is [REDACTED]"}`. The filter returns `Continue`, leaves the request body unchanged, and sets `ai_guardrails.status=redacted`.

## Recommendation

Until body reconstruction is implemented, fail closed for `GuardResult::Redact` by rejecting the request. Alternatively, replace the relevant message content with `modified_text` before returning `Continue`, and only publish `redacted` after successful replacement.

## Suggested regression test

Return a NeMo `modified` verdict for a body containing a recognizable secret and assert either that the forwarded body contains only the provider's modified text or that the request is rejected; also assert that `redacted` is never recorded while the original secret remains in the body.

---
[...]

PR #792, opened 2026-08-20
fix(guardrails): align NeMo status with /v1/guardrail/checks endpoint

### LINKED ISSUE #791: Guardrails: NeMo status constants don't match /v1/guardrail/checks endpoint
### Description

The NeMo provider in the `ai_guardrails` filter maps response
statuses that don't match the `/v1/guardrail/checks` endpoint
we're actually calling.

**Actual `/v1/guardrail/checks` response statuses:**

| Status | Meaning |
|---|---|
| `"success"` | all rails passed |
| `"blocked"` | at least one rail blocked |
| `"error"` | NeMo internal processing error |

Two problems:

1. **`"success"` hits the `other` (unknown) branch** - the code
   expects `"passed"` but the endpoint returns `"success"`. This
   triggers the fail-closed error path with a misleading log:
   `unknown status 'success'`.

2. **`"modified"` and `content` field are dead code** - the
   `/v1/guardrail/checks` endpoint does not support redaction.
   It never returns `"modified"` or a `content` field. These
   exist because the TrustyAI team originally planned to support
   redaction on this endpoint but ultimately decided not to
   include it.

Same issue in: opendatahub-io/ai-gateway-payload-processing#432
(same fix for the Go side).

### Steps to Reproduce

1. Deploy Praxis with the `ai_guardrails` filter configured
   to use a NeMo provider pointing at a TrustyAI
   NeMo-Guardrails instance serving `/v1/guardrail/checks`
2. Send a request with a non-toxic user message that passes
   all rails
3. NeMo returns `{"status": "success", ...}`
4. Observe: Praxis returns an error instead of allowing the
   request through, because `"success"` falls into the
   `other` arm of the match in `map_nemo_response`

### Expected Behavior

- `"success"` responses should map to `GuardResult::Pass`
  (request proceeds normally)
- `"error"` responses should return a `FilterError`
  (fail-closed with a clear message, not "unknown status")
- `"blocked"` stays as-is (already correct)
- `"modified"` match arm and `content` field should be removed
  (with a TODO for future `/v1/checks` migration)
- `blocked_rail_names` filter already checks for `"blocked"`
  which is correct for this endpoint

### Praxis AI Version

v0.1.0

### Area

AI / Inference

liavweiss:
## Summary

- Fix NeMo provider status mapping to match the actual `/v1/guardrail/checks`
  endpoint responses. The code expected `"passed"` but the endpoint returns
  `"success"`, causing every non-toxic request to hit the unknown-status
  error path and get rejected.
- Remove dead `"modified"` match arm and `content` field - the
  `/v1/guardrail/checks` endpoint does not support redaction. TrustyAI
  originally planned to add it but decided not to.
- Add `"error"` status handling with `guardrails_data` extraction so NeMo
  processing errors produce actionable log messages instead of
  `unknown status 'error'`.
- Add unit tests covering `success`, `blocked`, and `error` status paths.

## Related issue

Companion to [opendatahub-io/ai-gateway-payload-processing#432](https://github.com/opendatahub-io/ai-gateway-payload-processing/issues/432)
(same fix for the Go side).

## Fix
#791 

## Validation

- [x] Unit tests
- [X] Integration or functional tests
- [X] `make lint`

Bug confirmed by sending a non-toxic request to a live TrustyAI NeMo instance
and observing the `"success"` status in the response, which the old code
would reject as unknown.

## Checklist

- [ ] New capabilities include an example config and functional example test.
- [ ] User-facing behavior and generated documentation are updated.

## Breaking changes

None. This is a bug fix - the previous behavior was incorrect (all
non-toxic requests were rejected).

### DIFF
--- a/tests/integration/tests/suite/examples/guardrails.rs
+++ b/tests/integration/tests/suite/examples/guardrails.rs
@@ -27,7 +27,7 @@ fn nemo_guardrails_config_parses_correctly() {
 #[test]
 fn nemo_guardrails_forwards_to_backend() {
     let backend = start_backend_with_shutdown("ok");
-    let nemo = nemo_mock(r#"{"status":"passed","rails_status":{"self check input":{"status":"success"}}}"#);
+    let nemo = nemo_mock(r#"{"status":"success","rails_status":{"self check input":{"status":"success"}}}"#);
     let proxy_port = free_port();
     let config = load_example_config(
         "nemo-guardrails.yaml",
@@ -42,7 +42,7 @@ fn nemo_guardrails_forwards_to_backend() {
         r#"{"model":"test","messages":[{"role":"user","content":"Hello, how are you?"}]}"#,
     );
 
-    assert_eq!(status, 200, "NeMo 'passed' should forward to upstream");
+    assert_eq!(status, 200, "NeMo 'success' should forward to upstream");
     assert_eq!(body, "ok", "upstream response should reach the client");
 }
 
@@ -73,13 +73,13 @@ fn nemo_guardrails_block_rejects_with_403() {
     );
 }
 
-/// `NeMo` returns `"modified"` (redact placeholder) → request is forwarded
-/// to the upstream unchanged and the upstream response is returned.
+/// `NeMo` returns `"error"` with `guardrails_data` → proxy fails closed
+/// with a 500 and does not forward to the upstream.
 #[test]
-fn nemo_guardrails_redact_placeholder_continues() {
+fn nemo_guardrails_error_status_does_not_forward() {
     let backend = start_backend_with_shutdown("ok");
     let nemo = nemo_mock(
-        r#"{"status":"modified","content":"my ssn is [REDACTED]","rails_status":{"pii masking":{"status":"blocked"}}}"#,
+        r#"{"status":"error","rails_status":{},"guardrails_data":{"error":"Config load failed.","details":"bad path"}}"#,
     );
     let proxy_port = free_port();
     let config = load_example_config(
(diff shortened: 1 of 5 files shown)


liavweiss:
This PR removes the `"modified"` match arm, the `content` field, and the
`GuardResult::Redact` usage from the NeMo provider. These existed because
redaction support was originally planned for `/v1/guardrail/checks`, but
TrustyAI decided not to include it on this endpoint.

Redaction (`"modified"` status + transformed `content`) is only available
on the upstream `/v1/checks` endpoint. However, that endpoint currently
evaluates only the last user message, which means multi-message
conversations would require splitting into N sequential HTTP calls -
adding ~37% latency overhead per additional user message (benchmarked
with pod-to-pod RTT).

We opened a feature request with Nvidia to support multi-message
evaluation on `/v1/checks`:
- [NVIDIA-NeMo/Guardrails#2277](https://github.com/NVIDIA-NeMo/Guardrails/issues/2277)

aslakknutsen (approved):

aslakknutsen:
@liavweiss Could you please clear up the conflicts here and we're good to go. .:)

liavweiss:
> @liavweiss Could you please clear up the conflicts here and we're good to go. .:)

Hey @aslakknutsen, thank you for the review.
I added this [comment](https://github.com/praxis-proxy/ai/issues/673#issuecomment-5399213299), it will be more reasonable to revert #721 and then merge mine, I can create a revert pr if needed.

christinaexyou:
@liavweiss do you mind waiting for my review ?

tests/integration/tests/suite/examples/guardrails.rs
christinaexyou: non-blocking but we should update this stale assertion

christinaexyou:
thanks @liavweiss ! just a couple of nits from me; they are non-blocking so i'll approve but we should address them either in this PR or a follow-up: 
* stale assertion message; uses passed instead of success: https://github.com/praxis-proxy/ai/blob/b90724df41c82f10af7388d5128a57d1018feb45/tests/integration/tests/suite/examples/guardrails.rs#L45

* stale comment; same reason as above: https://github.com/praxis-proxy/ai/blob/b90724df41c82f10af7388d5128a57d1018feb45/examples/configs/nemo-guardrails.yaml#L6

christinaexyou (approved):

liavweiss:
> thanks @liavweiss ! just a couple of nits from me; they are non-blocking so i'll approve but we should address them either in this PR or a follow-up:
> 
> * stale assertion message; uses passed instead of success: https://github.com/praxis-proxy/ai/blob/b90724df41c82f10af7388d5128a57d1018feb45/tests/integration/tests/suite/examples/guardrails.rs#L45
> * stale comment; same reason as above: https://github.com/praxis-proxy/ai/blob/b90724df41c82f10af7388d5128a57d1018feb45/examples/configs/nemo-guardrails.yaml#L6

Done.

liavweiss:
Hey @aslakknutsen and @christinaexyou
The code is ready to be merged. Thank you for the reviews.

Issue #794, opened 2026-08-20
test-praxis-main red repo-wide: praxis@main drift (policy-engine rename + parse_json_rpc_body signature change)


jordigilh:
## Summary

`test-praxis-main` (`.github/workflows/praxis-compat.yaml`) is currently red on every open PR I checked (`#750`, `#760`, `#769`, `#772`, `#658`, `#709`), independent of each PR's own content. Root cause is upstream drift between `praxis-proxy/praxis@main` (unreleased) and the `praxis-proxy-filter` version this repo's `Cargo.toml` pins.

Two distinct incompatibilities landed on `praxis@main` this week, via [praxis#943](https://github.com/praxis-proxy/praxis/pull/943) ("move the policy filter onto the Praxis Policy Engine"):

### 1. Cargo feature renamed: `cpex-policy-engine` → `policy-engine`

`praxis`'s `filter/Cargo.toml` used to have:
```toml
cpex-policy-engine = ["dep:cpex"]
```
and now has:
```toml
policy-engine = ["dep:ppe"]
```

This repo's `tests/integration/Cargo.toml` still declares a local feature named after the old name:
```toml
[features]
cpex-policy-engine = []
```
which should track the rename.

### 2. `parse_json_rpc_body()` signature change

`praxis::filter::builtins::http::payload_processing::body_parsing::parse_json_rpc_body`'s `body` parameter changed from `&mut Option<Bytes>` to `Option<&Bytes>` on `praxis@main`. This repo's two call sites still pass the old shape directly:

- `filters/src/agentic/a2a/mod.rs:179`
- `filters/src/agentic/mcp/mod.rs:149`

```
error[E0308]: mismatched types
   --> filters/src/agentic/a2a/mod.rs:179:13
    |
179 |             body,
    |             ^^^^ expected `Option<&Bytes>`, found `&mut Option<Bytes>`
```

## Why this can't be fixed unilaterally in `ai` right now

`test-praxis-main` is the *only* CI job that overrides `praxis-proxy-filter` to a path dependency on freshly-cloned `praxis@main` (via `.github/actions/patch-praxis`); every other job (`test`, `lint`, `coverage`, `msrv`, …) builds against the registry-pinned `praxis-proxy-filter` version, which still has the *old* feature name and the *old* `parse_json_rpc_body` signature. Updating the two call sites to `.as_ref()` and renaming the feature would fix `test-praxis-main` but immediately break every other job, since they'd no longer match the pinned dependency.

[...]

PR #795, opened 2026-08-20
fix: track praxis@main policy-engine rename + parse_json_rpc_body signature (blocked on praxis release)

jordigilh:
## Summary

Draft fix for #794 — `test-praxis-main` is red across every open PR in this repo because `praxis@main` (unreleased) has drifted from the `praxis-proxy-filter` version pinned in `Cargo.toml`. This PR makes the `ai`-side changes needed to match `praxis@main`:

- `filters/src/agentic/a2a/mod.rs` / `filters/src/agentic/mcp/mod.rs`: pass `body.as_ref()` instead of `body` to `parse_json_rpc_body()`, matching the signature change from `&mut Option<Bytes>` to `Option<&Bytes>` on `praxis@main` (from [praxis#943](https://github.com/praxis-proxy/praxis/pull/943)).
- `tests/integration/Cargo.toml`: rename the orphaned local `cpex-policy-engine` feature to `policy-engine`, tracking the same rename on `praxis`'s `filter` crate.

## Why this is a draft, and why it's expected to fail CI right now

Every job except `test-praxis-main` builds against the **registry-pinned** `praxis-proxy-filter` version, which still has the *old* feature name and the *old* `parse_json_rpc_body` signature. So as of this commit:

- `test-praxis-main` should now pass (or at least get past this specific error).
- `test` / `lint` / `coverage` / `msrv` are expected to **fail**, because the call sites here no longer match the pinned registry version.

This PR is not mergeable until `praxis-proxy-filter` publishes a release containing both the `policy-engine` rename and the `parse_json_rpc_body` signature change. At that point, this PR also needs the `Cargo.toml` version requirement for `praxis-filter`/`praxis-core`/etc. bumped to the new release, in the same commit, so registry and `praxis@main` builds agree again.

## Test plan

- [ ] Blocked on a `praxis-proxy-filter` release containing praxis#943's changes
- [ ] Once released: bump the `praxis-filter` version requirement in `Cargo.toml`
- [ ] Confirm `test`, `lint`, `coverage`, `msrv`, and `test-praxis-main` all pass together


leseb:
@jordigilh should we re-open now?

PR #790, opened 2026-08-20
feat(token-rate-limit): add shared Valkey quota enforcement

nerdalert:
## Summary

**Dicalimer 🔴** This is not for merge, but a fairly cleaned up effort at a groundwork for quotas used for this demo:   **Distributed token quota demonstration: grid-distributed-token-rate-limit:
    https://github.com/praxis-proxy/experimental/tree/main/demos/grid-distributed-token-rate-limit**

  This PR adds authenticated, model-scoped token quota enforcement to Praxis AI through the new token_rate_limit filter.

  The filter admits or rejects a request before provider routing begins. It reserves an estimated token amount against one or more sliding-window budgets, then reconciles that reservation with the actual token usage reported after the response completes.

  The quota mechanism is independent of Grid routing:

```
  Authentication
        |
        | trusted identity.user_id
        v
  token_rate_limit
        |
        | admitted request
        v
  Routing
    - intelligent_route
    - static routing
    - another load balancer
        |
        v
  Inference backend
        |
        | actual token usage
        v
  token_rate_limit reconciliation
```

  It can therefore be used with Grid provider selection, static upstream routing, or another routing implementation. Grid does not store quotas, contact the quota backend, or participate in quota enforcement.

  ## Project Stack

```

  Praxis core
    Establishes trusted authenticated identity metadata
    identity.user_id
          |
          v
  Praxis AI
    Reserves and reconciles token quota
    Returns 429 before routing when quota is exhausted
          |
          v
  Grid + Praxis AI provider selection
    Selects an eligible provider only for admitted requests
          |
          v
  Provider gateway and inference stack
    Processes the admitted request
```

  The ownership boundaries are intentional:

  - Praxis core authenticates the caller and publishes trusted request-scoped identity metadata.
  - Praxis AI owns quota admission, reservation, reconciliation, and quota responses.
  - Grid determines provider eligibility and routing state.
  - intelligent_route selects a provider from the current in-memory routing snapshot.
  - The inference stack executes the admitted request and reports token usage.

  Quota enforcement remains outside Grid and does not add a quota lookup to Grid reconciliation or provider selection.

  ## Related Work

  - AI token rate-limiting proposal: docs/proposals/00121_token-rate-limiting.md
    https://github.com/praxis-proxy/ai/blob/main/docs/proposals/00121_token-rate-limiting.md

  - AI token-rate-limiting implementation: praxis-proxy/ai#658
    https://github.com/praxis-proxy/ai/pull/658

[...]

Issue #121, opened 2026-04-03
Epic: Token Rate Limiting


shaneutt:
## Overview

Rate limiting based on token consumption rather than request count. Requires token counts from the Token Counting filter (#20) in the filter chain.

## Algorithm

Token bucket (preferred for burstiness tolerance) with configurable refill rate and burst capacity. Independent buckets per token type: input, output, and total. Token bucket allows short bursts while enforcing sustained rate limits.

## Bucket Keys

Rate limit buckets keyed by: per-IP, per-header (API key, tenant ID), per-model, or composite keys combining multiple dimensions.

## Hierarchical Quotas

Cascading limits at org > team > user levels. Each level enforced independently. A user within their personal quota can still be blocked by an exhausted org-level bucket.

## Response Format

- 429 status code when token quota exceeded
- `Retry-After` header calculated from token consumption rate and bucket refill
- `X-RateLimit-Limit-Tokens`: configured token limit
- `X-RateLimit-Remaining-Tokens`: tokens remaining in current window
- `X-RateLimit-Reset`: time until bucket refills

## Distributed Counters

- In-memory counters for single-instance deployments
- Redis/Valkey backend for multi-instance consistency via KV store registry

## Filter Design

Separate `token_rate_limit` filter, distinct from the existing request-count `rate_limit` filter. Depends on token counting filter earlier in the chain.

shaneutt:
Note: we're going to expand the scope of this epic shortly based on new feedback, this will include more subtasks and our goals/requirements, stay tuned.

nerdalert:
Cross-component follow-on tracker after #796: praxis-proxy/grid#101. It separates the initial limiter foundation from the remaining identity, distributed enforcement, lifecycle, observability, and metering work while keeping quota enforcement outside Grid.

asaadbalum:
## Post-#796 Status Update

[#796](https://github.com/praxis-proxy/ai/pull/796) merged Aug 31 (experimental, feature-gated). Core M1/M2/M6 implementation is in main.

### Closed (covered by #796)
- #120 — Response-cost deduction
- #122 — Token bucket rate limiter
- #124 — Token-aware 429 responses
- #126 — Distributed token counters via KV store
- #127 — Kuadrant integration (superseded by #844)
- #789 — Sliding window rate limiting

### Updated (partially covered, remaining scope narrowed)
- #123 — Configurable bucket keys (M5 keying resolution pending)
[...]

PR #796, opened 2026-08-20
feat(token_rate_limit): sliding-window/token-bucket rate limiting (M1/M2/M6)

jordigilh:
## Summary

Implements the uncontested MVP core of the token rate limiting proposal (`00121_token-rate-limiting.md` in [praxis-proxy/enhancements](https://github.com/praxis-proxy/enhancements/blob/main/proposals/00121_token-rate-limiting.md), epic [ai#121](https://github.com/praxis-proxy/ai/issues/121)):

- **M1** — ordered `rules:`, each an optional static header-value match condition bound to its own budget (catch-all `default` rule when no `match:` is given)
- **M2** — reservation-based admission: reserve an estimated token cost at request time, reconcile against actual provider-reported usage (`token_count`'s `token.total`) once the response completes
- **M6** — 429 responses with `X-RateLimit-*-Tokens` headers on hard deny

Two admission algorithms, chosen per rule via `algorithm:` (`sliding_window` | `token_bucket`, per the maintainer framing on ai#789/praxis#551 that this is "a per-rule choice, similar to shadow/enforcement-action knobs elsewhere"), both behind a pluggable backend: in-process by default, or a shared Valkey backend (`backend: {kind: valkey}`) for state shared across gateway instances/replicas.

**Descoped from this PR:** M5 (per-header/per-app bucket keys, [ai#129](https://github.com/praxis-proxy/ai/issues/129)) has been pulled out. It conflicted with the quota-key design [ai#790](https://github.com/praxis-proxy/ai/pull/790) already established for the Valkey-backed shared quota work (`namespace + rule + principal + canonical model`, not an arbitrary header value), and ai#790 itself isn't approved yet. Every rule now resolves to a single shared budget regardless of request headers; ai#129 tracks per-key budgets separately pending the quota-key question being resolved with ai#790's stakeholders.

Also deliberately deferred (see `filters/src/token_rate_limit/mod.rs` module doc for the full list and rationale): configurable estimation (M3), token-type-aware weighting (M4), multiple budgets/soft-limit tiers per rule, observability, and metering.

## Additional fixes included

- Valkey backend now reuses a cached multiplexed connection instead of opening a fresh one per `reserve`/`reconcile` call.
- Fixed a doc-generator bug where `token_rate_limit`'s `AlgorithmConfig` enum collided with an unrelated struct of the same name in a sibling filter, producing phantom fields in generated docs; renamed to `RuleAlgorithm`. Also fixed the generator leaking Rust's `r#` raw-identifier prefix into the `match` field name.

## Testing

- 113 unit tests in `filters/src/token_rate_limit/`, covering both algorithms, both backends, config validation, and reservation/reconciliation edge cases (expired-reservation reaping, lost-request handling, capacity denials on all three configured bounds).
- Integration test exercising the full HTTP round trip against both example configs (`examples/configs/token-rate-limit.yaml`, `examples/configs/token-rate-limit-mixed-algorithms.yaml`).
- `make lint` (clippy, fmt, `cargo doc -D warnings`, separator/filter-doc/example-test xtask lints) passes clean.

Also fixed: every Valkey-backed test is gated on `TOKEN_RATE_LIMIT_VALKEY_URL`, but CI never set it, so they silently no-opped instead of running. Added `.github/workflows/valkey.yaml` (same shape as `postgres.yaml`) to run them for real, which surfaced a few more gaps now closed too: `ValkeyEval`'s invalidate-on-failure path, the in-process backends' unused `reconcile`/`enqueue_reconcile` trait methods, and two no-op paths in `mod.rs`.

Supersedes the closed draft [#773](https://github.com/praxis-proxy/ai/pull/773), which called out the sliding-window-vs-token-bucket question as unresolved; this PR resolves it by supporting both, per rule.


examples/configs/token-rate-limit-mixed-algorithms.yaml
```diff
@@ -0,0 +1,65 @@
...
+            cluster: backend
+
+      - filter: token_rate_limit
+        rules:
+          - name: team-alpha
+            match:
+              headers:
+                x-app-id: alpha
+            algorithm: sliding_window
+            window: 1h                    # exact trailing-window budget
```
jland-redhat, line 44: What if the user has different rates for different time windows?

I.E. I want 1000 tokens per hour but max out at 10000 tokens per day?

examples/configs/token-rate-limit-mixed-algorithms.yaml
```diff
@@ -0,0 +1,65 @@
...
+            capacity: 100000              # max tokens admitted within `window`
+            estimate_tokens: 500          # fixed cost reserved per request at admission
```
jland-redhat: nit: I feel like we should call this `reserved_tokens` estimated tokens feels more confusing to me

examples/configs/token-rate-limit-mixed-algorithms.yaml
```diff
@@ -0,0 +1,65 @@
...
+    address: "127.0.0.1:8080"
+    filter_chains:
+      - main
+
+filter_chains:
+  - name: main
+    filters:
+      - filter: router
+        routes:
+          - path_prefix: "/"
```
jland-redhat: So this would only be if a user hit `127.0.0.1:8080`?

We would in theory want to apply this to `/v1/chat/completions` right?

filters/src/token_rate_limit/config.rs
```diff
@@ -0,0 +1,353 @@
...
+    /// migration/rename path today -- routine config hygiene (e.g.
+    /// renaming `"gold"` to `"gold-tier"`) silently resets that rule's
+    /// state.
+    pub name: String,
+
+    /// Static header-value match condition. Every listed header must be
+    /// present on the request with an exact value match (`ANDed`) for
+    /// this rule to apply. Omit entirely for a catch-all rule.
+    #[serde(default)]
+    pub r#match: Option<MatchConfig>,
```
nerdalert, line 97: Thanks for documenting the matching semantics clearly. Could we make the trust boundary explicit here? These are ordinary client request headers, so a caller can omit or change a value such as `x-app-id`; with no catch-all, that makes the request unmatched and therefore unmetered. My preference would be to match trusted filter metadata established by authentication. If header matching must remain in this PR, please require/document that a preceding trusted filter overwrites the header and consider rejecting configurations without an explicit catch-all unless bypass is deliberately enabled.

filters/src/token_rate_limit/mod.rs
```diff
@@ -0,0 +1,808 @@
...
+        BodyAccess::ReadOnly
+    }
+
+    fn response_body_mode(&self) -> BodyMode {
+        BodyMode::Stream
+    }
+
+    async fn on_request(&self, ctx: &mut HttpFilterContext<'_>) -> Result<FilterAction, FilterError> {
+        let now_ms = self.now_ms();
+        let Some((rule_index, rule)) = self.matching_rule(&ctx.request.headers) else {
```
nerdalert, line 770: Could we avoid reserving quota for unrelated or invalid requests? With a catch-all rule, every request reaching this hook is charged, including readiness probes and malformed/non-inference traffic that never produces token usage; the current integration test even accounts for `GET /` consuming a reservation. That also gives unauthenticated traffic a cheap way to exhaust a shared budget. A focused fix would be to require authentication plus request/model validation before this filter and validate that ordering, or add an explicit request/path scope so only qualified inference requests reach reservation.

filters/src/token_rate_limit/mod.rs
```diff
@@ -0,0 +1,808 @@
...
+        .into());
+    }
+    if rule.estimate_tokens == 0 {
+        return Err(format!(
+            "token_rate_limit: rule '{}': estimate_tokens must be greater than 0",
+            rule.name
+        )
+        .into());
+    }
+    if rule.estimate_tokens > capacity {
```
nerdalert: Could we apply a Lua-safe integer bound to sliding-window values here? The Valkey script converts `capacity`, `estimate_tokens`, and reconciled usage with Lua `tonumber`, so integers above `2^53` are not represented exactly and may produce incorrect admission decisions. The token-bucket implementation already defines a safe-integer ceiling; reusing that validation for sliding-window capacity/estimate and rejecting oversized actual usage before reconciliation would keep the memory and Valkey backends consistent.

filters/src/token_rate_limit/backend.rs
```diff
@@ -0,0 +1,1666 @@
...
+    /// wedged), so this call's worst-case latency is up to 2x
+    /// [`VALKEY_TIMEOUT`], not [`VALKEY_TIMEOUT`] itself -- relevant
+    /// since `reserve`/`reconcile` sit synchronously on the request path.
+    async fn eval<const N: usize>(
+        &self,
+        script: &str,
+        keys: &[String; N],
+        args: &[String],
+    ) -> Result<Vec<i64>, BackendError> {
+        let mut command = redis::cmd("EVAL");
```
nerdalert, line 604: Connection reuse is a good improvement. One remaining hot-path optimization: this still sends the complete Lua source through `EVAL` for every reservation. Could we use `redis::Script`/`EVALSHA`, cache the script digest, and retry once with script loading on `NOSCRIPT`? That keeps the normal request-path command small while retaining bounded recovery after a Valkey restart or script-cache flush.

nerdalert (commented):
Thanks for putting this together. The reservation/reconciliation split and shared backend direction are solid. I found four areas where tightening the trust boundary and hot-path behavior would make the implementation safer to operate.

asaadbalum:
Reviewed the full diff, correct proposal alignment (M1/M2/M6), and both algorithms working with pluggable backends.

On Brent's review comments:
Trust boundary and probe traffic: agree these are real concerns. I'd suggest tracking them as follow-up issues rather than blocking this PR since the proposal's Non-Goals already state "assumes identity resolved by upstream," so this is a known boundary.

Approving, let's get this in.

.github/workflows/valkey.yaml
```diff
@@ -0,0 +1,155 @@
...
+
+  unit:
+    needs: [changes]
+    if: |
+      always() &&
+      (needs.changes.result == 'skipped' || needs.changes.outputs.relevant == 'true')
+    runs-on: ubuntu-24.04
+    permissions:
+      contents: read
+    services:
```
jordigilh, line 103: Flagging separately: every Valkey-backed test early-returns instead of failing when `TOKEN_RATE_LIMIT_VALKEY_URL` is unset. CI never set it, so they were silently no-opping there too, not just locally -- confirmed by killing a real Valkey mid-test and watching them fail for real.

This wires one in, same shape as `postgres.yaml`: a service container plus dedicated `unit`/`integration` jobs, split out so the fast jobs don't pay for a container on unrelated changes.

jordigilh: @nerdalert @asaadbalum FYI

jordigilh:
Two commits landed after your approval, flagging since the diff changed:

`991724a` fixes Brent's Lua-safe-integer-bound comment (`mod.rs:305`/`311`) -- `sliding_window` was missing the same 2^53 capacity check `token_bucket` already had, a real gap once Valkey's Lua scripts convert it with `tonumber`. Added regression coverage for both algorithms.

`d45e1e9`: a coverage audit found every Valkey-backed test (cross-instance state, background-worker reconciliation) silently no-ops in CI -- `TOKEN_RATE_LIMIT_VALKEY_URL` is required but was never set anywhere, so they'd been passing by skipping, not running. Added `.github/workflows/valkey.yaml` (same shape as `postgres.yaml`) to run them for real, plus tests for the gaps that surfaced once they did.

Trust-boundary and probe-traffic are unchanged, still open per your note above and `grid#101`.

Let me know if you want another look given the new commits.

filters/src/token_rate_limit/config.rs
nerdalert: Agreed that this is a real trust-boundary concern. I’m comfortable deferring it from #796 because the proposal’s non-goals explicitly assume identity has already been resolved upstream; this PR should remain the limiter foundation rather than define the authentication contract. The follow-up is now tracked in praxis-proxy/grid#101: consume trusted request-scoped identity metadata, prevent client override/spoofing, and validate authentication-before-quota ordering.

filters/src/token_rate_limit/mod.rs
nerdalert: Agreed that probe and unrelated traffic must not consume token quota. I’m comfortable treating that as follow-up integration work rather than blocking #796, provided the initial deployment guidance preserves the proposal’s upstream-validation assumption. praxis-proxy/grid#101 now explicitly tracks scoping readiness, health, malformed, unauthenticated, and other non-inference traffic out of reservation through pipeline validation or explicit request/path scope.

nerdalert:
> Reviewed the full diff, correct proposal alignment (M1/M2/M6), and both algorithms working with pluggable backends.
> 
> On Brent's review comments: Trust boundary and probe traffic: agree these are real concerns. I'd suggest tracking them as follow-up issues rather than blocking this PR since the proposal's Non-Goals already state "assumes identity resolved by upstream," so this is a known boundary.
> 
> Approving, let's get this in.

@asaadbalum  wfm! Added them to the post merge tracking issue. https://github.com/praxis-proxy/grid/issues/101

@jordigilh the additions look good. Lets get CI passing and the review issues resolved and once @jland-redhat gives a 👍 we will get it queued for merge, ty!

filters/src/token_rate_limit/mod.rs
jordigilh: Deployment guidance is in 5824f4f -- the module doc and the example config now both call out that this filter assumes upstream identity resolution, so a catch-all rule reserves quota for every request that reaches it, including probes. grid#101 tracks the follow-on scoping work.

filters/src/token_rate_limit/mod.rs
jordigilh: Fixed in 991724a -- validate_rule_bounds now rejects sliding_window capacity/estimate above MAX_F64_SAFE_INTEGER, the same ceiling token_bucket already enforced, plus a test covering the rejection.

filters/src/token_rate_limit/backend.rs
jordigilh: Agreed it's a good hot-path win, but there's no existing EVALSHA/script-cache pattern anywhere in praxis or ai to follow yet. OK to track this as a fast-follow instead of blocking here?

asaadbalum:
LGTM

asaadbalum (approved):

filters/src/token_rate_limit/backend.rs
jordigilh: Issue hand rolled https://github.com/praxis-proxy/ai/issues/831

filters/src/token_rate_limit/backend.rs
```diff
@@ -0,0 +1,1625 @@
...
+        let client = redis::Client::open(url).map_err(|e| BackendError::Unavailable(e.to_string()))?;
+        Ok(Self { client })
+    }
+
+    /// Open a fresh multiplexed connection, bounded by [`VALKEY_TIMEOUT`]
+    /// so an unreachable Valkey fails the request quickly (fail closed)
+    /// rather than hanging it indefinitely.
+    async fn connection(&self) -> Result<MultiplexedConnection, BackendError> {
+        tokio::time::timeout(VALKEY_TIMEOUT, self.client.get_multiplexed_async_connection())
+            .await
```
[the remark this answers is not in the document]
jordigilh: Fixed in d542702 -- `connection()` was opening a fresh multiplexed connection per call. Now cached and reused, invalidated only on failure/timeout.

docs/filters/token_rate_limit.md
```diff
@@ -0,0 +1,56 @@
...
+| `rules[].name` | string | yes | Human-readable rule identifier, folded into Valkey key namespacing so distinct rules sharing one backend never collide. Renaming a live `valkey`-backed rule is therefore not a no-op for operators: it changes the Valkey key hash, so the old name's tracked budget is orphaned (left to expire on its own TTL) and the new name starts with a fresh budget. There's no migration/rename path today -- routine config hygiene (e.g. renaming `"gold"` to `"gold-tier"`) silently resets that rule's state. |
+| `rules[].r#match` | MatchConfig | no | Static header-value match condition. Every listed header must be present on the request with an exact value match (`ANDed`) for this rule to apply. Omit entirely for a catch-all rule. |
+| `rules[].r#match.headers` | object<string, string> | yes | Every header must be present on the request with this exact value for the rule to match (`ANDed` across all entries). |
+| `rules[].enabled` | bool | no | Whether this algorithm is enabled. |
+| `rules[].level` | integer | no | Compression level for this algorithm. |
```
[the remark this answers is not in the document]
jordigilh: Fixed in 0013774 -- the generator was picking up an unrelated compression filter's AlgorithmConfig for the phantom enabled/level fields and not flattening the real algorithm/window/capacity/refill_rate. Renamed the enum to RuleAlgorithm to remove the name collision; the table now shows the real fields.

docs/filters/token_rate_limit.md
```diff
@@ -0,0 +1,56 @@
...
+
+## Configuration
+
+| Field | Type | Required | Description |
+|-------|------|---------|-------------|
+| `rules` | RuleConfig[] | yes | Evaluated in order; the first rule whose `match` is satisfied (or which has no `match` at all) applies to a given request. A request satisfying no rule's `match` is not rate limited by this filter instance -- add a trailing rule with no `match` to enforce a catch-all budget instead. |
+| `rules[].name` | string | yes | Human-readable rule identifier, folded into Valkey key namespacing so distinct rules sharing one backend never collide. Renaming a live `valkey`-backed rule is therefore not a no-op for operators: it changes the Valkey key hash, so the old name's tracked budget is orphaned (left to expire on its own TTL) and the new name starts with a fresh budget. There's no migration/rename path today -- routine config hygiene (e.g. renaming `"gold"` to `"gold-tier"`) silently resets that rule's state. |
```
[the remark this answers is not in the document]
jordigilh: Also fixed in 0013774, which strips the r# prefix before rendering so the table shows match, matching what serde actually deserializes.

docs/filters/token_rate_limit.md
```diff
@@ -0,0 +1,58 @@
...
+rules:
+  - name: team-alpha                 # human-readable, unique per filter instance
+    match:                           # optional: omit for a catch-all rule
+      headers:
+        x-app-id: alpha
+    algorithm: sliding_window        # sliding_window | token_bucket
+    window: 1h                       # sliding_window only: window duration
+    capacity: 100000                 # max tokens admitted (sliding_window) or held (token_bucket)
+    estimate_tokens: 500             # fixed cost reserved per request at admission
+    backend:                         # optional: defaults to in-process state
```
jland-redhat: Is there a reason to make the backends at a rule level and not just a filter level?

I get that it adds flexibility but is that something we are realistically going to want?

Also I would assume that this would require some sort of secret reference for authentication right?

examples/configs/token-rate-limit-mixed-algorithms.yaml
```diff
@@ -0,0 +1,68 @@
+# Token Rate Limiting -- Mixed Algorithms Per Rule
+#
+# Extends token-rate-limit.yaml with per-rule algorithm choice (ai#789 /
```
jland-redhat, line 3: Do we have an explanation as to what happens if traffic matches both rules? 

It could not happen in this example but you could have two different rules keyed off two different headers and that may be something you want.

I.E. I have one rule for "application" and another rule for that app's "organization" and I want both to be applied

examples/configs/token-rate-limit-mixed-algorithms.yaml
```diff
@@ -0,0 +1,68 @@
...
+            match:
+              headers:
+                x-app-id: beta
+            algorithm: token_bucket
+            capacity: 50000               # max tokens held at once
+            refill_rate: 50               # tokens refilled per second, up to `capacity`
+            estimate_tokens: 200
+
+      - filter: token_count
+        provider: openai   # openai | anthropic | google | bedrock | azure
```
jland-redhat, line 57: Does the token count filter apply to the token rate limit?

(i.e. the rate limit knows it is receiving an anthropic response if the token count is set)?

jordigilh: Yes, `token_count` will be applied before `token_rate_limit` in the response phase.

examples/configs/token-rate-limit-mixed-algorithms.yaml
jordigilh: It will use the first one that comes from the `find()` [call](https://github.com/jordigilh/praxis-ai/blob/1d4c16d742a83724a4c5aec2bd76db6c384a90a4/filters/src/token_rate_limit/mod.rs#L509-L513).

    /// The first rule (in configured order) whose `match` is satisfied
    /// by `headers`, alongside its index for reconciliation.
    fn matching_rule(&self, headers: &http::HeaderMap) -> Option<(usize, &CompiledRule)> {
        self.rules.iter().enumerate().find(|(_, rule)| rule.matches(headers))
    }

From my understanding, first-match-wins is the standard semantic for this case (envoy, nginx, firewall rules). Do you think we should consider evaluating multi rule match?

examples/configs/token-rate-limit-mixed-algorithms.yaml
jordigilh: Will updated to `/v1/chat/completions` prefix.

examples/configs/token-rate-limit-mixed-algorithms.yaml
jordigilh: Good point and worth mentioning that this is part of the M5 epic that was not yet confirmed. I implemented the plumbing parts for M5 that can be replaced with a more concrete implementation that supports multi window (#790 has already a proven implementation demo'd that covers your scenario) once M5 is confirmed.

examples/configs/token-rate-limit-mixed-algorithms.yaml
jordigilh: Done

alexsnaps:
So, just for my understanding here, is this meant to become the TRLPolicy enforcing mechanism at some point? 
It is a conscious decision that break backwards compatibility here? If I read this right probably on semantic, but certainly on "counters"? Finally, this is a 3.6 grid feature, yes? No model serving? Should we as such move this behind a feature flag? Experimental even?

nerdalert:
> So, just for my understanding here, is this meant to become the TRLPolicy enforcing mechanism at some point? It is a conscious decision that break backwards compatibility here? If I read this right probably on semantic, but certainly on "counters"? Finally, this is a 3.6 grid feature, yes? No model serving? Should we as such move this behind a feature flag? Experimental even?

Thanks for looking @alexsnaps!  I think experimental makes a lot of sense! It's currently opt-in but making it explicitly experimental wfm  since there is a lot of follow up work. WDYT @jordigilh? We should track something for how it ties back to `TRLPolicy` explicitly if there isn't already an issue. 

Tagging in cc/ @jland-redhat @asaadbalum  ty! Jamie already flagged the `TRLPolicy` CR somewhere. Ty!

docs/filters/token_rate_limit.md
jordigilh: No reason and it's makes more sense be at the filter level as you suggest: I can't see any reason why it should be per rule. I'll change that.

As for the credentials, right now it has none supported. I'll open an issue to cover this since it requires extra plumbing that can be added later. My take is to use mTLS and avoid passwords, I'll add that to the issue as an alternative to credentials.

alexsnaps:
Moving this forward in experimental is perfectly fine with me, to be clear. But the concerns wrt backwards compatibility are a big one to me, let's make sure we're (BU?) are all aligned on this. 

If we then move forward with this implementation, we need to create follow up work on a few things (don't want to slow a PoC down tho, hence the experimental proposal, but HA, clustered Redis/sharded redis-alikes, and failure scenarios need to be addressed and probably consolidated - this currently suffer from a few pathological "edge cases").

docs/filters/token_rate_limit.md
jordigilh: Issue opened https://github.com/praxis-proxy/ai/issues/833

examples/configs/token-rate-limit-mixed-algorithms.yaml
jordigilh: Done

examples/configs/token-rate-limit-mixed-algorithms.yaml
jland-redhat: Yep that makes sense to me, just want to make sure we are thinking about it from that perspecitve

jland-redhat:
> Moving this forward in experimental is perfectly fine with me, to be clear. But the concerns wrt backwards compatibility are a big one to me, let's make sure we're (BU?) are all aligned on this.
> 
> If we then move forward with this implementation, we need to create follow up work on a few things (don't want to slow a PoC down tho, hence the experimental proposal, but HA, clustered Redis/sharded redis-alikes, and failure scenarios need to be addressed and probably consolidated - this currently suffer from a few pathological "edge cases").

And just as a note from the MaaS and MaaSSubscription perspective at least. My hope is that we can make keep backwards with existing subscriptions that we can just translate differently on the backend. While also exposing this new functionality.  

Basically an existing subscription should just be a Sliding Window config with 0 reserved tokens

jland-redhat (approved):

jordigilh:
Feature gate's in: `token-rate-limit-filter` Cargo feature, off by default. The PR itself is also labeled experimental, for reviewer visibility.

TRLPolicy relationship: Will open a spike shortly and link it here once filed.

HA/clustered Valkey: Will open a spike to enumerate the failure modes before committing to a design. Will link here too.

docs/filters/token_rate_limit.md
```diff
@@ -0,0 +1,60 @@
...
+
+## Configuration
+
+| Field | Type | Required | Description |
+|-------|------|---------|-------------|
+| `rules` | RuleConfig[] | yes | Evaluated in order; the first rule whose `match` is satisfied (or which has no `match` at all) applies to a given request. A request satisfying no rule's `match` is not rate limited by this filter instance -- add a trailing rule with no `match` to enforce a catch-all budget instead. |
+| `rules[].name` | string | yes | Human-readable rule identifier, folded into Valkey key namespacing so distinct rules sharing one backend never collide. Renaming a live `valkey`-backed rule is therefore not a no-op for operators: it changes the Valkey key hash, so the old name's tracked budget is orphaned (left to expire on its own TTL) and the new name starts with a fresh budget. There's no migration/rename path today -- routine config hygiene (e.g. renaming `"gold"` to `"gold-tier"`) silently resets that rule's state. |
+| `rules[].match` | MatchConfig | no | Static header-value match condition. Every listed header must be present on the request with an exact value match (`ANDed`) for this rule to apply. Omit entirely for a catch-all rule. |
```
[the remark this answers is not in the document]
jordigilh: Added here https://github.com/praxis-proxy/ai/pull/796/changes#diff-64e2b5e28db1560f504ad88aba6a12f364bca8ea522178dbe7c531a5520d1d44R24

jordigilh:
Force pushed after rebasing and removed all commits that were not part of this PR. All comments were already addressed prior to the force push, but feel free to re-review again at your discretion.

PR #800, opened 2026-08-21
test: opt loopback fixtures into allow_private_endpoints

usize:
## Summary

praxis-core main now validates clusters defined **inline** in `load_balancer` filters and rejects private/loopback/link-local endpoints unless the config sets `insecure_options.allow_private_endpoints: true`. Core updated its own conformance fixtures and shipped examples for this change, but nothing propagated to `ai`, so the entire tree fails `test-praxis-main` (unit, schema, **and** integration steps) against core main.

Upstream commits this mirrors (no associated PR is discoverable via the API — core main uses a linear/rebase history, so referencing the commits directly):

- [`894ac18` — fix(core): validate clusters defined inline in load-balancer filters](https://github.com/praxis-proxy/praxis/commit/894ac18313ff447ae9692fdf75148b158197e584)
- [`185545a` — test(conformance): opt loopback fixtures into allow_private_endpoints](https://github.com/praxis-proxy/praxis/commit/185545ae2445c9d666772890f212cfd374117ba8)

The flag already ships in released **praxis 0.5.3**, so this fix is **release-independent** — unlike the `parse_json_rpc_body` / `cpex-policy-engine` drift tracked in #794, which is blocked on a core release. Verified green on both the 0.5.3 pin and core main.

## Changes (mirrors core's remediation)

- Append a top-level `insecure_options.allow_private_endpoints: true` block to the **52 shipped example configs** whose inline `load_balancer` cluster targets a loopback/private backend.
- Port core's `allow_loopback_endpoints()` helper into `praxis-test-utils` so harness-loaded example configs opt in centrally at load time.
- Add the flag inline to every **raw-parse fixture** (schema, integration, and the `dump`/`pipelines` unit tests) that builds an inline `load_balancer` cluster on a private endpoint.
- Merge (rather than duplicate) `insecure_options` in the two replay negative tests that append their own block onto a now-flagged example config.

Public-endpoint fixtures, listener addresses, `ip_acl` allow lists, tcp upstreams, and top-level `clusters:` blocks are left untouched.

## Verification (locally, patched against core main)

- **Unit:** 0 failed (2530 + 1072 + 59, plus `dump`/`pipelines`)
- **Schema:** 98 passed, 0 failed
- **Integration:** 469 passed, 0 failed
- **Inference fixtures:** 0 failed
- Default 0.5.3 pin: all green · `fmt`/`clippy` clean · example README + filter docs in sync

Refs #794 (the remaining, release-blocked drift items are unaffected by this PR).

### DIFF
--- a/examples/configs/a2a-agent-card-routing.yaml
+++ b/examples/configs/a2a-agent-card-routing.yaml
@@ -46,3 +46,6 @@ filter_chains:
           - name: "default"
             endpoints:
               - "127.0.0.1:9000"
+
+insecure_options:
+  allow_private_endpoints: true # example proxies to local backends
--- a/examples/configs/a2a-classifier-routing.yaml
+++ b/examples/configs/a2a-classifier-routing.yaml
@@ -127,3 +127,6 @@ filter_chains:
           - name: "default-backend"
             endpoints:
               - "127.0.0.1:9000"
+
+insecure_options:
+  allow_private_endpoints: true # example proxies to local backends
--- a/examples/configs/a2a-task-routing.yaml
+++ b/examples/configs/a2a-task-routing.yaml
@@ -87,3 +87,6 @@ filter_chains:
           - name: "agent-b"
             endpoints:
               - "127.0.0.1:9002"
+
+insecure_options:
+  allow_private_endpoints: true # example proxies to local backends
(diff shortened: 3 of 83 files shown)


franciscojavierarceo (approved):

shaneutt:
> which is blocked on a core release

ACK

Issue #801, opened 2026-06-10
Epic: Upstream provider authentication (AWS SigV4, Azure AD, GCP ADC)


shaneutt:
Provider-authentication epic for upstream inference and tool backends: AWS SigV4 (Bedrock), Azure AD (Entra ID), and GCP ADC. Groups the credential-source, workload-identity/OIDC federation, and FIPS work needed to make upstream provider authentication production-ready for v1. Rescoped from AWS-SigV4-only.

szedan-rh:
###  What & Why

**Feature:** Sign outbound requests to AWS using Signature Version 4 (SigV4).

**Context:** AWS Bedrock requires every request to be cryptographically signed. Without this implementation, Praxis cannot successfully proxy requests to Bedrock.

> AWS doesn't use simple API keys. Instead, every request must be signed. We compute a cryptographic hash of the entire request (URL + headers + body) using your AWS secret key. This proves to AWS that **(1)** you have the key, and **(2)** nobody tampered with the request in transit. *It's like sealing an envelope with a personal wax stamp that only you possess.*

### Example: How it works

**1. Client sends a standard request to Praxis**
*(Notice: No AWS credentials are required from the client)*

```bash
curl http://praxis:8080/v1/chat/completions \
  -d '{"model":"anthropic.claude-3","messages":[...]}'

```

```http
POST /v1/chat/completions HTTP/1.1
Host: praxis:8080

```

**2. The Praxis SigV4 Filter intercepts the request**

* Computes the SHA-256 hash of the request body.
* Builds the canonical request string.
* Signs it with HMAC-SHA256 using the configured AWS secret key.

**3. Praxis sends the signed request to AWS Bedrock**

```http
POST /model/anthropic.claude-3/invoke HTTP/1.1
Host: bedrock-runtime.us-east-1.amazonaws.com
Authorization: AWS4-HMAC-SHA256 Credential=AKIA.../20260821/us-east-1/bedrock/aws4_request, SignedHeaders=host;x-amz-date;x-amz-content-sha256, Signature=a3f2b8c9d...
x-amz-date: 20260821T192000Z
x-amz-content-sha256: e3b0c44298fc...
Content-Type: application/json

{"model":"anthropic.claude-3","messages":[...]}

```

szedan-rh:
### Implementation Plan

* [ ] **Dependencies:** Add the `aws-sigv4` or `ring` crate to `filter/Cargo.toml`.
[...]

PR #802, opened 2026-08-23
aws: add aws_sigv4_sign filter for signing upstream AWS requests

szedan-rh:
## Summary

Adds an `aws_sigv4_sign` filter that signs outbound requests to AWS
services (e.g. Bedrock) using SigV4, built on the official
`aws-sigv4`/`aws-credential-types` crates. Covers static/env-var
credentials only; OIDC and per-backend credentials are follow-ups.

## Related issue

Part of #801 (not closing — credential files, OIDC, and per-backend
config from the issue are deferred follow-ups).


szedan-rh:
> **Unsigned commits**: [4eea4e0](https://github.com/praxis-proxy/ai/commit/4eea4e0a37e0e0bde1e475a265fd84eb85cf2b34), [7df566a](https://github.com/praxis-proxy/ai/commit/7df566ac548c3268add8d8978e537c6b4595cf6e), [7be53ab](https://github.com/praxis-proxy/ai/commit/7be53abc810993dc619bf83cf6e299ab0f1df3fe), [19572b4](https://github.com/praxis-proxy/ai/commit/19572b41ce21e0972bb39a345715a7284d50998c), [d189969](https://github.com/praxis-proxy/ai/commit/d1899693615df7e3c48340c91b22fd483ac38587), [07aabab](https://github.com/praxis-proxy/ai/commit/07aabab2eec271600a1907a596ea26fdea993d96), [1482a75](https://github.com/praxis-proxy/ai/commit/1482a75831b28368950e473365229a5d296b8745), [7faac9c](https://github.com/praxis-proxy/ai/commit/7faac9cb25510ed9f4e83bfc4d6373bfd616ac11). Please [sign your commits](https://docs.github.com/en/authentication/managing-commit-signature-verification).

all commits are signed off

szedan-rh:
PR in testing with bedrock account,

szedan-rh:
# `aws_sigv4_sign` — live verification against real AWS Bedrock

Verified the filter end-to-end against the real **Amazon Bedrock**
runtime API. A request routed **through the proxy** is signed with
SigV4 and accepted by AWS (**HTTP 200**); the **same request sent
unsigned** is rejected (**HTTP 403**). Since signing is the only
variable between the two, the 200 is attributable to the filter.

- **Branch commit:** `3ad1e93`
- **Praxis core:** `praxis-proxy` `0.5.3`
- **Region:** `us-east-1`
- **Model:** `amazon.nova-micro-v1:0` (cheapest on-demand Bedrock text model)
- **Endpoint:** `bedrock-runtime.us-east-1.amazonaws.com:443` (TLS)
- **Prompt:** `What is Moore's law?`

## Config under test

```yaml
listeners:
  - name: gateway
    address: "127.0.0.1:8080"
    filter_chains:
      - bedrock

filter_chains:
  - name: bedrock
    filters:
      - filter: router
        routes:
          - path_prefix: "/model/"
            cluster: bedrock
      - filter: load_balancer
        clusters:
          - name: bedrock
            endpoints:
              - "bedrock-runtime.us-east-1.amazonaws.com:443"
            tls:
              sni: bedrock-runtime.us-east-1.amazonaws.com
      - filter: aws_sigv4_sign
        region: us-east-1
        service: bedrock
        host: bedrock-runtime.us-east-1.amazonaws.com
        access_key_env_var: AWS_ACCESS_KEY_ID
        secret_key_env_var: AWS_SECRET_ACCESS_KEY
```

## Test 1 — through the proxy (filter signs the request) → ✅ 200

```console
$ curl -sS -w "\n---HTTP_STATUS: %{http_code}---\n" \
    "http://127.0.0.1:8080/model/amazon.nova-micro-v1:0/invoke" \
    -H 'Content-Type: application/json' \
    -d '{"schemaVersion":"messages-v1","messages":[{"role":"user","content":[{"text":"What is Moore'\''s law?"}]}],"inferenceConfig":{"maxTokens":300}}'

{"output":{"message":{"content":[{"text":"Moore's Law is a fundamental observation in the field of microelectronics and computing, named after Gordon Moore, the co-founder of Intel Corporation. The law states that the number of transistors that can be placed inexpensively on an integrated circuit doubles approximately every two years. This observation has been a key driver of the rapid advancement in computing power and has been remarkably accurate since its formulation in 1965.\n\nMoore's Law implies that the computational power of microprocessors will continue to increase exponentially, which translates into smaller, faster, and more efficient electronic devices. This has had profound implications for various technological fields, including computing, telecommunications, and consumer electronics.\n\nIt's important to note that while Moore's Law has held remarkably true for several decades, there are ongoing discussions in the tech community about its future as we approach physical limits in semiconductor technology. Researchers are exploring new materials and techniques, such as quantum computing, to continue the trend of advancing computing capabilities."}],"role":"assistant"}},"stopReason":"end_turn","usage":{"inputTokens":7,"outputTokens":197,"totalTokens":204,"cacheReadInputTokenCount":0,"cacheWriteInputTokenCount":0}}
---HTTP_STATUS: 200---
```

AWS authenticated the SigV4 signature and returned a real model
completion.

## Test 2 — negative control: same request unsigned → ❌ 403

```console
$ curl -sS -w "\n---HTTP_STATUS: %{http_code}---\n" \
    "https://bedrock-runtime.us-east-1.amazonaws.com/model/amazon.nova-micro-v1:0/invoke" \
    -H 'Content-Type: application/json' \
    -d '{"schemaVersion":"messages-v1","messages":[{"role":"user","content":[{"text":"What is Moore'\''s law?"}]}],"inferenceConfig":{"maxTokens":300}}'

{"Message":"Authorization header is missing"}
---HTTP_STATUS: 403---
```

Without the filter, AWS rejects the identical request for lack of a
signature — confirming the 200 in Test 1 came from the filter's SigV4
signing, not from any ambient auth.

## Result

| Path | Signature | HTTP | Outcome |
|------|-----------|------|---------|
| Through proxy (`aws_sigv4_sign`) | present | **200** | Model completion returned |
| Direct, unsigned (control) | absent | **403** | `Authorization header is missing` |

filters/src/aws/sigv4.rs
```diff
@@ -0,0 +1,672 @@
...
+            &self.region,
+            &self.service,
+            SystemTime::now(),
+            ctx.request.method.as_str(),
+            &uri,
+            std::iter::once(("host", self.host.as_str())),
+            &body,
+        );
+
+        let Ok(signed_headers) = signed_headers else {
```
aslakknutsen: "When sign_headers returns Err, on_request discards the FilterError and rejects with a bare 503 (lines 285-288). Fail-closed behavior is correct, but operators cannot distinguish signing misconfiguration, clock skew, or library rejection from generic upstream unavailability. Add at least a tracing::warn! with a redacted reason before rejecting."

aslakknutsen (changes requested):

filters/src/aws/sigv4.rs
```diff
@@ -0,0 +1,672 @@
...
+//! environment variable at filter construction time). OIDC / web
+//! identity federation and the AWS default credential chain are a
+//! planned follow-up that will plug into the same
+//! [`aws_credential_types::provider::SharedCredentialsProvider`] seam
+//! this filter already uses internally, with no change to the signing
+//! or request-handling code.
+//!
+//! # FIPS considerations
+//!
+//! Signing uses the [`aws-sigv4`](https://docs.rs/aws-sigv4) crate,
```
aslakknutsen, line 50: @alexsnaps Heads up...

filters/src/aws/sigv4.rs
```diff
@@ -0,0 +1,672 @@
...
+impl Sigv4SignFilter {
+    /// Build a filter instance from parsed config, resolving static
+    /// credentials from the configured environment variables.
+    ///
+    /// # Errors
+    ///
+    /// Returns [`FilterError`] if either credential environment
+    /// variable is unset or not valid UTF-8.
+    fn new(config: Sigv4SignConfig) -> Result<Self, FilterError> {
+        let access_key = std::env::var(&config.access_key_env_var).map_err(|e| {
```
aslakknutsen: std::env::var requires Deployment env change and a restart of the Proxy pod to update values. 

I'm not sure we have a great Core solution for this yet, but e.g. a pattern is to support Env, File, or Inline: 
https://github.com/praxis-proxy/ai/blob/main/filters/src/routing/credential_inject.rs#L128-L145

shaneutt (commented):
First pass looks OK, I agree with many of the comments provided already.

Once those comments are resolved I'll take another pass 👍

szedan-rh:
Thanks, addressed all four in a fixup commit:

1. Signing error dropped — on_request now logs before failing closed. Used tracing::warn!(error = %e, ...) rather than error! since this is a per-request path that can be triggered by client input; the error text only contains signing-input diagnostics (URI/header/params), never credential material.

2. No test for the 503 path — added on_request_rejects_with_503_when_signing_fails, which forces SignableRequest::new to fail via a control character in rewritten_path and asserts Reject(status = 503) plus no partially-set headers.

3. Host validated per-request — the Host value is now validated once in new() and stored as a HeaderValue; the request path just moves it into request_headers_to_set. Bad host config now fails at startup.

4. max_body_bytes unconstrained — new() now rejects 0 and values above a 64 MiB ceiling. Note there's no #[serde(try_from)] usage anywhere in this repo yet, so I followed the existing post-parse-validation style (as in the mcp filter) instead of introducing the pattern; happy to switch to try_from if you'd prefer to establish it here.

filters/src/aws/sigv4.rs
szedan-rh: Fiixed.

filters/src/aws/sigv4.rs
szedan-rh: still in discussion

filters/src/aws/sigv4.rs
szedan-rh: Good point on the Env/File/Inline seam. This PR intentionally scoped credentials to static env-var only, with OIDC / credential files called out as follow-ups (tracked under #801). Two options: I can adopt the credential_inject-style Env/File/Inline source enum in this PR now, or keep it env-only here and fold File/Inline into the same follow-up as OIDC. 
Do you consider the File/Inline sources blocking for this PR, or acceptable as the next one? Happy to do either  just want to keep this change reviewable.

aslakknutsen: Not blocking. As long as it's tracked that's fine.

shaneutt (approved):
Let's create some follow-ups if we need to (e.g. FIPS followup) and keep iterating on this 👍

Issue #811, opened 2026-06-10
Azure AD / Entra ID upstream authentication


shaneutt:
## What

Obtain and inject Azure AD (Entra ID) access tokens for upstream requests to Azure OpenAI and other Azure AI services.

## Why

Azure OpenAI requires either API keys or Azure AD tokens. Enterprise Azure deployments use Entra ID with managed identities and OIDC federation. Without this, Praxis cannot proxy to Azure OpenAI in enterprise environments that prohibit static API keys.

## Goals

- Azure AD token acquisition via client credentials (client secret, client certificate)
- OIDC federation support
- Managed identity support for AKS deployments
- Automatic token refresh before expiry
- Per-backend credential configuration

szedan-rh:
```markdown
### 🔐 What & Why: Azure AD (Entra ID) Token Injection

Obtain and inject Azure AD (Entra ID) access tokens for Azure OpenAI. Enterprise Azure environments prohibit static API keys — they require OAuth2 tokens from Entra ID.

> **💡 In plain words:** 
> Azure OpenAI doesn't let you just send an API key in production enterprise environments. Instead, you need to log in to Microsoft's identity system (Entra ID) and get a short-lived "access token" (like a 1-hour guest pass). 
> 
> Praxis handles this login automatically in the background, caches the token, and refreshes it before it expires. **Your app never knows about Azure AD.**

---

### ⚙️ Example: How it works

#### 1️⃣ Background Process (happens automatically every ~55 min)
```http
# Praxis requests a new token
POST [https://login.microsoftonline.com/](https://login.microsoftonline.com/){tenant}/oauth2/v2.0/token
client_id=xxx&client_secret=yyy&scope=cognitiveservices

# Azure responds (200 OK)
{ 
  "access_token": "eyJ0eXAi...", 
  "expires_in": 3600 
}
# 📌 Token is cached in memory (ArcSwap for lock-free reads)

```

#### 2️⃣ Per-Request Flow

```bash
# Your app's request (Unaware of Entra ID):
curl http://praxis:8080/v1/chat/completions \
  -d '{"model":"gpt-4o","messages":[...]}'

```

*↓ Praxis reads the cached token and injects it ↓*

```http
# What Azure OpenAI actually receives:
POST /openai/deployments/gpt-4o/chat/completions HTTP/1.1
Host: mycompany.openai.azure.com
Authorization: Bearer eyJ0eXAiOiJKV1Qi...
Content-Type: application/json

[...]

PR #815, opened 2026-08-25
feat(azure): add azure_ad upstream authentication filter

szedan-rh:
## Summary

Adds an `azure_ad` filter that authenticates upstream requests to Azure OpenAI (and other Azure AI services) with Entra ID, resolving [#811](https://github.com/praxis-proxy/ai/issues/811).

It acquires an OAuth2 bearer token via the **client-credentials** grant, caches it in a lock-free `ArcSwap`, refreshes it in the background before expiry, and injects `Authorization: Bearer <token>` on every proxied request. The downstream client never sees Entra ID.

## Design

- **Injects only the `Authorization` header.** Unlike SigV4, an Entra token is not bound to Host, path, or body, so routing stays the operator's responsibility (cluster `endpoints` + `tls.sni`) and the filter has no ordering constraint relative to path-rewrite filters.
- **Secret via env var, never YAML.** The client secret is read from the environment variable named in `client_secret_env_var`.
- **Non-blocking acquisition.** Token fetch runs on a dedicated background thread with its own runtime + `CancellationToken` shutdown (the `routing::overlay` pattern), so building or hot-reloading a pipeline never blocks on a network round-trip.
- **Fails closed → 503** until the first token is acquired and whenever the cached token is missing or expired — never forwards unauthenticated.

## Scope

Client-secret credential only. Managed identity (AKS/IMDS), client certificates (`private_key_jwt`), and OIDC/workload-identity federation slot into the same cache/refresh machinery as follow-ups, with no change to request handling.

## Tests

- 14 unit tests (config parsing, token fetch/parse, refresh timing, header injection, fail-closed paths)
- 2 integration tests against the example config (`examples/configs/azure-ad.yaml`), including a deterministic offline fail-closed check
- Registry tests updated to cover the new security filter

## Verification

`cargo clippy`, `cargo +nightly fmt --check`, all unit/integration/registry tests, and the xtask gates (filter-docs, example-tests, example-readme, separators, deps) pass.


szedan-rh:
Under testing, please don't start the review yet.

szedan-rh:
Test the change with api key not through the filter.

filters/src/azure/azure_ad.rs
```diff
@@ -0,0 +1,980 @@
...
+            None
+        },
+    }
+}
+
+/// Repeatedly acquire a token, publish it to the shared cache, and sleep
+/// until the next refresh — until cancelled.
+async fn refresh_loop(params: RefresherParams, shutdown: CancellationToken) {
+    let client = match reqwest::Client::builder().timeout(TOKEN_REQUEST_TIMEOUT).build() {
+        Ok(client) => client,
```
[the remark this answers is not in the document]
shaneutt: Additionally, we need to add an issue to the ai repository that tracks replacing reqwest in time (e.g. like core did in v0.4.0)

filters/src/azure/azure_ad.rs
```diff
@@ -0,0 +1,980 @@
...
+            }
+            drop(handle.join());
+        }
+    }
+}
+
+/// Spawn the background token refresher on a dedicated thread with its
+/// own current-thread runtime (mirrors `routing::overlay`), so token
+/// acquisition never blocks the pipeline build or the request hot path.
+fn spawn_refresher(params: RefresherParams) -> RefreshHandle {
```
aslakknutsen, line 290: Filters should really not be dealing with their own Threads, but we're currently lacking any Core solution for cache refresh/background tasks. WIP

alexsnaps: But couldn't the filter go straight to tokio? Not that I think we should support this moving forward, but spawning threads should really be avoided. 

```rust
tokio::spawn(async move {
    loop {
        interval.tick().await; // Wait for the next interval
        do_work().await;
    }
}).await.unwrap();

async fn do_work() {
    println!("Running recurring task...");
}
```

aslakknutsen: For the "thread" part it probably could. The Lifecycle / monitoring / leak etc could still be a potential issue.

alexsnaps: Totally. This can _NOT_ be a long term solution here.

filters/src/azure/azure_ad.rs
```diff
@@ -0,0 +1,980 @@
...
+    let shutdown = CancellationToken::new();
+    let token = shutdown.clone();
+
+    let thread = std::thread::Builder::new()
+        .name("azure-ad-token-refresher".to_owned())
+        .spawn(
+            move || match tokio::runtime::Builder::new_current_thread().enable_all().build() {
+                Ok(runtime) => runtime.block_on(refresh_loop(params, token)),
```
alexsnaps, line 298: That's actually _not_ creating a new thread tho.

alexsnaps: That isn't but... yeah, now I see - sorry - ignore this!

shaneutt (approved):

filters/src/azure/azure_ad.rs
shaneutt: Agreed. Create an issue in core for this before merging this PR and let's work together on a solution, and then iterate here.

alexsnaps: I'd still argue if [this approach](https://github.com/praxis-proxy/ai/pull/815#discussion_r3854487847) works, I'd do that for now. And _later_ make sure we 🔥 that as well and replace with "whatever solution we come to for this". 

Where I think "this" is interesting. There is spawning background tasks and then there is "background cache refreshes" which I'd consider separately (but base on the former tho).

shaneutt:
We will need to iterate on this. @szedan-rh before we merge can you please ensure everything here is behind an experimental build flag please?

filters/src/azure/azure_ad.rs
aslakknutsen: https://github.com/praxis-proxy/praxis/issues/1042

szedan-rh: Thanks all. I looked into whether `tokio::spawn` works at the point this runs, and it can't: `server/src/main.rs` `main()` is a plain `fn` (no `#[tokio::main]`), and pipeline construction (`resolve_pipelines` → `FilterPipeline::build_with_chains` → each filter's `from_config`) is fully synchronous and runs *before* Pingora starts any runtime. So `AzureAdFilter::new` executes with no `Handle::current()`  a bare `tokio::spawn` there panics with "there is no reactor running."

That's why this uses a dedicated thread with its own current-thread runtime, and it mirrors the existing idiom for the same reason: the overlay watcher (`filters/src/routing/overlay.rs`), the config watcher (`server/src/watcher.rs`), and the health checks (`server/src/reload.rs`, `server/src/server.rs`) all do the same.

So I agree this shouldn't be the long-term home for it — but the fix isn't `tokio::spawn` (would break at boot); it's a core primitive. I've opened praxis-proxy/praxis#1043 to track a shared background-task / cache-refresh mechanism, and I'll migrate this filter once it lands. I'd suggest keeping the current approach as the interim since it matches the established pattern, unless you'd prefer to hold this PR until the core work is ready.

szedan-rh: @alexsnaps @aslakknutsen @shaneutt

szedan-rh:
@shaneutt done, everything for this filter is now behind a new `azure-ad-filter` cargo feature (off by default, activates the `experimental` marker), mirroring `http-callout-filter`. Pushed in 130ce5e.

- The `azure` module, its registration, and the example integration test are `#[cfg]`-gated; the server and integration-test crates forward the feature.
- Registry tests assert the filter is **present** with the feature and **absent** without it.
- Default build/CI (`make test-unit`, `make lint`) run with default features, so it stays off there, same as the existing experimental filters.

Verified: default build (feature off) + feature-on build both clean; clippy clean both ways; azure unit tests (22) and the gated integration example test pass; `cargo machete`/dep + doc/example lints green.

Happy to iterate on moving the background token refresher onto the core primitive once praxis-proxy/praxis#1043 lands.

alexsnaps (approved):
As per @shaneutt 's comment, there are things we have to iterate upon before promoting this "out of experimental", but given the feature flag, let's merge this and we can address following up as part of the "experimental clean up phase" later. 

Thanks @szedan-rh

Issue #49, opened 2026-06-11
ai_guardrails: NeMo mask / redact action


liavweiss:
### Description

Implement the redaction flow: when NeMo returns `"modified"` status,
extract the masked content and replace the original request body.
Link for the Proposal: [docs/proposals/00138_external_guardrails.md](https://github.com/praxis-proxy/praxis/blob/main/docs/proposals/00138_external_guardrails.md)

* When NeMo returns `"modified"` status, the redacted content is available in the NeMo response at: `content` filed
* `NemoProvider` extracts this redacted text and returns it as `GuardResult::Redact { modified_text, reason }`
* In `filter.rs`: replace the original body bytes with the redacted content, update the `Content-Length` header
* notice that if redact was happen, we should change all messages that were redacted.

NeMo deployment for testing:
* [TrustyAI Helm chart](https://github.com/trustyai-explainability/trustyai-llm-demo/tree/add-mcp-guardrails/mcp-guardrails/deploy)
* You can use the latest image from trusty.

Acceptance:
* Request containing PII/sensitive content is forwarded with masked body when NeMo returns `"modified"`
* `Content-Length` header is updated to match the new body size
* `FilterResultSet` status is written as `"redacted"`

### Motivation

Pass and block are binary outcomes, but many guardrail policies need a
third option: mask sensitive content and forward the sanitized request.
NeMo supports this via its `"modified"` status, and the filter needs to
extract the redacted text and rewrite the body in-flight.

### Alternatives Considered

Rejecting the entire request when sensitive content is detected instead
of redacting. This is already supported via the `Block` result; redaction
adds a less disruptive option for operators who want to sanitize rather
than reject.

### reference:
#1044 

### Parent Epic (optional)

praxis-proxy/ai#76

PR #821, opened 2026-08-26
Revert "fix(guardrails): fail closed on NeMo redact verdict until body replac…"

szedan-rh:
Reverts praxis-proxy/ai#721

The Continue on GuardResult::Redact was intentional - we have a dedicated task for implementing the actual redaction flow in https://github.com/praxis-proxy/ai/issues/49, which includes extracting the masked content from NeMo and replacing the request body in-flight. The Continue was a placeholder until that work lands.

The /v1/guardrail/checks endpoint does not support redaction today - TrustyAI originally planned to add it but decided not to. Redaction ("modified" status + transformed content) is only available on the /v1/checks endpoint. We are planning to migrate to /v1/checks to support redaction.

In the meantime, I have https://github.com/praxis-proxy/ai/pull/792 which removes the "modified" match arm entirely since /v1/guardrail/checks never returns it. This aligns the code with the actual endpoint behavior and eliminates the misleading redacted status.

### DIFF
--- a/examples/configs/nemo-guardrails.yaml
+++ b/examples/configs/nemo-guardrails.yaml
@@ -5,7 +5,7 @@
 #
 #   passed   → request forwarded to the upstream unchanged
 #   blocked  → 403 returned to the client (triggered rail names in body)
-#   modified → 403 returned to the client (triggered rail names in body)
+#   modified → forwarded unchanged (body replacement deferred to #579)
 #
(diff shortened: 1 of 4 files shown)


aslakknutsen:
@szedan-rh missing signoff

aslakknutsen (approved):

szedan-rh:
@aslakknutsen - signed

PR #829, opened 2026-08-27
fix(azure_ad): schedule token refresh within skewed validity window

szedan-rh:
## Summary

Fixes a scheduling gap in the `azure_ad` token refresher that produces deterministic windows of 503s, plus three smaller hardening changes on the same file.

**Refresh scheduling (the bug):** the next refresh was scheduled at `refresh_ratio * ttl`, while the cached token is treated as expired at `ttl - skew`. Whenever `ratio * ttl` lands past `ttl - skew`, the cached token goes invalid before the refresh fires, so every cycle has a window where all requests fail closed against a perfectly healthy token endpoint. At the defaults (`refresh_ratio` 0.75, 30s skew) this hits any token with a TTL under 120s — e.g. a 60s token is valid for 30s but only refreshed at 45s, a guaranteed 15s outage per 45s cycle. A high `refresh_ratio` (e.g. 0.99) reproduces it even with standard 3600s tokens. The refresh is now scheduled from the skew-adjusted usable lifetime (`ttl - skew`) via a new `publish_token` helper, and the short-TTL regression test now asserts the cached token is still valid at the scheduled refresh instant (the assertion fails without the fix).

Also in this change:

- The token fetch is raced against cancellation in the refresh loop, so dropping a pipeline (hot reload, shutdown) is not blocked behind an in-flight token request for up to its 30s timeout — `RefreshHandle::drop` only waits its 2s join bound.
- The missing-token condition is logged on state transitions (fail-closed entered / token available again) instead of once per rejected request, so a token outage under load cannot flood the logs.
- The request path reads the token cache with `arc_swap`'s `load()` instead of `load_full()`, avoiding an atomic refcount bump per request.

Behavior note: the refresh cadence shifts slightly earlier for healthy tokens (a 3600s token refreshes at ~2677s instead of 2700s); no configuration changes are required.

## Related issue

Follow-up hardening on the filter added for #811. The refresher-thread pattern itself is slated to move onto the shared background-task primitive tracked in praxis-proxy/praxis#1042 and praxis-proxy/praxis#1043; these fixes apply to the current implementation until that lands.

## Validation

- [x] Unit tests — `cargo test -p praxis-ai-filters --features azure-ad-filter azure` (22 passed, including the extended short-TTL regression test)
- [x] Integration or functional tests — no change to the example config path; the existing gated `examples::azure_ad` test still passes
- [x] `make lint` (`cargo clippy -p praxis-ai-filters --all-targets --features azure-ad-filter -- -D warnings` clean; note the default-features lint does not compile this gated module — #828 adds experimental-feature coverage to the make targets)

### DIFF
--- a/docs/filters/azure_ad.md
+++ b/docs/filters/azure_ad.md
@@ -20,5 +20,5 @@ client_id: 11111111-1111-1111-1111-111111111111
 scope: https://cognitiveservices.azure.com/.default
 client_secret_env_var: AZURE_CLIENT_SECRET
 authority_host: login.microsoftonline.com   # optional, for sovereign clouds
-refresh_ratio: 0.75                          # optional, refresh at 75% of TTL
+refresh_ratio: 0.75                          # optional, refresh at 75% of the usable lifetime
 ```
--- a/filters/src/azure/azure_ad.rs
+++ b/filters/src/azure/azure_ad.rs
@@ -60,7 +60,10 @@
 //! ```
 
 use std::{
-    sync::Arc,
+    sync::{
+        Arc,
+        atomic::{AtomicBool, Ordering},
+    },
     thread::JoinHandle,
     time::{Duration, Instant},
 };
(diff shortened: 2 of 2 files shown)


aslakknutsen (commented):
docs/filters/azure_ad.md say refresh at 75% of TTL

aslakknutsen (approved):

Issue #50, opened 2026-06-11
ai_guardrails: NeMo response-side guardrails


liavweiss:
### Description

Inspect LLM responses before they reach the client using the same
NeMo provider.
Link for the Proposal: [docs/proposals/00138_external_guardrails.md](https://github.com/praxis-proxy/praxis/blob/main/docs/proposals/00138_external_guardrails.md)

* Reuse the same NeMo provider for response inspection — `extract_messages()` already handles response body formats (`choices[].message.content`, MCP `result.content[].text`)
* Call `provider.evaluate(messages, Response)` at end-of-stream
* Act on result: Pass → forward, Block → replace response with error, Redact → replace body with masked text
* Set `response_body_access()` to `ReadWrite` and `response_body_mode()` to `StreamBuffer`

NeMo deployment for testing:
* [TrustyAI Helm chart](https://github.com/trustyai-explainability/trustyai-llm-demo/tree/add-mcp-guardrails/mcp-guardrails/deploy)
* Debug image for E2E testing: `quay.io/rgeada/nemo-debug:latest` ([quay.io](https://quay.io/repository/rgeada/nemo-debug?tab=tags&tag=latest))

**Blocked:** `on_response_body` is currently sync in the `HttpFilter`
trait (Pingora constraint). This task requires making `on_response_body`
async. See "Investigate async `on_response_body` support".

Acceptance:
* LLM response with policy violations is blocked before reaching the client
* LLM response with PII is masked before reaching the client
* Provider errors propagate as `FilterError`

### Dependencies

* Depends on "[ai_guardrails: NeMo provider and request-side integration](https://github.com/praxis-proxy/ai/issues/48)"
* Depends on "[Investigate async `on_response_body` support](https://github.com/praxis-proxy/ai/issues/51)"

### Motivation

Request-side guardrails protect the LLM from malicious input, but
operators also need to inspect LLM output before it reaches the client
(e.g. PII leakage, policy violations in generated content). This
completes the bidirectional guardrail coverage.

### Alternatives Considered

None

### Area

Filter Pipeline

### Parent Epic (optional)

praxis-proxy/ai#76

liavweiss:
https://github.com/praxis-proxy/ai/pull/832 - merged

Issue #51, opened 2026-06-11
Investigate async `on_response_body` support


liavweiss:
### Description

`on_response_body` is currently synchronous in the `HttpFilter` trait
due to a Pingora constraint. External guardrail providers (and any
filter that needs async HTTP calls during response processing) require
an async version of this hook.

* Investigate the Pingora `response_body_filter` hook — it is defined as a sync `fn`, which propagates to Praxis's `on_response_body`
* Evaluate feasibility of a sync-to-async bridge at the Pingora boundary.
* Consider whether upstream Pingora changes are needed or if this can be solved within Praxis
* Document findings and recommended approach

Context:
* `on_request_body` is already async — this is only a response-side limitation
* "[ai_guardrails: NeMo response-side guardrails](https://github.com/praxis-proxy/ai/issues/50)" is blocked on this
* Any future filter that needs to make HTTP calls during response processing is also blocked

### Motivation

Multiple features require async operations during response body
processing (AI guardrails response-side, HTTP callout filter for
responses, etc.). Without this, any filter that needs to call an
external service during response processing cannot be implemented.

### Alternatives Considered

Running the external call synchronously with `block_on` — this risks
blocking the Pingora event loop and degrading proxy throughput. Needs
investigation to determine if it is safe in practice.

### Area

Filter Pipeline

### Parent Epic (optional)

praxis-proxy/ai#76

liavweiss:
## Investigation: Async `on_response_body` Support

### Summary

`on_response_body` is synchronous because Pingora's
`ProxyHttp::response_body_filter` is a sync `fn`. This propagates
through the entire Praxis stack. The request-side equivalent
(`request_body_filter`) **is** async - the asymmetry is an intentional
Pingora design choice due to boxed-future allocation concerns.

**The constraint originates in Pingora's `ProxyHttp` trait**, not in
Praxis.

### Options

#### Option A: Make `response_body_filter` async

Make `response_body_filter` an `async fn` - either in the
`praxis-proxy/pingora` fork ourselves, or by pulling an upstream
change if/when Cloudflare merges one.

> **Note:** The Praxis-side work is the same regardless of who writes
> the Pingora change. Even if upstream merges an async
> `response_body_filter`, we still need to pull the new Pingora
> version and update every Praxis filter that implements
> `on_response_body`.

**Upstream status:**

| Issue | Opened | Status |
|-------|--------|--------|
[...]

PR #832, opened 2026-08-27
Guardrails: Implement response phase

liavweiss:
## Summary

- Implement response-side guardrail evaluation in the `ai_guardrails`
  filter, enabling NeMo to inspect upstream responses before they reach
  the client.
- Add `on_response_body` using `block_in_place` to bridge the sync
  Pingora constraint into the async `GuardProvider::evaluate` call
  (tracked in #51 for a proper async solution).
- Unify `record_verdict` to handle both request and response phases:
  request-phase blocks return a clean 403; response-phase blocks replace
  the body with a JSON error payload padded to the committed
  `Content-Length` (Pingora commits headers before `on_response_body`
  runs, so status code cannot be changed).
- Add `extract_response_messages` to parse OpenAI Chat Completion
  response format (`choices[].message`), fail-closed on unrecognized
  bodies.
- Add `fit_to_committed_length` utility that pads (space) or truncates
  the replacement body to match the original `Content-Length`, preventing
  `PrematureBodyEnd` framing errors.
- Remove the `phase.response: true` rejection guard from `from_config`
  and the `#[dead_code]` suppression on `GuardPhase::Response`.
- Add 18 new tests covering response body access, skip conditions,
  validation (fail-closed), provider verdicts (pass/block/error via
  wiremock), and `fit_to_committed_length` edge cases (equal, padding,
  truncation, `None` body).
- Update `nemo-guardrails.yaml` example to document both request and
  response phase behavior.

## Validation

- [x] Unit tests: 66 guardrails tests pass (`cargo test -p praxis-ai-filters --lib -- guardrails`)
- [x] Manual testing against real NeMo Guardrails instance (port 18000) and mock vLLM upstream
- [x] Verified response-phase blocking produces valid JSON error with correct padding
- [x] Verified request-phase behavior unchanged (403 on block, pass-through on success)
- [x] `make lint` (no new warnings)

## Checklist

## Breaking changes

None. The `phase.response` config field already existed with a default of
`false`. Existing configs that omit it or set it to `false` are unaffected.

## Fix
https://github.com/praxis-proxy/ai/issues/50


leseb:
@liavweiss did you address the bot comments?

liavweiss:
@leseb, I addressed them now.
I also ran all the unit and integration tests I ran previously to make sure everything is still working as expected.

filters/src/guardrails/filter.rs
```diff
@@ -136,41 +128,166 @@ impl HttpFilter for AiGuardrailsFilter {
...
+
+        let Some(bytes) = body.as_ref() else {
+            return Ok(FilterAction::Continue);
+        };
+
+        if bytes.is_empty() {
+            return Ok(FilterAction::Continue);
+        }
+
+        let messages = extract_response_messages(bytes)?;
```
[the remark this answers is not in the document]
liavweiss: Done.

filters/src/guardrails/filter.rs
```diff
@@ -136,41 +128,166 @@ impl HttpFilter for AiGuardrailsFilter {
...
+        }
+    }
+
+    fn response_body_mode(&self) -> BodyMode {
+        BodyMode::StreamBuffer {
+            max_bytes: Some(DEFAULT_MAX_BODY_BYTES),
+        }
+    }
+
+    fn on_response_body(
```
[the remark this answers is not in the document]
liavweiss: I'll add integration coverage for both request and response phases in a dedicated follow-up PR.

filters/src/guardrails/filter.rs
```diff
@@ -167,41 +171,171 @@ impl HttpFilter for AiGuardrailsFilter {
...
+
+        let Some(bytes) = body.as_ref() else {
+            return Ok(FilterAction::Continue);
+        };
+
+        if bytes.is_empty() {
+            return Ok(FilterAction::Continue);
+        }
+
+        if ctx.get_metadata(META_RESPONSE_SSE).is_some() {
```
christinaexyou: IIUC, for SSE, we still wait for the stream to finish (buffer) and forward along the buffered bytes. as a result, we modify streaming endpoints into buffered ones which can cause them to break. 

can we default to streaming instead and only buffer when we need the full body ? we can always switch from Stream to StreamBuffer later in the request pathway (see `merge_body_mode` in `body.rs`) but the StreamBuffer to Stream is not possible.

it seems like there's already precedent for this in `responses_to_chat_completions`. alternative fail-hard guardrails is enabled with SSE instead of forwarding along the request

filters/src/guardrails/filter.rs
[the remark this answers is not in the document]
christinaexyou: is there a reason why integration tests can't be added to this PR ? imo it isn't complete unless you test out the behavior

examples/configs/nemo-guardrails.yaml
```diff
@@ -1,25 +1,36 @@
...
+#     success  - request forwarded to the upstream unchanged
+#     blocked  - 403 returned to the client (triggered rail names in body)
+#     error    - 500 returned to the client
 #
-# Start a local NeMo Guardrails instance before running this config
+#   Response phase:
+#     success  - response forwarded to the client unchanged
+#     blocked  - response body replaced with a JSON error payload
+#                (status remains 200 because headers are already committed)
+#     error    - 500 returned to the client
```
christinaexyou: update this since we don't return a 500 on error

filters/src/guardrails/filter.rs
```diff
@@ -133,6 +130,13 @@ impl HttpFilter for AiGuardrailsFilter {
         Ok(FilterAction::Continue)
     }
 
+    async fn on_response(&self, ctx: &mut HttpFilterContext<'_>) -> Result<FilterAction, FilterError> {
```
christinaexyou, line 130: instead of returning a `FilterError`, would it make more sense to replace the content body with error JSON similar to the `Block`so that the client receives the error message ?

filters/src/guardrails/filter.rs
liavweiss: Thank you, I agree with you.

filters/src/guardrails/filter.rs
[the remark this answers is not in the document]
liavweiss: I missed that we have an integration test for the request, so I added an integration test for the response too. Thank you.

examples/configs/nemo-guardrails.yaml
liavweiss: Since I added an integration test for the response side, I will use nemo-guardrails-response.yaml config.
I have updated the correct error return there.

filters/src/guardrails/filter.rs
liavweiss: Done.

liavweiss:
@christinaexyou, thank you for your review! I’ve addressed all your comments.

@leseb, we’re good to go.

szedan-rh:
Per @christinaexyou and @leseb - PR is ready be merged.
Thank you folks for the review.

szedan-rh (approved):

leseb (changes requested):
A few comments:

* why are we bypassing streaming?
* why don't we handle auth anywhere? do we expect the guardrail endpoint to accept anonymous requests?
* since this is using SubRequestClient (it should), we will need to refactor this code once https://github.com/praxis-proxy/praxis/issues/1072 starts getting shape, now https://github.com/praxis-proxy/ai/issues/961

liavweiss:
> A few comments:
> 
> * why are we bypassing streaming?
> * why don't we handle auth anywhere? do we expect the guardrail endpoint to accept anonymous requests?
> * since this is using SubRequestClient (it should), we will need to refactor this code once [praxis-proxy/praxis#1072](https://github.com/praxis-proxy/praxis/issues/1072) starts getting shape, now #961

1. Guardrails require the full response to evaluate intent, so per-chunk evaluation isn't viable. Buffering would break streaming behavior, so we currently skip SSE responses and evaluate only non-streaming responses. This is a known limitation we plan to revisit once (if) on_response_body becomes async (#51).

2. NeMo Guardrails doesn't require authentication by default - the endpoint is expected to be internal (sidecar / cluster-local).

3. Once #961 lands, we'll refactor the NeMo provider to align with the new SubRequestClient API.

leseb:
> > A few comments:
> > 
> > * why are we bypassing streaming?
> > * why don't we handle auth anywhere? do we expect the guardrail endpoint to accept anonymous requests?
> > * since this is using SubRequestClient (it should), we will need to refactor this code once [praxis-proxy/praxis#1072](https://github.com/praxis-proxy/praxis/issues/1072) starts getting shape, now [feat(callout): route AI guardrail requests through filtered subrequests #961](https://github.com/praxis-proxy/ai/issues/961)
> 
> 1. Guardrails require the full response to evaluate intent, so per-chunk evaluation isn't viable. Buffering would break streaming behavior, so we currently skip SSE responses and evaluate only non-streaming responses. This is a known limitation we plan to revisit once (if) on_response_body becomes async ([Investigate async `on_response_body` support #51](https://github.com/praxis-proxy/ai/issues/51)).

makes sense

> 2. NeMo Guardrails doesn't require authentication by default - the endpoint is expected to be internal (sidecar / cluster-local).

ok 🤷🏻 

> 3. Once [feat(callout): route AI guardrail requests through filtered subrequests #961](https://github.com/praxis-proxy/ai/issues/961) lands, we'll refactor the NeMo provider to align with the new SubRequestClient API.

Thanks! Please comment on it so I assign you.

leseb (approved):

PR #861, opened 2026-09-01
feat(auth): add cache-through token caching for azure_ad and gcp_adc

### LINKED ISSUE #811: Azure AD / Entra ID upstream authentication

szedan-rh:
## Summary

Adds `apis::token_cache::TokenCache`, a reusable cache-through credential
cache: a request that finds the cache stale acquires a fresh value inline,
double-checked under an exclusive lock so concurrent callers racing an
empty cache trigger at most one fetch. There is no background refresh loop
and no server-side retry/backoff; a failed fetch is simply not cached, so
the next request tries again.

- Retrofits `azure_ad` onto it, removing its per-instance background
  refresher thread and private tokio runtime entirely.
- Implements `gcp_adc`'s metadata-server token fetch (GCE/GKE Workload
  Identity) on the same cache, unblocking what was previously a permanent
  503 stub. Service-account key file fetch remains a documented follow-up
  (needs JWT signing, not currently a workspace dependency) — it now fails
  closed with a clear "not implemented" reason instead of silently
  503ing forever.
- `refresh_ratio` is removed from both filters' config: it's meaningless
  under cache-through, since there's no scheduled ahead-of-time refresh.
- `gcp_adc` gains a `metadata_host` config field, mirroring `azure_ad`'s
  `authority_host`, so a non-default or test metadata endpoint can be
  configured directly.

This is the smaller, "make it work" cache-through implementation the
team converged on as a faster path than the full refresh-ahead
credential-store design — see discussion on #811.

## Related issue

Closes #811

(No equivalent tracking issue exists yet for the `gcp_adc` half of this
PR — that spike, praxis#554, lives in the `praxis-proxy/praxis` repo.)

## Validation

- [x] Unit tests — `cargo test -p praxis-ai-apis` and
      `cargo test -p praxis-ai-filters --features azure-ad-filter,gcp-adc-filter`
- [x] Integration or functional tests — `azure_ad`/`gcp_adc` example-config
      suites in `tests/integration`
- [x] `make lint` (workspace + experimental features)
- [x] `make test-unit`

## Checklist

## Breaking changes

`refresh_ratio` is removed from the `azure_ad` and `gcp_adc` filter
config schemas. Any deployment setting it will now fail config
validation (`deny_unknown_fields`) and must remove the field.

### DIFF
--- a/apis/src/token_cache.rs
+++ b/apis/src/token_cache.rs
@@ -21,6 +21,7 @@
 use std::time::{Duration, Instant};
 
 use tokio::sync::RwLock;
+use tracing::warn;
 
 // -----------------------------------------------------------------------------
 // Margin
@@ -29,7 +30,7 @@ use tokio::sync::RwLock;
 /// Safety margin to subtract from a TTL before treating a cached value
 /// as expired, capped at half the TTL so a short-lived credential stays
 /// usable for part of its life instead of being cached already-expired.
-pub(crate) fn effective_margin(ttl: Duration, margin: Duration) -> Duration {
+fn effective_margin(ttl: Duration, margin: Duration) -> Duration {
     margin.min(ttl / 2)
 }
 
@@ -89,6 +90,11 @@ impl<T: Clone + Send + Sync> TokenCache<T> {
     ///
     /// Returns whatever `fetch` returns on failure. Nothing is cached
     /// in that case, so the next call tries again.
+    #[expect(
+        clippy::significant_drop_tightening,
+        reason = "guard's last use is immediately before each return; an explicit drop() would just \
+                   restate that, not shorten the actual hold time"
+    )]
     pub async fn get_or_refresh<F, Fut, E>(&self, fetch: F) -> Result<T, E>
     where
         F: FnOnce() -> Fut + Send,
@@ -105,17 +111,22 @@ impl<T: Clone + Send + Sync> TokenCache<T> {
         let mut guard = self.cache.write().await;
         let fresh = valid_cached_value(guard.as_ref());
         if let Some(value) = fresh {
-            drop(guard);
             return Ok(value);
         }
 
         let (value, ttl) = fetch().await?;
+        if ttl <= self.margin {
+            warn!(
+                ttl_secs = ttl.as_secs(),
+                margin_secs = self.margin.as_secs(),
+                "token TTL is unusually short; validity margin reduced"
+            );
+        }
         let expires_at = Instant::now() + ttl.saturating_sub(effective_margin(ttl, self.margin));
         *guard = Some(Entry {
             value: value.clone(),
             expires_at,
         });
-        drop(guard);
         Ok(value)
     }
 }
--- a/apis/src/token_cache.rs
+++ b/apis/src/token_cache.rs
@@ -1,0 +1,257 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright (c) 2026 Praxis Contributors
+
+//! [`TokenCache`] — a cache-through, on-demand-refreshed cache for a
+//! single short-lived credential (an upstream bearer token, an access
+//! token, ...).
+//!
+//! # Semantics
+//!
+//! There is no background refresh. Every call to
+//! [`TokenCache::get_or_refresh`] checks the cache; if the cached value
+//! is still valid it is returned immediately (a shared read lock, no
+//! network call). If it is missing or past its safety margin, the
+//! caller takes an exclusive lock, checks again in case a concurrent
+//! caller already refreshed it while this one was waiting for the lock,
+//! and only then calls `fetch` — at most once per group of callers that
+//! all observe a stale cache at the same time. A failed fetch is
+//! propagated to the caller and nothing is cached, so the next call
+//! tries again; there is no server-side retry/backoff loop.
+
+use std::time::{Duration, Instant};
+
+use tokio::sync::RwLock;
+
+// -----------------------------------------------------------------------------
+// Margin
+// -----------------------------------------------------------------------------
+
+/// Safety margin to subtract from a TTL before treating a cached value
+/// as expired, capped at half the TTL so a short-lived credential stays
+/// usable for part of its life instead of being cached already-expired.
+pub(crate) fn effective_margin(ttl: Duration, margin: Duration) -> Duration {
+    margin.min(ttl / 2)
+}
+
+// -----------------------------------------------------------------------------
+// TokenCache
+// -----------------------------------------------------------------------------
+
+/// A cached value and the instant after which it must not be used.
+struct Entry<T> {
+    /// The cached value itself.
+    value: T,
+
+    /// Instant, already adjusted by the cache's margin, after which the
+    /// value must not be used.
+    expires_at: Instant,
+}
+
+/// Return a clone of `entry`'s value if it is still valid, `None`
+/// otherwise (missing or past its safety-margin-adjusted expiry).
+fn valid_cached_value<T: Clone>(entry: Option<&Entry<T>>) -> Option<T> {
+    entry
+        .filter(|entry| entry.expires_at > Instant::now())
+        .map(|entry| entry.value.clone())
+}
+
+/// Cache-through cache for one credential. See the module docs.
+pub struct TokenCache<T> {
+    /// Safety window subtracted from a fetched value's TTL (capped at
+    /// half the TTL) before it is treated as expired.
+    margin: Duration,
+
+    /// The cached entry, if any.
+    cache: RwLock<Option<Entry<T>>>,
+}
+
+impl<T: Clone + Send + Sync> TokenCache<T> {
+    /// Build an empty cache. `margin` is the safety window subtracted
+    /// from a fetched value's TTL (capped at half the TTL) before it is
+    /// treated as expired.
+    pub fn new(margin: Duration) -> Self {
+        Self {
+            margin,
+            cache: RwLock::new(None),
+        }
+    }
+
+    /// Return a cached, valid value — fetching a new one first if the
+    /// cache is empty or the cached value is within its safety margin
+    /// of expiry.
+    ///
+    /// At most one concurrent caller ever calls `fetch`: everyone else
+    /// who finds the cache stale queues on the exclusive lock behind
+    /// the first caller and, once it releases the lock, re-checks and
+    /// finds the value that caller just published.
+    ///
+    /// # Errors
+    ///
+    /// Returns whatever `fetch` returns on failure. Nothing is cached
+    /// in that case, so the next call tries again.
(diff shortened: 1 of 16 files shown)


alexsnaps:
Generally fine with the approach. There is the pathological timeouts on network partitions: probably should file an issue about that in order to follow up.

apis/src/token_cache.rs
```diff
@@ -0,0 +1,270 @@
...
+use tokio::sync::RwLock;
+
+// -----------------------------------------------------------------------------
+// Margin
+// -----------------------------------------------------------------------------
+
+/// Safety margin to subtract from a TTL before treating a cached value
+/// as expired, capped at half the TTL so a short-lived credential stays
+/// usable for part of its life instead of being cached already-expired.
+pub(crate) fn effective_margin(ttl: Duration, margin: Duration) -> Duration {
```
alexsnaps: Don't think you need that much visibility on that fn

apis/src/token_cache.rs
```diff
@@ -0,0 +1,270 @@
...
+        // already refreshed it while this one waited for the lock).
+        let mut guard = self.cache.write().await;
+        let fresh = valid_cached_value(guard.as_ref());
+        if let Some(value) = fresh {
+            drop(guard);
+            return Ok(value);
+        }
+
+        let (value, ttl) = fetch().await?;
+        let expires_at = Instant::now() + ttl.saturating_sub(effective_margin(ttl, self.margin));
```
alexsnaps, line 125: Might want to account for unusually short ttls here and log?

apis/src/token_cache.rs
```diff
@@ -0,0 +1,270 @@
...
+        if let Some(value) = fresh {
+            return Ok(value);
+        }
+
+        // Slow path: exclusive lock, re-check (another caller may have
+        // already refreshed it while this one waited for the lock).
+        let mut guard = self.cache.write().await;
+        let fresh = valid_cached_value(guard.as_ref());
+        if let Some(value) = fresh {
+            drop(guard);
```
alexsnaps: Those `drop`s are just visual noise. We don't need them

alexsnaps (commented):
There are some failures in CI too

aslakknutsen:
(ai): The PR checklist claims user-facing and generated documentation were updated, but docs/filters/azure_ad.md still shows refresh_ratio in the example and docs/filters/gcp_adc.md still states token acquisition is not implemented and documents refresh_ratio. cargo xtask lint-filter-docs reports both files stale; make lint includes this check and will fail until cargo xtask generate-filter-docs is run."
(ai): "tests/integration/tests/suite/examples/azure_ad.rs lines 47-48 say authority_host is pointed at a closed port so the background refresher can never acquire a token. The refresher was removed; the comment should say inline cache-through fetch instead to avoid misleading future readers."

aslakknutsen:
The approach look good. I'm wondering if we could iterate on it and possible split it up into two separate times; 
soft expiry (ttl-margin) and hard expiry (ttl).  

* At "soft expiry" we would trigger a refresh, blocking the single request. The rest of the requests could get the "old still valid token" while it's being refreshed. 
* If we hit the "hard expiry" we would block requests waiting for the refresh

And I think we need to address what happens if the the refresh never happens, e.g. the server is down. Not sure we can just "queue" up requests for ever.  At some point we need to start returning some form of error to the client to free up the requests.

szedan-rh:
> The approach look good. I'm wondering if we could iterate on it and possible split it up into two separate times; soft expiry (ttl-margin) and hard expiry (ttl).
> 
> * At "soft expiry" we would trigger a refresh, blocking the single request. The rest of the requests could get the "old still valid token" while it's being refreshed.
> * If we hit the "hard expiry" we would block requests waiting for the refresh
> 
> And I think we need to address what happens if the the refresh never happens, e.g. the server is down. Not sure we can just "queue" up requests for ever. At some point we need to start returning some form of error to the client to free up the requests.

I think soft expiry (ttl-margin) / hard expiry (ttl) is a reasonable idea to explore. On the failure scenario  fair pushback, I was overstating it as "IdP down," and you're right that's rare for GCP/Azure specifically.

The sharper risk isn't IdP availability though, it's request duration: even with a perfectly healthy IdP, a pass-through request that picks up the token right as it crosses into the soft -> hard window could still be mid-flight when hard expiry hits, purely because its own request ran longer than the margin  nothing to do with refresh succeeding or failing. That's the actual reason the margin exists today (never hand out a token that could outlive the request holding it), so pass-through would need an explicit answer for that, not just a faster/rarer refresh-failure path.

I'd rather not fold this into the current PR it adds real synchronization complexity (a second "who's refreshing" signal separate from the read/write lock) on top of what we deliberately scoped down to. If we want to pursue it, I'd want to spec it as its own follow-up with an explicit answer to "is it OK for a request to receive a token that might expire before its own response completes," rather than assume the hard-expiry backstop dissolves that risk.

szedan-rh:
@alexsnaps / @aslakknutsen can you please take another look?

filters/src/gcp/config.rs
```diff
@@ -66,20 +65,21 @@ pub(super) struct GcpAdcConfig {
...
     pub credentials_file: Option<String>,
 
-    /// Fraction of a token's TTL at which to refresh it. Must be in the
-    /// open interval `(0, 1)`.
-    #[serde(default = "default_refresh_ratio")]
-    pub refresh_ratio: f64,
+    /// GCE/GKE metadata server host. Override for testing (point at a
+    /// local mock) or a non-default metadata server; production
+    /// deployments should not need to set this.
+    #[serde(default = "default_metadata_host")]
```
alexsnaps: Document the limitations here? And regen docs?

filters/src/gcp/config.rs
```diff
@@ -98,22 +98,58 @@ pub(super) fn parse_gcp_adc_config(config: &serde_yaml::Value) -> Result<GcpAdcC
...
+/// Reject a `metadata_host` that isn't the real GCE/GKE metadata server or
+/// a loopback address (used only to point tests at a local mock).
+///
+/// The metadata endpoint is safe to reach over plain HTTP specifically
+/// because it is link-local and never routable off the VM/host. Any other
+/// host configured here would send the same plaintext request -- and
+/// receive the access token in the response -- over a real network path.
+fn validate_metadata_host(value: &str) -> Result<(), FilterError> {
+    let host = value.split(':').next().unwrap_or(value);
+    let is_safe = value == "metadata.google.internal" || host == "127.0.0.1" || host == "localhost";
```
alexsnaps: Are we fine with `localhost`? That can be changed to point to anything... Not a blocker for me. But wanted to highlight it nonetheless

szedan-rh: Agree with you will remove it and rely on the ip

filters/src/gcp/config.rs
szedan-rh: Done

aslakknutsen:
> The sharper risk isn't IdP availability though, it's request duration: even with a perfectly healthy IdP, a pass-through request that picks up the token right as it crosses into the soft -> hard window could still be mid-flight when hard expiry hits, purely because its own request ran longer than the margin nothing to do with refresh succeeding or failing. That's the actual reason the margin exists today (never hand out a token that could outlive the request holding it), so pass-through would need an explicit answer for that, not just a faster/rarer refresh-failure path.

The refresh tokens doesn't have TTL for seconds or minutes right, we're talking closer to 1h+? 

While not perfect, we could make the margin for soft expiry something like 5m and also use a margin for hard expiry at e.g. 1m. It creates over time a few more refresh calls, but also gives plenty of time to try to refresh even with a few retries and even pre hard expiry requests have a certain time buffer to complete before it's invalid.

szedan-rh:
> > The sharper risk isn't IdP availability though, it's request duration: even with a perfectly healthy IdP, a pass-through request that picks up the token right as it crosses into the soft -> hard window could still be mid-flight when hard expiry hits, purely because its own request ran longer than the margin nothing to do with refresh succeeding or failing. That's the actual reason the margin exists today (never hand out a token that could outlive the request holding it), so pass-through would need an explicit answer for that, not just a faster/rarer refresh-failure path.
> 
> The refresh tokens doesn't have TTL for seconds or minutes right, we're talking closer to 1h+?
> 
> While not perfect, we could make the margin for soft expiry something like 5m and also use a margin for hard expiry at e.g. 1m. It creates over time a few more refresh calls, but also gives plenty of time to try to refresh even with a few retries and even pre hard expiry requests have a certain time buffer to complete before it's invalid.

Fair point on the TTL scale  with hour-long tokens, your numbers hold up: pass-through requests would still have ≥1m of guaranteed validity, more conservative than the 30s margin we already ship today.
I'd keep this out of the current PR though  worth its own follow-up with configurable margins (request durations vary a lot across our use cases, so I wouldn't hardcode 5m/1m). I'll fold it into ai#864 so it doesn't get lost.

filters/src/azure/azure_ad.rs
```diff
@@ -417,29 +174,41 @@ async fn refresh_loop(params: RefresherParams, shutdown: CancellationToken) {
...
-    /// Lock-free cache of the current token, populated by the background
-    /// refresher and read on every request.
-    token: Arc<ArcSwap<Option<CachedToken>>>,
+    /// Cache-through token cache; see the module docs and
+    /// [`praxis_ai_apis::token_cache`].
+    cache: TokenCache<HeaderValue>,
+
+    /// HTTP client used for token-endpoint requests, built once with
+    /// [`TOKEN_REQUEST_TIMEOUT`].
+    client: reqwest::Client,
```
aslakknutsen, line 183: I wonder if this should be using the SubRequest from Core over a new reqwest Client. Minimally it should be using Pingora's native client instead (but we might want to do the move from reqwest over to Pingora's native client as a single PR as there are more uses in AI. Core has already swapped)

szedan-rh: you're right that Core's already there — this predates the PR though (was already in azure_ad since it was first added). SubRequestClient isn't wired into azure_ad/gcp_adc's registration path today (unlike web_search/http_callout, which already take a client param), so this isn't a local swap — it needs new registration wiring, plus a couple of the other reqwest sites in apis rely on DNS-pinning for SSRF protection that Pingora's client would need an equivalent for. Agree it's worth doing as its own PR across all the reqwest sites rather than one-off here. I'll file a follow-up to track it.

aslakknutsen (approved):

PR #877, opened 2026-09-02
fix(conversations): compare-and-swap the denormalized message cache

### LINKED ISSUE #662: Concurrent item appends can overwrite the denormalized conversation history
## Clawpatch finding
- **Severity:** high
- **Category:** concurrency
- **Confidence:** high
- **Feature(s):** feat_custom_apis_openai_core_protocol
- **Finding ID(s):** fnd_sig-feat-custom-apis-openai-core_d420002288
- **Run reports:** 20260810T070320-ef1253, 20260810T070811-31e99e, 20260810T071144-b067a0, 20260810T071621-ff9ceb, 20260810T071954-8437e6
- **Prior-run overlap:** No direct prior-run overlap detected; related to the broader conversation-history work tracked in #532 but this root cause is concurrent write lost-update behavior.

### Concurrent item appends can overwrite the denormalized conversation history
- `apis/src/openai/conversations/handlers.rs:277-310 (`handle_create_items`)`
- `apis/src/openai/conversations/filter.rs:573-593 (`persist_items`)`
- `apis/src/openai/conversations/handlers.rs:928-944 (`sync_conversation_messages`)`

### Concurrent item appends can overwrite the denormalized conversation history
Both explicit item creation and response append-back calculate the next position with a separate `MAX(position)` read before inserting. Concurrent writers can therefore assign overlapping positions. More critically, each writer rebuilds the denormalized `messages` value by reading a snapshot and then issuing a separate update. A slower writer can commit an older snapshot after a newer writer, removing the newer items from the conversation record even though their item rows still exist. Consumers that use the denormalized conversation messages then observe lost history.

### Concurrent item appends can overwrite the denormalized conversation history
Run two concurrent item-create operations for the same conversation and place barriers so both obtain the same maximum position. Allow writer A to collect its message snapshot, then let writer B insert, collect and update the cache, followed by A's stale cache update. Item rows contain both writes, while the conversation `messages` field omits B's items.

### Concurrent item appends can overwrite the denormalized conversation history
Move position allocation, item insertion and denormalized-message maintenance into one store-level transaction with appropriate per-conversation serialization or locking. Prefer incrementally updating or eliminating the denormalized cache rather than performing an unlocked read-all/write-all cycle.

---
Reported from scheduled Clawpatch Codex/xhigh review after rebasing `/Users/leseb/Documents/AI/ai` to `origin/main` (`1cc0a47`). Please validate before assigning or starting implementation.

leseb:
## Summary

Concurrent conversation item appends each rebuilt the denormalized `messages` cache from the authoritative item rows and wrote it back unconditionally, so a slower writer could overwrite a newer writer's cache with a stale snapshot and drop items from the history that rehydration consumes. This replaces the blind write with a compare-and-swap that retries a lost race with a freshly read snapshot, and reuses the create/delete caller's existing pre-mutation snapshot to skip a redundant cache reload; because the written value is always a fresh rebuild from item rows, a stale snapshot only ever loses the swap and never persists stale data. Exhausting the bounded retries logs and returns `Ok` since the row mutation is already durably committed and any later write rebuilds the cache, while genuine store errors and a disappeared conversation still propagate.

## Related issue

Closes #662

## Validation

- [x] Unit tests — `cargo test -p praxis-ai-apis` (2592 passed, 0 failed; six #662 cache-sync tests covering stale-cache repair, concurrent-clobber, snapshot fast-path with no reload, stale-snapshot fresh-read fallback, missing-conversation error, and retry exhaustion)
- [ ] Integration or functional tests — n/a; the concurrency is exercised deterministically via a fault-injecting store wrapper over the real SQLite store, and there is no new config surface to add a functional example test for
- [x] `make lint` and `make build`

## Checklist

- [ ] New capabilities include an example config and functional example test. — n/a; bug fix, no new capability or config surface
- [ ] User-facing behavior and generated documentation are updated. — n/a; internal cache-consistency behavior only
- [ ] Performance-sensitive changes include appropriate benchmark or load-test evidence. — n/a; correctness fix (bounded CAS retries only on contention)

## Breaking changes

None. `sync_conversation_messages` is a crate-internal (`pub(super)`) helper; the signature change is not part of any public API.

### DIFF
--- a/apis/src/openai/conversations/filter.rs
+++ b/apis/src/openai/conversations/filter.rs
@@ -26,7 +26,7 @@ use super::{
 };
 use crate::{
     openai::responses::{DEFAULT_TENANT_ID, TENANT_METADATA_KEY, state::ResponsesState},
-    store::{ConversationItemStore, ConversationRecord, PostgresResponseStore, SqliteResponseStore, StoreError},
+    store::{ConversationItemStore, PostgresResponseStore, SqliteResponseStore, StoreError},
 };
 
 // -----------------------------------------------------------------------------
@@ -601,14 +601,9 @@ async fn persist_items(
 
 /// Refresh the denormalized conversation message cache after item mutation.
 async fn refresh_message_cache(store: &dyn ConversationItemStore, tenant_id: &str, conversation_id: &str) {
-    let record = ConversationRecord {
-        conversation_id: conversation_id.to_owned(),
-        tenant_id: tenant_id.to_owned(),
-        created_at: 0,
-        metadata: Value::Object(serde_json::Map::default()),
-        messages: Value::Null,
-    };
-    if let Err(e) = handlers::sync_conversation_messages(store, record).await {
+    // The append-back path holds no pre-mutation cache snapshot, so it syncs with
+    // `None`; the refresh then re-reads the live cache as the swap's expected value.
+    if let Err(e) = handlers::sync_conversation_messages(store, tenant_id, conversation_id, None).await {
         warn!(error = %e, conversation_id, "conversation message sync failed after append-back");
     }
 }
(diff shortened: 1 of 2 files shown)


franciscojavierarceo (approved):
reviewed the current head, all available comments and review threads, the architectural paths, and the completed checks. i found no actionable issues. approving this commit.

Issue #842, opened 2026-08-31
refactor(streaming): adopt the shared Praxis SSE codec


leseb:
## Description

After praxis-proxy/praxis#986 provides the shared bounded SSE codec, migrate
Praxis AI's provider-neutral SSE framing code to that `praxis-filter` API.

Praxis AI currently has several independently evolved framing paths:

- `apis/src/openai/sse/frame.rs`
- `apis/src/openai/sse/responses/parser.rs`
- `apis/src/openai/responses/stream_events/`
- `apis/src/anthropic/stream_events/`
- `filters/src/agentic/a2a/sse.rs`
- the A2A parser reuse in `filters/src/token_usage/count.rs`

The refactor should share only the transport-independent SSE record framing.
Provider event schemas, JSON decoding, lifecycle validation, accumulation,
translation, failure policy, response IDs, and sequence handling remain owned
by their existing AI modules.

## Acceptance Criteria

- Update the Praxis dependency to a revision containing
  praxis-proxy/praxis#986.
- Replace `apis/src/openai/sse/frame.rs` with the shared incremental decoder and
  encoder, removing the local implementation once all callers are migrated.
- Migrate the Responses, Anthropic, A2A, and token-usage consumers where their
  framing requirements match the shared contract.
- Do not introduce a common provider event model; each consumer retains its
  typed event parsing and lifecycle rules above the shared SSE records.
- Preserve each consumer's existing pass-through, transformation, fail-open or
  fail-closed, terminal-event, and end-of-stream behavior.
- Preserve raw event bytes where a pass-through consumer requires byte-exact
  forwarding. If the shared codec cannot support a required behavior, refine
  praxis-proxy/praxis#986 rather than adding another AI-local framing parser.
- Retain consumer-specific limits such as event count, accumulated state, JSON
  size, and lifecycle timeouts; the shared framing bounds do not replace them.
- Tests cover arbitrary chunk splits, UTF-8 splits, LF/CRLF, comments,
  multiline data, multiple events per chunk, incomplete EOF, malformed input,
  and framing-limit overflow through the migrated production callers.
- No migrated streaming path buffers the complete response or weakens
  downstream backpressure.
- Remove obsolete duplicate parser tests and replace them with focused adapter
  tests plus the shared codec's conformance coverage.

## Non-Goals

- Moving OpenAI `ResponsesState`, Anthropic translation, A2A task semantics, or
  token accounting into Praxis.
[...]

PR #942, opened 2026-09-07
refactor(streaming): adopt shared Praxis SSE codec for Responses framing

leseb:
## What

Migrate the OpenAI Responses provider-neutral SSE **record framing** onto the shared `praxis_filter::sse` codec introduced by praxis-proxy/praxis#1060 (epic praxis#985). `SseFrameParser` now delegates byte-level line/record framing to `SseDecoder` while keeping the OpenAI-facing `SseFrame` / `SseParseError` surface, the `[DONE]` sentinel, the event-count budget, timeouts, and lifecycle rules in the Responses consumers unchanged.

Praxis has no release carrying the codec yet, so the praxis core crates are pinned to `main` at the #1060 merge commit (`c4cab79`) and Pingora moves to the `0.8.4` fork in lockstep with core.

## Scope — migrate only where framing matches the shared contract

- **OpenAI Responses** (`openai_stream_events`, `responses::parser`) → migrated onto the codec. `parse_chunk` now takes `&Bytes` and threads it into the decoder without copying.
- **A2A route-capture scanner + the token-usage counter that reuses it** → intentionally **not** migrated. They are fail-open and *recoverable* on an oversized record (drop it, resync at the next blank-line boundary), whereas the shared decoder poisons permanently. Adopting the codec there requires a recoverable/resync-on-overflow mode in praxis#986, not another AI-local parser. The gap is documented in the A2A `sse` module and referenced from the token-usage consumer.
- **Anthropic `anthropic_stream_events`** → deferred to a follow-up PR.

## Behavior preservation

`max_buffer_bytes` maps onto both codec framing bounds (`max_line_bytes` and `max_record_bytes`). For the single-`data:`-line blocks the Responses API emits, the overflow trip point and peak retained bytes match the pre-codec parser; the `new()` doc describes the mapping truthfully (including how a multi-line record — which the Responses API never emits — is accounted differently). No full-body buffering is introduced and backpressure is not weakened for real traffic.

Duplicate pure-codec framing tests are removed in favor of adapter-focused tests plus the codec's own conformance suite.

## Testing

- `make lint` ✓
- `make test` ✓
- `make doc` ✓ (verifies the new intra-doc links)

Refs #842

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -147,46 +147,18 @@ checksum = "d3fb67a6e08acf24fdeccbac2cb6ac4305825bd1f117462e0e6f2f193345ad56"
 
 [[package]]
 name = "asn1-rs"
-version = "0.6.2"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "5493c3bedbacf7fd7382c6346bbd66687d12bbaad3a89a2d2c303ee6cf20b048"
-dependencies = [
- "asn1-rs-derive 0.5.1",
- "asn1-rs-impl",
- "displaydoc",
- "nom",
- "num-traits",
- "rusticata-macros",
- "thiserror 1.0.69",
- "time",
-]
-
-[[package]]
-name = "asn1-rs"
 version = "0.7.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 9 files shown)


leseb:
Draft since we need a new core release with the SSE codec code.

leseb:
will revisit once the core impl fixes the remaining whatwg spec issue - tracked in https://github.com/praxis-proxy/praxis/issues/985

PR #977, opened 2026-09-07
fix: fail closed on malformed compaction translation

### LINKED ISSUE #936: Malformed compaction items are silently dropped during Responses-to-Chat translation
## Clawpatch finding

- **Severity:** high
- **Category:** data-loss
- **Confidence:** high
- **Triage:** confirmed-bug
- **Finding ID:** `fnd_sig-feat-custom-apis-openai-core_fcd232c9f7`
- **Run:** `20260907T052823-e5e76e`
- **Prior-run overlap:** Narrow follow-up to closed #552 / merged #868. That fix validates other malformed input variants but leaves compaction decoding fail-open.

## Evidence

- `apis/src/openai/translation/chat_completions.rs:512-521`
- `apis/src/openai/translation/chat_completions.rs:553-566`
- `apis/src/openai/translation/mod.rs:738-774`

## Description

Unlike other typed input items, the compaction branch cannot return a TranslationError. Missing encrypted_content, a non-string value, invalid base64, or decoded non-UTF-8 data all become an empty string through chained Option conversions and unwrap_or_default. The item is then omitted while translation succeeds, so the backend receives the current request without the compacted prior context and may return a semantically incorrect response. This validation cannot be delegated to the backend because the malformed item has already been removed from the translated request.

## Reproduction

Call responses_request_to_chat_request with {"model":"m","input":[{"type":"compaction","id":"compact_1","encrypted_content":"%%%not-base64%%%"},{"role":"user","content":"What did we decide?"}]}. Translation succeeds and produces only the user message instead of returning an error.

## Expected behavior

The proxy must fail closed or preserve the documented API and filter contract without silently changing successful output, permissions, durability, or resource limits.

## Recommendation

Make append_compaction_item return Result<(), TranslationError>. Require encrypted_content to be a string containing valid base64 and valid UTF-8, and propagate a translation error for malformed or missing content. Only intentionally empty, valid decoded summaries should be omitted if that behavior is desired.

## Validation

Source inspection confirms missing, non-string, invalid-base64, and non-UTF-8 `encrypted_content` all collapse to an empty string and the item is omitted after the original Responses input has been consumed.

---
Generated from scheduled Clawpatch Codex `gpt-5.6-sol` / `xhigh` review runs `20260907T052439-47e401, 20260907T052823-e5e76e, 20260907T053119-d89f7d, 20260907T053325-beffe2, 20260907T053712-f49e16`. Validate before assigning or implementing.

mkoushni:
## Summary

Responses-to-Chat translation treated missing, non-string, invalid-base64, or non-UTF-8 compaction `encrypted_content` as an empty string and dropped the item. Translation still succeeded, so Chat Completions backends received the current turn without compacted history.

`append_compaction_item` now returns `TranslationError` for those cases. Valid empty decoded summaries are still omitted.

## Related issue

Closes #936

## Validation

- [x] Unit tests
- [x] `make lint`

Covered by Responses-to-Chat translation unit tests, including the issue reproduction (`%%%not-base64%%%`). `make coverage-check` passed at 96.06% line coverage. No new example config is required; this is fail-closed validation on an existing input item type.

## Checklist

## Breaking changes

Malformed compaction items that previously translated successfully (with compacted history silently removed) now fail translation.


leseb (changes requested):
missing fixture coverage

mkoushni:
> missing fixture coverage

done

leseb (approved):

PR #980, opened 2026-09-07
feat(token-rate-limit): key quotas by authenticated subject

nerdalert:
## 🔴  Dependency and merge readiness

Praxis PR praxis-proxy/praxis#1108 has merged into Praxis `main`. This AI PR is therefore ready to leave draft and receive review.

It can merge after all of the following are complete:

1. Praxis publishes a release containing praxis-proxy/praxis#1108.
2. AI updates its Praxis dependency and lockfile from `0.5.4` to that released version.
3. A cross-filter integration test proves that Basic Auth establishes `AuthenticatedIdentity` and `key: authenticated_subject` consumes it successfully without any local path or `[patch.crates-io]` override.

Merging the code before that release would not change existing quota behavior because `key: global` remains the default. However, Basic Auth with `key: authenticated_subject` would fail closed with HTTP 401 while AI still resolves Praxis `0.5.4`. The dependency bump and integration test are therefore required before merge, not merely before publication.

## Summary

Adds an opt-in `authenticated_subject` key to the token-rate-limit filter. When selected, each verified application or user receives an independent quota while gateway replicas share that subject's budget through the existing Valkey backend.

The default remains `global`, preserving the current behavior and configuration compatibility.

## Cross-repository contract

This complements the broader token-rate-limiting work tracked in #121. It implements one secure bucket-key source; it does not close the epic.

Praxis PR praxis-proxy/praxis#1108 is merged. It makes Basic Auth publish its verified username through the private request-local `AuthenticatedIdentity` extension. This filter consumes that same authentication-neutral type. Policy/JWT authentication already uses the type, allowing the quota key to be reused by JWT/OIDC/OAuth-backed authentication without coupling token quota to Basic Auth.

The important boundary is that AI never derives quota identity from a caller-controlled header.

## Behavior

- Adds `key: authenticated_subject` at the token-rate-limit filter level.
- Keeps `key: global` as the default.
- Reads only the trusted request-local `AuthenticatedIdentity` subject.
- Fails closed with HTTP 401 when authenticated-subject keying is configured but no verified identity exists.
- Hashes the subject with SHA-256 and URL-safe base64 before using it in memory, Valkey keys, metadata, or metrics.
- Uses the same resolved opaque key for reservation and settlement.
- Retains existing rule matching, quota algorithms, reservation bounds, and backend behavior.

## Use case

Three applications can authenticate through the same endpoint and share the same quota rule configuration while receiving separate subject-keyed budgets. Requests for one application share quota state across gateway replicas, but cannot consume another application's capacity. Grid provider selection remains independent and occurs only after quota admission.

## Tests and validation

Focused coverage includes:

- backward-compatible global default;
- parsing and rejection of key-source values;
- stable, distinct, opaque subject keys;
- fail-closed behavior without identity;
- independent subjects under one rule;
- shared subject state across filter instances;
- reservation and settlement using the same key.

The complete Grid qualification passed twice, 9/9 scenarios per run, using this code with the merged Praxis identity producer. The committed branch contains no `[patch.crates-io]`, sibling path dependency, generated evidence, or local build compatibility changes.

Related to #121 and praxis-proxy/grid#101.
Depends on praxis-proxy/praxis#1108.

Companion qualification: praxis-proxy/grid#127.
Companion demo: praxis-proxy/demos#20.


leseb (changes requested):
need a new praxis core version

filters/src/token_rate_limit/tests.rs
```diff
@@ -54,6 +54,42 @@ fn from_config_parses_valid_config() {
...
+    let yaml = single_rule_yaml_with(
+        "key: authenticated_subject",
+        "algorithm: sliding_window\nwindow: 1h\ncapacity: 100\nreserved_tokens: 5",
+    );
+    let filter = TokenRateLimitFilter::from_config(&yaml).unwrap();
+    let req = crate::test_utils::make_request(http::Method::POST, "/v1/chat");
+    let mut ctx = crate::test_utils::make_filter_context(&req);
+
+    let action = filter.on_request(&mut ctx).await.unwrap();
+
```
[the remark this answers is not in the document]
nerdalert: Agreed on the coverage requirement. Praxis #1108 is now merged, but no release containing it has been published yet; AI still resolves Praxis 0.5.4, whose Basic Auth does not publish AuthenticatedIdentity. The suggested direct construction is intentionally unavailable because AuthenticatedIdentity::new is crate-private, preserving the trusted-producer boundary. I am leaving this thread open: when the Praxis release is published, this PR will bump the dependency and add the real Basic Auth -> AuthenticatedIdentity -> token_rate_limit positive integration test without a path or patch override before merge.

nerdalert:
> need a new praxis core version

@leseb  Thanks, sorry I should have mentioned it needs https://github.com/praxis-proxy/praxis/pull/1108 That PR has merged, but am blocked until we cut a release containing it before I can update this PR. Ty!

filters/src/token_rate_limit/mod.rs
```diff
@@ -774,7 +786,16 @@ impl HttpFilter for TokenRateLimitFilter {
...
-        let key = FALLBACK_KEY.to_owned();
+        let Some(key) = self.resolve_key(ctx) else {
+            counter!(
+                "praxis_ai_token_rate_limit_requests_total",
+                "decision" => "denied",
+                "reason" => "missing_authenticated_subject",
+                "rule" => rule.name.clone(),
+            )
+            .increment(1);
+            return Ok(FilterAction::Reject(Rejection::status(401)));
```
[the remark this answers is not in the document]
nerdalert: Fixed in commit ad4f87d3. The missing-authenticated-subject 401 path now emits structured info-level tracing with the rule name, matching the existing 429 and 503 rejection observability. Targeted token-rate-limit tests pass.

jordigilh:
The existing integration tests cover generic token-rate-limit admission, reconciliation, and shared Valkey behavior, but they do not exercise the new authenticated-subject path.

Once the Praxis dependency is available, could we add a cross-filter integration test covering:

1. Valid Basic Auth establishes `AuthenticatedIdentity`, and the quota filter admits the request under that subject.
2. Missing or invalid credentials fail before provider contact; no anonymous/global fallback is used.
3. Repeated requests from the same subject consume one shared quota across gateway replicas.
4. A different authenticated subject receives an independent quota.
5. Caller-supplied identity headers cannot override the trusted subject.
6. An exhausted subject is rejected with 429 before routing/provider contact.
7. Settlement updates the same subject bucket that was reserved during admission.

The key missing evidence is the positive authenticated path and cross-replica subject isolation. Current unit coverage only verifies parsing, hashing, and fail-closed behavior when no identity is present.

nerdalert:
> The existing integration tests cover generic token-rate-limit admission, reconciliation, and shared Valkey behavior, but they do not exercise the new authenticated-subject path.
> 
> Once the Praxis dependency is available, could we add a cross-filter integration test covering:
> 
> 1. Valid Basic Auth establishes `AuthenticatedIdentity`, and the quota filter admits the request under that subject.
> 2. Missing or invalid credentials fail before provider contact; no anonymous/global fallback is used.
> 3. Repeated requests from the same subject consume one shared quota across gateway replicas.
> 4. A different authenticated subject receives an independent quota.
> 5. Caller-supplied identity headers cannot override the trusted subject.
> 6. An exhausted subject is rejected with 429 before routing/provider contact.
> 7. Settlement updates the same subject bucket that was reserved during admission.
> 
> The key missing evidence is the positive authenticated path and cross-replica subject isolation. Current unit coverage only verifies parsing, hashing, and fail-closed behavior when no identity is present.

@jordigilh Thanks, addressed!

- Added a feature-gated cross-filter integration test using two gateway replicas and shared Valkey state.
- Covered valid/missing/invalid Basic Auth, trusted subject quota isolation, spoofed-header rejection, 429 before provider contact, settlement, and global-key default behavior.

nerdalert:
Addressed the GitHub Advanced Security findings by removing embedded Basic Auth password literals from the integration fixture. Credentials are now generated deterministically at runtime; test behavior and diagnostics remain unchanged.

tests/integration/tests/suite/examples/token_rate_limit.rs
```diff
@@ -396,3 +400,269 @@
...
+    assert_eq!(parse_status(&invalid), 401, "invalid Basic Auth must fail closed");
+    assert_eq!(
+        backend_two.requests().len(),
+        0,
+        "invalid credentials must not contact the provider"
+    );
+
+    let subject_a_first = http_send(
+        proxy_one.addr(),
+        &basic_auth_json_post("/v1/chat/completions", "{}", "subject-a", "password-a", &[]),
```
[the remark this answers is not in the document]
nerdalert: Replaced all hard-coded Basic Auth test password literals with deterministic runtime-generated credentials.

nerdalert:
Addressed the additional GHAS findings by deriving the test credentials only from runtime identifiers; no hard-coded password literals remain. The focused integration suite still passes formatting, diff, and compilation validation.

tests/integration/Cargo.toml
```diff
@@ -18,6 +18,9 @@ azure-ad-filter = ["praxis-ai-proxy/azure-ad-filter"]
 # Experimental: gates the gcp_adc example test, which needs the gcp_adc
 # filter registered in the in-process proxy registry.
 gcp-adc-filter = ["praxis-ai-proxy/gcp-adc-filter"]
+# Experimental: gates authenticated-subject token-rate-limit tests, which need
+# Praxis's released Basic Auth filter registered in the in-process registry.
+basic-auth-filter = ["praxis-ai-proxy/basic-auth-filter"]
```
jordigilh, line 23: [P1] This new `basic-auth-filter` feature is defined in this PR, but `Makefile:15,89-91` does not include it in `EXPERIMENTAL_FEATURES`. As a result, `make test-integration` compiles out the `#[cfg(feature = "basic-auth-filter")]` cross-filter tests; the dedicated target at `Makefile:107-109` also enables only `token-rate-limit-filter`. Please enable both features (or add a target) so green CI executes the required Basic Auth coverage.

tests/integration/tests/suite/examples/token_rate_limit.rs
```diff
@@ -396,3 +400,292 @@ fn mixed_algorithm_rules_valkey_backend_isolates_budgets_across_gateway_replicas
...
+         \x20 - name: main\n\
+         \x20   filters:\n\
+         \x20     - filter: basic_auth\n\
+         \x20       strip_authorization: true\n\
+         \x20       credentials:\n\
+         \x20         - username: subject-a\n\
+         \x20           password: {subject_a_credential}\n\
+         \x20         - username: subject-b\n\
+         \x20           password: {subject_b_credential}\n\
+         \x20     - filter: router\n\
```
jordigilh: [P2] The test claims an exhausted subject is rejected "before routing," but `router` runs before `token_rate_limit` and assigns `ctx.cluster` before quota rejection. A dedicated probe confirmed: `router -> probe -> token_rate_limit` observed `Some("backend")` for both admitted and exhausted requests, while `token_rate_limit -> router -> probe` observed it only for the admitted request. Both orderings contacted the provider exactly once, so provider contact is prevented. Please move quota before router for the literal requirement, or change the assertion to "before provider contact."

tests/integration/tests/suite/examples/token_rate_limit.rs
```diff
@@ -396,3 +400,292 @@ fn mixed_algorithm_rules_valkey_backend_isolates_budgets_across_gateway_replicas
...
+    );
+    assert!(
+        !backend_one.requests()[0]
+            .headers
+            .to_ascii_lowercase()
+            .contains("authorization:"),
+        "Basic Auth must be stripped before provider contact"
+    );
+
+    let subject_a_second = http_send(
```
jordigilh: [P1] This assertion depends on asynchronous Valkey settlement completing before the next request. `token_rate_limit` enqueues reconciliation onto a background worker rather than awaiting it (`filters/src/token_rate_limit/mod.rs:689-694`), but `subject_a_second` is sent immediately after `subject_a_first` and expects `200`. Because the first request reserves 20 tokens and settles at 10, the second request can race the refund and transiently receive `429`. Please poll for admission, like the existing `poll_until_admitted` helper, or wait on a deterministic settlement signal.

leseb (approved):
https://github.com/praxis-proxy/ai/pull/1188 merged so this is ready - just need a rebase and conflict resolution

tests/integration/Cargo.toml
nerdalert: Addressed in `9444bcaf`. `basic-auth-filter` is now enabled in `EXPERIMENTAL_FEATURES`, the explicit all-feature lint path, and the dedicated Valkey integration target, so the cross-filter coverage is compiled and executed by the normal targets.

nerdalert:
Rebased onto current `main`. The branch now resolves Praxis core `0.5.5` (and its compatible Pingora `0.9.0` dependency set), so the released core `AuthenticatedIdentity` producer required by `key: authenticated_subject` is present without a local path override or `[patch.crates-io]` entry.

tests/integration/tests/suite/examples/token_rate_limit.rs
nerdalert: Addressed in `9444bcaf`. The test pipeline now places `token_rate_limit` before `router`, so an exhausted subject is rejected before route selection as well as before provider contact.

tests/integration/tests/suite/examples/token_rate_limit.rs
nerdalert: Addressed in `9444bcaf`. The second-replica admission now polls within a bounded window until the asynchronous Valkey settlement refund is visible, removing the race without adding an unconditional sleep.

filters/src/token_rate_limit/mod.rs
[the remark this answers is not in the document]
nerdalert: Addressed in `9444bcaf`. The structured 401 trace is restored through a shared helper used by both header-only and body-dependent admission paths.

PR #999, opened 2026-09-08
fix(validate): reject conflicting history selectors

### LINKED ISSUE #840: Responses rehydration silently chooses previous_response_id when conversation is also set
## Clawpatch finding

- **Severity:** medium
- **Category:** api-contract
- **Confidence:** high
- **Feature:** `feat_custom_apis_openai_responses_state_store_stream`
- **Finding ID:** `fnd_sig-feat-custom-apis-openai-resp_79525b1400`
- **Run ID:** `20260831T084818-f6ae5e`
- **Reviewer model:** `gpt-5.6-sol` via the local Codex provider at `xhigh`
- **Prior overlap:** No matching GitHub issue or prior-run finding was found. The OpenAI Responses contract states that `previous_response_id` cannot be used with `conversation`.

## Evidence

- `apis/src/openai/responses/rehydrate/mod.rs:97-115` (`RehydrateFilter::rehydrate`)
- `apis/src/openai/responses/rehydrate/tests.rs:1424-1473` (`previous_response_id_takes_precedence_over_conversation`)

## Description

The current Open Responses request schema states that `previous_response_id` cannot be used with `conversation`. Instead of preserving that invalid combination for backend validation or rejecting it because the proxy must choose a local state source, rehydration silently selects the previous response. This changes an invalid client request into locally defined behavior and can combine history from a response with a different requested conversation.

## Reproduction

Store `resp_win` and `conv_lose` with different histories, then POST `/v1/responses` with both `previous_response_id":"resp_win"` and `conversation":"conv_lose"`. The filter releases the request with state populated from `resp_win` rather than returning an invalid-request error.

## Recommendation

When both non-null selectors are present, return an OpenAI-compatible 400 rejection before fetching either record. Alternatively, leave both completely untouched and avoid local rehydration so the provider can enforce the contract, but do not apply precedence.

---
Reported from a Clawpatch review of `apis/src` at commit `543259683b9ff9a3f42772703a6d9ca134948475`.

derekhiggins:
Reject Responses API requests that provide both previous_response_id and conversation instead of silently prioritizing one history source. Return the OpenAI-compatible invalid-request JSON error for both streaming and non-streaming requests.

Fixes: #840

### DIFF
--- a/apis/src/openai/responses/error.rs
+++ b/apis/src/openai/responses/error.rs
@@ -17,11 +17,16 @@ use praxis_filter::Rejection;
 ///
 /// Produces `{"error":{"message":"<msg>","type":"<code>","param":null,"code":"<code>"}}`.
 pub(crate) fn responses_error_body(code: &str, message: &str) -> Bytes {
+    responses_error_body_with_code(code, code, message)
+}
+
+/// Build a non-streaming OpenAI API error JSON body with distinct type and code.
+fn responses_error_body_with_code(error_type: &str, code: &str, message: &str) -> Bytes {
     Bytes::from(
         serde_json::json!({
             "error": {
                 "message": message,
-                "type": code,
+                "type": error_type,
                 "param": null,
                 "code": code,
             },
@@ -77,9 +82,19 @@ fn responses_error_sse_payload_at_sequence(code: &str, message: &str, sequence_n
 /// emitted by the `stream_events` filter, not here. Every rejection therefore
 /// uses `application/json`.
 pub(crate) fn responses_error_rejection(status: u16, code: &str, message: &str) -> Rejection {
+    responses_error_rejection_with_code(status, code, code, message)
+}
+
+/// Build a [`Rejection`] with distinct OpenAI error type and code values.
+pub(crate) fn responses_error_rejection_with_code(
+    status: u16,
+    error_type: &str,
+    code: &str,
+    message: &str,
+) -> Rejection {
     Rejection::status(status)
         .with_header("content-type", "application/json")
-        .with_body(responses_error_body(code, message))
+        .with_body(responses_error_body_with_code(error_type, code, message))
 }
 
 // -----------------------------------------------------------------------------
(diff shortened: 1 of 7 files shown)


apis/src/openai/responses/validate/mod.rs
```diff
@@ -163,6 +162,21 @@ fn parse_request_body(ctx: &HttpFilterContext<'_>, body: &Option<Bytes>) -> Resu
...
+/// Reject requests that select both supported sources of conversation history.
+fn reject_conflicting_history_selectors(ctx: &HttpFilterContext<'_>, body: &serde_json::Value) -> Option<FilterAction> {
+    let conflicts = body.get("previous_response_id").is_some_and(|value| !value.is_null())
+        && body.get("conversation").is_some_and(|value| !value.is_null());
+    conflicts.then(|| {
+        let streaming = ctx
+            .get_metadata("openai_responses_format.stream")
+            .is_some_and(|value| value == "true");
+        reject_invalid(
+            "Mutually exclusive parameters. Ensure you are only providing one of: 'previous_response_id' or 'conversation'.",
```
leseb, line 174: we should return something similar to openai i guess:

{
  "message": "Mutually exclusive parameters: ''. Ensure you are only providing one of: 'pre..._id' or 'conversation'.",
  "type": "invalid_request_error",
  "param": null,
  "code": "mutually_exclusive_parameters"
}

i don't mind about the message but the code is incorrect here

leseb (changes requested):

leseb:
@derekhiggins also please add a test in test_openai_responses_vllm.py to validate the error shape with the client, thanks (streaming and buffered mode).

apis/src/openai/responses/validate/mod.rs
derekhiggins: I can fix this, but I think we have the same (pre existing problem) on many error responses, 
when developing this all of the errors messages I looked at used `responses_error_body` (or `responses_error_sse_payload`), this takes a single `code` param and then uses it for both `code` and `type` in the error response.

https://github.com/praxis-proxy/ai/blob/d5145ff64b0306f30473401c2f8cc8b8a2a83431/apis/src/openai/responses/error.rs#L16-L24

leseb: Let's fix it here and have another PR for the rest of the code, how does that sound?

apis/src/openai/responses/rehydrate/mod.rs
```diff
@@ -100,9 +100,6 @@ impl RehydrateFilter {
     /// Parse body, resolve rehydration source (`previous_response_id` or
```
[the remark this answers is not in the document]
derekhiggins: done

apis/src/openai/responses/validate/mod.rs
derekhiggins: done

apis/src/openai/responses/error.rs
```diff
@@ -115,6 +130,21 @@ mod tests {
...
+            400,
+            "invalid_request_error",
+            "mutually_exclusive_parameters",
+            "bad input",
+        );
+        let body: serde_json::Value = serde_json::from_slice(rejection.body.as_deref().unwrap()).unwrap();
+
+        assert_eq!(body["error"]["type"], "invalid_request_error");
+        assert_eq!(body["error"]["code"], "mutually_exclusive_parameters");
+        assert!(body["error"]["param"].is_null());
```
leseb: no bare assert as per project convention

leseb (approved):
just one nit

Issue #36, opened 2026-06-08
Translate Chat Completions streaming responses to Responses SSE


leseb:
## Tracking

- Parent translation epic: #363
- Part of the Responses API program: #93
- Request and finite-response counterpart: #35
- Reasoning extensions: #31

## Purpose

Translate a streaming Chat Completions response from a backend that only speaks
`/v1/chat/completions` into canonical Responses API SSE for a client that called
`/v1/responses` with `stream: true`.

This is the streaming response half of `responses_to_chat_completions`. The base
filter already translates the request and finite response but intentionally
leaves Chat Completions SSE unchanged.

## Scope

The portable translator covers the semantics that standard Chat Completions
chunks can represent faithfully:

- assistant text;
- refusal content;
- function tool calls and incrementally delivered arguments;
- model, service tier, finish reason, and usage; and
- Responses lifecycle, item, content-part, and terminal events required for
  those outputs.

It does not attempt to synthesize every event in the Responses stream-event
union. Provider-specific reasoning or other extension fields require an
explicit backend dialect and are tracked by #31.

## Activation and representation

- Run only when `responses_to_chat_completions` translated a streaming
  Responses request.
- Translate only a successful upstream `text/event-stream` response.
- Reject an unexpected success status, media type, content encoding, or other
  untranslatable representation while response headers are still mutable.
- Preserve the existing finite provider-error normalization for non-streaming
  error responses.
- Rewrite representation headers invalidated by transforming the stream.
- Leave native `/v1/responses` streams and unrelated traffic unchanged.

## Event mapping

Emit canonical event names and payloads from the pinned OpenAI schema.

### Response lifecycle

- Emit exactly one `response.created` followed by one
  `response.in_progress` before the first translated semantic delta, after a
  valid upstream stream has been accepted.
- Reuse the response ID, creation time, request context, and state established
  by the base translation path.
- Assign a strictly increasing `sequence_number` to every emitted event.
- End with exactly one `response.completed`, `response.incomplete`, or
  `response.failed` when a terminal event can still be written.
- Response lifecycle events carry the schema-defined Response resource
  snapshot. Item, content-part, and delta events carry only their schema-defined
  payloads.

### Assistant text and refusal

- Open the assistant message once with `response.output_item.added`.
- Open each represented content part once with
  `response.content_part.added`.
- Convert text chunks to `response.output_text.delta` and finish them with
  `response.output_text.done`.
- Convert refusal chunks to `response.refusal.delta` and finish them with
[...]

Issue #31, opened 2026-06-01
Responses-to-Chat Completions reasoning translation


leseb:
## Tracking

- Parent translation epic: #363
- Builds on request and finite-response translation: #35
- Streaming response conversion: #36
- Related Responses program: #93

## Purpose

Translate reasoning semantics when the client calls `/v1/responses` but the selected inference backend only supports `/v1/chat/completions`.

This issue applies only to the explicit `responses_to_chat_completions` compatibility path. Native `/v1/responses` backends remain passthrough and do not use this translation.

## Current state

The base translation implemented by #35:

- maps `reasoning.effort` to Chat Completions `reasoning_effort`;
- converts finite Chat Completions responses into Responses resources; and
- leaves streaming Chat Completions conversion to #36.

The remaining reasoning contract is not defined:

- `reasoning.summary` has no portable Chat Completions equivalent;
- some compatible backends expose provider-specific raw reasoning fields;
- raw reasoning must not be mislabeled as a user-safe reasoning summary; and
- unsupported reasoning options must not be silently discarded.

### Request translation

- Continue mapping `reasoning.effort` to `reasoning_effort`.
- Preserve supported effort values according to the selected backend.
- Treat deprecated `reasoning.generate_summary` as an alias for `reasoning.summary` only where the Responses compatibility policy requires it.
- Reject conflicting `summary` and `generate_summary` values.
- Do not silently drop a requested reasoning summary.
- When the selected Chat Completions backend has no configured summary capability, return a Responses-compatible unsupported-capability error before forwarding the request.
- Handle backend-specific extensions through an explicit, named reasoning dialect or capability rather than guessing from arbitrary JSON fields.

### Finite response translation

When the backend returns a recognized reasoning representation:

- Create a Responses `reasoning` output item before the corresponding assistant message.
- Map a provider-declared safe summary into the reasoning item's `summary` array.
- Map provider-declared raw reasoning into the appropriate reasoning-content representation.
- Never promote raw reasoning or chain-of-thought into `summary`.
- Preserve backend-reported `reasoning_tokens` in `usage.output_tokens_details.reasoning_tokens`.
- Apply explicit size limits to provider reasoning fields.
- Treat malformed configured reasoning fields as an invalid upstream response rather than silently omitting them.

When the backend returns no reasoning representation:

- Return the ordinary assistant message when no summary was requested.
- Return an unsupported-capability error when a summary was explicitly requested and the backend cannot provide one.

### Streaming response translation

[...]

PR #1014, opened 2026-09-08
feat(openai): Reasoning translation for responses to chat completions

s-akhtar-baig:
## Summary

Adds a configurable reasoning dialect to the responses_to_chat_completions filter so that raw chain-of-thought returned by a Chat Completions backend is surfaced as a first-class Responses reasoning output item. The first dialect is vllm; the default (none) preserves existing behavior.

A reasoning summary request is currently rejected and is out of scope for #31.  

## Related issue

Implements part one for #31. Streaming reasoning translation is deferred to #36 and will be added as a follow-up PR once all the bug fixes land in main.   

## Validation

- [x] Unit tests
- [x] `make lint`

## Checklist


leseb (changes requested):
[P1] Reasoning breaks stored continuations. Emitted reasoning items are persisted and rehydrated, but Chat translation rejects type: "reasoning" as unsupported. previous_response_id and agentic tool continuations therefore return 400 after a reasoning-bearing response. [Emission](https://github.com/praxis-proxy/ai/blob/6f8ab27295596ebae34d3f67778010eded92c207/apis/src/openai/translation/chat_completions.rs#L1605-L1613), [rejection](https://github.com/praxis-proxy/ai/blob/6f8ab27295596ebae34d3f67778010eded92c207/apis/src/openai/translation/chat_completions.rs#L540-L553)

[P1] Streaming silently discards vLLM reasoning. Streaming requests pass validation, but ChatDelta does not model vLLM’s delta.reasoning; unknown fields are ignored. Either translate reasoning deltas or reject streaming when the dialect is enabled. [Validation](https://github.com/praxis-proxy/ai/blob/6f8ab27295596ebae34d3f67778010eded92c207/apis/src/openai/responses/responses_to_chat_completions/mod.rs#L545-L559), [stream delta](https://github.com/praxis-proxy/ai/blob/6f8ab27295596ebae34d3f67778010eded92c207/apis/src/openai/responses/responses_to_chat_completions/stream/chat.rs#L56-L77)

[P2] Malformed summary controls are silently accepted. summary and generate_summary use Value::as_str, making booleans, objects, arrays, and numbers indistinguishable from absent fields. Requests such as "summary": true are forwarded with the option silently dropped. [reasoning.rs](https://github.com/praxis-proxy/ai/blob/6f8ab27295596ebae34d3f67778010eded92c207/apis/src/openai/translation/reasoning.rs#L54-L79)

s-akhtar-baig:
Thanks, @leseb, for reviewing this PR! 

I have updated accordingly and will implement streaming reasoning in a follow-up PR. PTAL when you get a chance.

leseb:
> Thanks, @leseb, for reviewing this PR!
> 
> I have updated accordingly and will implement streaming reasoning in a follow-up PR. PTAL when you get a chance.

Thanks, another round :)

[P1] Continuation reasoning is silently discarded. The previous 400 was replaced with an empty match arm. Stored continuations and agentic tool rounds now succeed but omit the prior raw reasoning, which reasoning models may require between tool calls. [chat_completions.rs:558](https://github.com/praxis-proxy/ai/blob/eaec739d7ecc19625641a96e65260066d002d6e3/apis/src/openai/translation/chat_completions.rs#L548-L562)

[P2] Reasoning-only completions are rejected. Response validation counts only content, refusal, or tool calls as output. A valid vLLM response containing reasoning but no final text fails before the new reasoning item can be emitted. [chat_completions.rs:1267](https://github.com/praxis-proxy/ai/blob/eaec739d7ecc19625641a96e65260066d002d6e3/apis/src/openai/translation/chat_completions.rs#L1267-L1278)

[P2] Non-object reasoning is silently ignored. Values such as "reasoning": true bypass validation, lose their controls during translation, and can receive an ordinary successful response. This proxy-owned field should return 400 when it is neither an object nor null. [reasoning.rs:92](https://github.com/praxis-proxy/ai/blob/eaec739d7ecc19625641a96e65260066d002d6e3/apis/src/openai/translation/reasoning.rs#L90-L104)

leseb:
we also need new tests in tests/integration/sdk/openai/test_openai_responses_vllm.py

s-akhtar-baig:
Thanks, @leseb!

```
[P1] Continuation reasoning is silently discarded. The previous 400 was replaced with an empty match arm. Stored continuations and agentic tool rounds now succeed but omit the prior raw reasoning, which reasoning models may require between tool calls.
```

For P1, I am not sure where to add reasoning across calls. I looked at the OpenAI and vLLM chat completions docs and neither of those show re-adding raw reasoning as part of input messages or any other field. Did you have anything specific in mind?

s-akhtar-baig:
@leseb, can you please re-review when you get a chance? Thanks!

leseb:
> Thanks, @leseb!
> 
> ```
> [P1] Continuation reasoning is silently discarded. The previous 400 was replaced with an empty match arm. Stored continuations and agentic tool rounds now succeed but omit the prior raw reasoning, which reasoning models may require between tool calls.
> ```
> 
> For P1, I am not sure where to add reasoning across calls. I looked at the OpenAI and vLLM chat completions docs and neither of those show re-adding raw reasoning as part of input messages or any other field. Did you have anything specific in mind?

The main concern is the silent drop. For this PR, we should either:

* reject reasoning input items and document continuation as unsupported.
* combine the rehydrated reasoning item with that assistant turn.

I'm leaning implementing the replay now.

s-akhtar-baig:
@leseb, I have implemented replay for reasoning. The prior turn's chain-of-thought now gets added to the assistant message that follows it, wrapped in markers (<think>…</think> by default). Note that these delimiters are configurable. PTAL when you get a chance, thanks!

leseb (changes requested):
One blocker remains.

The silent drop is fixed, but reasoning is replayed as ordinary content:

```json
{"content":"<think>...</think>answer"}
```

For the `vllm` dialect, it should instead be:

```json
{"reasoning":"...","content":"answer"}
```

vLLM supports this field and uses it in its own [Responses converter](https://github.com/vllm-project/vllm/blob/main/vllm/entrypoints/openai/responses/utils.py). Marker injection is model/template-dependent.

Recommendation: replace the marker logic with `assistant.reasoning`, update the tests, then approve. The focused 160 tests pass and the branch merges cleanly, but full CI is still pending. I posted nothing on GitHub.

s-akhtar-baig:
Thanks, @leseb, for pointing this out!

I have removed the markers and replaced with the reasoning field. PTAL when you get a chance.

leseb:
@s-akhtar-baig last one: multi-round agentic reasoning produces duplicate rs_<response_id> item IDs because the response ID stays constant across rounds. Make the ID round/item-specific and add a two-round regression test.

leseb (approved):

leseb:
@s-akhtar-baig ci failures

s-akhtar-baig:
@leseb, auto-merge failed on conflicts with main. I have updated the recordings accordingly. Please help merge this when you get a chance. Thanks again!

PR #1030, opened 2026-09-09
fix(stream-events): enforce timeout_secs after a silent backend

### LINKED ISSUE #938: openai_stream_events timeout cannot terminate an idle backend stream
## Clawpatch finding

- **Severity:** medium
- **Category:** bug
- **Confidence:** high
- **Triage:** confirmed-bug
- **Finding ID:** `fnd_sig-feat-custom-apis-openai-resp_43b04931ed`
- **Run:** `20260907T053712-f49e16`
- **Prior-run overlap:** No exact prior Clawpatch finding or GitHub issue found.

## Evidence

- `apis/src/openai/responses/stream_events/config.rs:36-39`
- `apis/src/openai/responses/stream_events/mod.rs:247-263`
- `apis/src/openai/responses/stream_events/mod.rs:333-340`
- `apis/src/openai/responses/stream_events/mod.rs:565-580`
- `apis/src/openai/responses/stream_events/tests.rs:49-54`

## Description

The timeout is checked on incoming chunks and at EOS, so it cannot fire while the backend remains silent. After the first SSE chunk, a backend can keep the connection open indefinitely without sending another chunk or closing it; no filter callback occurs and timeout_secs never causes termination. This makes the documented wall-clock timeout ineffective for the primary stalled-stream failure mode and can retain request state and connections until a separate transport deadline intervenes.

## Reproduction

Configure timeout_secs: 1. Have a backend send one valid non-terminal SSE event and then sleep for at least 10 seconds without sending data or closing the connection. Observe that openai_stream_events does not terminate the stream after one second; it detects the timeout only if another chunk or EOS eventually arrives.

## Expected behavior

The proxy must fail closed or preserve the documented API and filter contract without silently changing successful output, permissions, durability, or resource limits.

## Recommendation

Enforce this deadline in an asynchronous transport or request-lifecycle layer capable of waking without body traffic, or remove the misleading per-filter timeout and explicitly rely on a required transport deadline. Ensure the effective timeout aborts or terminates the stream when no further chunks arrive.

## Validation

The filter checks elapsed time only from body callbacks and EOS. Once a backend becomes silent, no callback occurs, so `timeout_secs` cannot enforce its documented wall-clock stream deadline without transport-layer support.

---
Generated from scheduled Clawpatch Codex `gpt-5.6-sol` / `xhigh` review runs `20260907T052439-47e401, 20260907T052823-e5e76e, 20260907T053119-d89f7d, 20260907T053325-beffe2, 20260907T053712-f49e16`. Validate before assigning or implementing.

mkoushni:
## Summary

`openai_stream_events.timeout_secs` was only checked when an SSE chunk or end-of-stream arrived. After the first event, a silent backend never invokes those callbacks, so the documented wall-clock budget did not fire.

- Cap the selected peer's `read_timeout` at `timeout_secs` (without relaxing a tighter cluster timeout) from `on_request` and `on_request_body`.
- Treat IRR idle/deadline stream termination as a stream timeout and record the same `responses.stream_error_*` metadata as a parser timeout.
- Pair the stream-events and agentic-loop examples with cluster `read_timeout_ms` so a silent connection is torn down without waiting for another body callback.

A filter-only timer cannot wake Pingora without body traffic. Cluster `read_timeout_ms` remains the reliable idle deadline when load balancing has not yet set `ctx.upstream`.

## Related issue

Closes #938

## Validation

- [x] Unit tests
- [x] `make lint`

```console
cargo clippy -p praxis-ai-apis --all-targets -- -D warnings
cargo test -p praxis-ai-apis --lib -- stream_events
cargo test -p praxis-tests-integration --test suite -- \
  stream_events_idle_backend_is_cut_off_by_read_timeout \
  stream_events_accumulates_state
```

The new integration test sends one `response.in_progress` event, stalls 10s, and asserts the proxy cuts the stream in under 4s via `read_timeout_ms: 1000` plus `timeout_secs: 1`.

## Checklist


leseb (changes requested):
blocked by https://github.com/praxis-proxy/ai/pull/1013 since this uses logical_stream and 1031 removes it 

in the meantime:

P2 — Lines 330–333 ignore the actual filter-induced timeout classification. Praxis 0.5.4 reports a winning peer read_timeout as StreamTerminationCause::Io, not IdleTimeout; logical error bytes are then discarded because IRR still considers the termination unhandled.
P2 — Lines 311–316 implement a per-read timeout, not the documented duration “from first chunk to completion.” It can expire before the first event, while a stream that stalls near the deadline can survive almost twice the configured duration.

leseb (changes requested):
P1: timeout_secs is never applied in the documented IRR ordering because both cap attempts occur before load balancing selects ctx.upstream. The integration test passes only because it separately sets read_timeout_ms.
P2: matching every StreamTerminationCause::Io mislabels ordinary transport failures as timeouts. Coverage fails deterministically: 622 passed, 1 failed.
P2: read_timeout remains per-read, not an absolute deadline from the first SSE chunk; it may fire before that chunk or permit nearly 2T.

mkoushni:
P1 — cap after load balancing. openai_responses_proxy buffers the request body, so IRR runs that phase before load_balancer sets ctx.upstream. Both earlier cap attempts were no-ops; the idle test only passed because it also set read_timeout_ms: 1000. The documented stream-events.yaml (and irr-terminal-streaming.yaml) now place openai_stream_events after load_balancer, and on_request caps the selected peer. The idle test no longer tightens cluster read_timeout_ms.

Agentic configs keep the filter before openai_agentic_loop so the loop still sees raw SSE on the reverse response path. Idle there still pairs with cluster read_timeout_ms.

P2 — Io mislabel. Praxis reports a winning peer read_timeout as Io, but so do truncated chunks. Io is a stream timeout only after a first SSE chunk and timeout_secs has elapsed. late_upstream_failure_does_not_replace_committed_sse now asserts the body does not contain exceeded timeout.

P2 — remaining budget. The parser deadline still starts at the first SSE chunk. After that chunk the peer cap is the leftover time, not another full period. Praxis snapshots read_timeout at dispatch, so later remaining-budget caps apply only while ctx.upstream is still visible.

leseb:
@mkoushni final 

After each chunk, the filter tries to shorten ctx.upstream.read_timeout, but IRR response-body contexts have upstream: None; Praxis already copied the original timeout into the live body. A chunk near deadline therefore restarts the full per-read timer, allowing nearly 2T instead of an absolute T.

leseb:
@mkoushni Still no. The new patch restores and mutates a saved Upstream, but Praxis already copied its timeout into the live SubResponseBody. The restored object is detached, so the active read timer remains unchanged and can still overrun.

Almost there!

mkoushni:
> @mkoushni Still no. The new patch restores and mutates a saved Upstream, but Praxis already copied its timeout into the live SubResponseBody. The restored object is detached, so the active read timer remains unchanged and can still overrun.
> 
> Almost there!

@leseb I added the solution in Praxis project, which change copies leftover budget onto the live SubResponseBody after body filters. Restoring ctx.upstream only publishes the cap; it is not the active read timer. openai_stream_events still writes leftover onto the restored peer so 1030 stays on crates.io 0.5.4. Once 1172 is in a Praxis release, bump AI onto it and leftover recaps will actually shorten the next stall.
https://github.com/praxis-proxy/praxis/pull/1172

If you don't agree with this, I can put everything back the way it was.
But it will be partially fix:

Cap the selected peer at arm time, after load balancing, so dispatch snapshots timeout_secs (or a tighter cluster value).
Treat a later idle/Io abort as a stream timeout once the first chunk has exhausted that budget.
Enforce the parser deadline when another chunk or EOS actually arrives.

Cargo.toml
```diff
@@ -88,11 +88,12 @@ tokio-stream = "0.1.19"
 zeroize = "1.9.0"
 
 # Praxis core dependencies
-praxis-core = { version = "0.5.5", package = "praxis-proxy-core" }
-praxis-filter = { version = "0.5.5", package = "praxis-proxy-filter" }
-praxis-protocol = { version = "0.5.5", package = "praxis-proxy-protocol" }
-praxis-tls = { version = "0.5.5", package = "praxis-proxy-tls" }
-praxis = { version = "0.5.5", package = "praxis-proxy" }
+# Pin to mkoushni/praxis until praxis-proxy/praxis ships cap_stream_deadline.
```
leseb: ???

leseb: i'm gonna tag 0.5.6 is core today

leseb (commented):
blocked on 0.5.6 core release

Cargo.toml
mkoushni: how i can missed it?

mkoushni: all temp test remarks removed

leseb (approved):
ready once rebased over https://github.com/praxis-proxy/ai/pull/1203

PR #1047, opened 2026-09-10
perf(anthropic-translation): consume owned request fields and tool schemas

### LINKED ISSUE #968: perf(anthropic-translation): consume owned request fields and tool schemas
## Summary

Consume already-owned fields while translating an Anthropic Messages request to a Chat Completions-compatible request, instead of cloning values from a JSON object that is discarded immediately after translation.

Tool input schemas can be substantially larger than scalar request parameters, so preserve ownership through tool conversion as well as the top-level mapping.

## Scope

- Consume the parsed top-level request object and move mapped fields such as `model`, token limits, streaming configuration, stop sequences, and sampling parameters where possible.
- Consume translatable client-tool definitions so `input_schema` moves into the generated function parameters instead of being deep-cloned.
- Keep allocations that are required to reshape Anthropic messages or construct new Chat Completions content.

## Acceptance criteria

- [ ] Mapped top-level JSON values and client-tool input schemas move into the translated request without a deep clone.
- [ ] Generated Chat Completions requests remain semantically identical, including field names and tool schema contents.
- [ ] Typed Anthropic server tools continue to be excluded from client-tool translation so unknown server tools fail closed; only untyped or explicit `custom` tools are translated.
- [ ] System, message, tool-use, tool-result, document, image, tool-choice, and parallel-tool behavior remain unchanged.
- [ ] No removed clone is merely moved into another translation helper.
- [ ] Focused Anthropic-to-OpenAI translation tests pass.

## Parent

- #898

r-papso:
### Why the diff is shaped this way

`AnthropicMessagesToChatCompletionsFilter` contained many redundant deep clones at multiple nesting levels across the request JSON body. Replacing those clones with moves means each field can only be read once, so `transform_request` now deconstructs the parsed body into owned locals up front instead of threading a `&mut Map` through every helper. Threading it and `remove`ing incrementally could lead to silent errors: a later read returns `None` on already moved field. Each field is now a let moved exactly once, so a double-consume is a compile error.

Output is byte-identical, the change is ownership only.

### Addressed redundant allocations

1. **Double request body parse** (`51c48e7b`) — parsed once. Metadata borrows, translation consumes.
2. **Top-level mapped fields** (`cdce7809`) — `model`, `max_tokens`, `system`, `messages`, `stream`, `stream_options`, `stop_sequences`, `temperature`, `top_p`, `top_k`, `tools`, `tool_choice` moved instead of cloned. Remainder dropped.
4. **Tool `input_schema`** (`cdce7809`) — previously deep-cloned into `function.parameters`. The tool object is hand-built rather than via `json!`, which would re-serialize the function map.
5. **Message content payloads** (`99bbcf58`) — text, base64 image data, and `tool_use.input` consumed out of their blocks rather than copied via `as_str().to_owned()`.
7. **`search_result` / `document` flattening** (`796b0d2f`) — appended into a single `String` instead of `Vec<String>` + `join`, which copied every payload twice.
8. **Output buffer growth** (`149e5564`) — pre-sized from the original body length instead of growing from empty. `original_len` is a bounded estimate.

### Known remaining, deferred

Listed so the boundary is explicit:

- `split_tool_result_parts` and `hoist_system` — the same `Vec<String>` + `join` pattern fixed in `796b0d2f`.
- `flatten_document_source — `document `source.content` text copied three times (`Vec` → join → `format!`) (non-trivial to address, `format!` is needed for Anthropic → OpenAI translation).
- `request.rs:526` — `quote_label_value` allocates a temporary per label; `file:`/`base64:` sources allocate twice (minor - short string labels).

Closes #968 

## Validation

- [x] `cargo test -p praxis-ai-apis`
- [x] `make lint`

## Checklist

## Breaking changes

No breaking change.

### DIFF
--- a/apis/src/anthropic/messages_to_chat_completions/mod.rs
+++ b/apis/src/anthropic/messages_to_chat_completions/mod.rs
@@ -153,8 +153,10 @@ impl HttpFilter for AnthropicMessagesToChatCompletionsFilter {
             _ => return Ok(FilterAction::Continue),
         };
 
-        extract_request_metadata(ctx, bytes);
-        Ok(transform_request_body(body))
+        let value = serde_json::from_slice::<serde_json::Value>(bytes);
+
+        extract_request_metadata(ctx, &value);
+        Ok(transform_request_body(body, value))
     }
 
     fn on_response_body(
@@ -200,9 +202,9 @@ impl HttpFilter for AnthropicMessagesToChatCompletionsFilter {
 // Request Body Helpers
 // -----------------------------------------------------------------------------
 
-/// Extract streaming and model metadata from the request body.
-fn extract_request_metadata(ctx: &mut HttpFilterContext<'_>, bytes: &[u8]) {
-    let Ok(value) = serde_json::from_slice::<serde_json::Value>(bytes) else {
+/// Extract streaming and model metadata from the parsed request body.
+fn extract_request_metadata(ctx: &mut HttpFilterContext<'_>, value: &Result<serde_json::Value, serde_json::Error>) {
+    let Ok(value) = value else {
         ctx.set_metadata("anthropic_messages_to_chat_completions.streaming", "false");
         return;
     };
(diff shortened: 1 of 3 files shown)


leseb (changes requested):
[P2] Vec::with_capacity(original_len) can retain up to 64 MiB for a tiny translated request when input is mostly whitespace or filtered server-tool data. [Bytes::from(Vec) retains that capacity](https://github.com/praxis-proxy/ai/blob/6d8c1fb75d0338cbc76a931230099ef99a6b4aa0/apis/src/anthropic/messages_to_chat_completions/request.rs#L80-L85), increasing per-request memory and potentially making performance worse. Restore serde_json::to_vec or use a conservative bounded estimate.

r-papso:
@leseb thanks for the review, I addressed your comment here: [3ac4a00](https://github.com/praxis-proxy/ai/pull/1047/changes/3ac4a00dd0a0fc36815528883825089c1c5b0772).

leseb (approved):

Issue #770, opened 2026-08-18
bug(token_count): stream_options.include_usage never forced for direct Chat Completions streaming


jordigilh:
## Summary

For direct Chat Completions streaming (`stream: true`, no Responses API translation in front), nothing forces `stream_options.include_usage=true` on the outbound request. If the client doesn't set it themselves, OpenAI never sends a `usage` object in the final SSE chunk, so `token_count` extracts nothing and `token.input`/`token.output`/`token.total` metadata never gets set for that request.

## Where this actually matters

This isn't hypothetical for the examples in this repo: `examples/configs/token-counting.yaml` and `examples/configs/token-rate-limit.yaml` both wire `token_count` directly against a plain Chat Completions passthrough (`router` -> `token_count`/`token_rate_limit` -> `load_balancer`), with no Responses-API translation filter in the chain. Any deployment shaped like those examples that also streams will silently get zero usage data unless the *client* happens to set the flag itself.

## Root cause / where it *is* handled

`apis/src/openai/translation/chat_completions.rs::map_stream_options` (called from the Responses -> Chat Completions translation path only) already forces `stream_options.include_usage = true` on outbound streaming requests:

```rust
options.insert("include_usage".to_owned(), Value::Bool(true));
target.insert("stream_options".to_owned(), Value::Object(options));
```

That function has exactly one call site, inside the `responses_to_chat_completions` translation. There's no equivalent for the direct/passthrough Chat Completions path -- `filters/src/token_usage/` (the `token_count` filter and friends) never reads or mutates `stream_options` at all, it's a pure response-side parser.

Anthropic doesn't need an equivalent: its streaming API returns usage in `message_delta`/`message_stop` unconditionally, no opt-in flag required (confirmed: no `stream_options`/`include_usage` references anywhere under `apis/src/anthropic/`).

## Related

- Called out explicitly (but left unaddressed) in #713's description: "The separate issue of injecting `stream_options.include_usage=true` for Chat Completions streaming is not addressed here -- that requires a request-body mutation filter."
- Raised by @usize on the `token_rate_limit` proposal thread (ai#658, https://github.com/praxis-proxy/ai/pull/658#discussion_r3798214138), which is what prompted this writeup -- accurate reconciliation in that proposal (and in the already-shipped `token_count`/`token_usage_headers` filters) depends entirely on usage actually coming back from the provider.
[...]

PR #1132, opened 2026-09-11
feat(token-rate-limit): apply token-type weights at reconciliation

### LINKED ISSUE #884: Token-type-aware accounting with configurable weights
## What

Apply configurable weights per token type during reconciliation in the
`token_rate_limit` filter. Different token types consume different
amounts of budget based on their weight.

Currently, [#796](https://github.com/praxis-proxy/ai/pull/796) uses
`token.total` only — all tokens are weighted equally.

## Why

Providers charge significantly less for cached input tokens than uncached
(~10x for some providers). A quota system that treats all tokens equally
will over-restrict users who benefit from caching. The
[proposal](https://github.com/praxis-proxy/enhancements/blob/main/proposals/00121_token-rate-limiting.md)
defines this as M4.

## Goals

- Filter-wide default weights (e.g., `cached_input: 0.1`, `reasoning: 0.9`)
- Per-rule weight overrides
- Weighted cost at reconciliation: `cost = Σ(tokens_of_type × weight_of_type)`
- Fallback to `token.total` with weight 1.0 when granular types are unavailable

## Dependencies

- Cache tokens (`token.cache_read`, `token.cache_write`) are already
  emitted by `token_count` — weighting for cached tokens can start now.
- Reasoning tokens require the reasoning token extraction issue
  to land first. Weighting for reasoning tokens is blocked on that.

## References

- Proposal M4: https://github.com/praxis-proxy/enhancements/blob/main/proposals/00121_token-rate-limiting.md
- grid#101 P1: https://github.com/praxis-proxy/grid/issues/101
- Foundation: [#796](https://github.com/praxis-proxy/ai/pull/796) (merged)

abdallahsamabd:
## Summary

Teach `token_rate_limit` to charge quota by token type at reconciliation (proposal M4 / ai#884), instead of always settling `token.total` at weight 1.0. Cached prompt tokens are much cheaper at the provider; treating them as full-price over-restricts cache-friendly traffic.

**What operators configure**

- Filter-wide `default_weights` (omitted types stay `1.0`).
- Optional per-rule `weights` overlay (only named types replace the defaults).
- Admission is unchanged: still reserves unweighted `reserved_tokens` (M3 / ai#882).

**How the cost is computed** (not a naive sum of parent + child):
cost = ceil( uncached_input × input + cache_read × cached_input + cache_write × cache_write + visible_output × output + reasoning × reasoning )

Cache read/write are partitioned *out of* `token.input`. Reasoning is nested inside `token.output` on OpenAI / Azure / Anthropic (`token.total ≈ input + output`) and additive on Google (`token.total ≈ input + output + reasoning`). The filter picks nested vs additive by comparing those two identities; missing `token.total` defaults to nested so the common providers never double-count.

**Fallbacks (same fail-closed spirit as #796)**

| Condition | Settlement |
|---|---|
| `token.status == overflow` | keep the reservation estimate |
| typed `token.input` + `token.output` present | partitioned weighted cost |
| parents missing, `token.total` present | `token.total` at weight 1.0 (Bedrock Converse, Chat Completions without details, …) |
| nothing usable (streaming without `include_usage`, #770) | keep the reservation estimate |
| fractional cost (9 × 0.1 = 0.9) | `ceil` → 1 |

Does **not** re-parse provider JSON in the limiter (that caused #351). Reads `token.*` metadata that `token_count` already published. Ledgers / Valkey Lua stay `u64`; rounding happens in the filter.

### Changes

- **`filters/src/token_rate_limit/weights.rs`** — `TokenTypeWeightsConfig` / `TokenWeights`, `UsageCounts::from_context`, partitioned `weighted_cost`, nested-vs-additive heuristic, `ceil`.
- **`config.rs` / `mod.rs`** — `default_weights` on the filter, `weights` on each rule; resolve at compile time; `reconciliation_context` uses `weighted_cost` instead of parsing only `token.total`.
- **`token_usage/mod.rs`** — `token.*` metadata key constants are `pub(crate)` so the limiter shares the #351 strings instead of duplicating them.
- **Examples** — `token-rate-limit.yaml` ships proposal-like defaults; mixed-algorithms overlays `cached_input: 0.05` on `team-alpha`.
- **Docs** — regenerated `docs/filters/token_rate_limit.md` from the module YAML example.
- **Tests** — unit coverage for cache partition, Anthropic cache_write 1.25×, OpenAI nested vs Google additive reasoning, overflow, total-only fallback, ceil, overlay / unknown-key / `.nan` rejection; HTTP-level tests for both algorithms; integration test that the example's `cached_input: 0.1` admits a second request that unweighted `token.total` would deny.

## Related issue

Closes #884

## Validation

- [x] Unit tests
- [ ] `make lint`


szedan-rh:
The change making sense and looks good.
@alexsnaps / @aslakknutsen - are we on code freeze or still we can merge?

szedan-rh:
Following up on my earlier "looks good" — one regression I want to flag before this merges, plus a few small cleanups.

### Bedrock Converse with prompt caching stops charging cached tokens

`weighted_cost` uses `token.total` only for the nested-vs-additive reasoning check and otherwise discards it once `token.input` and `token.output` are both present. That is fine for OpenAI, Anthropic and Google, where `token.input` already includes cache tokens or `token.total` equals the sum of the typed parts. It is not fine for Bedrock Converse.

`parse_bedrock` reads only `inputTokens`, `outputTokens` and `totalTokens`, and does not parse `cacheReadInputTokens` / `cacheWriteInputTokens`. Per the AWS prompt caching docs, when caching is enabled `inputTokens` holds only the non-cached tokens, so `token.input` for Bedrock is uncached-only (unlike Anthropic, where `token_count` normalizes it to uncached + read + write). The cached tokens live only in `token.total`.

Concrete example (cached Converse response):

| field | value |
|---|---|
| inputTokens | 9 |
| outputTokens | 214 |
| cacheReadInputTokens | 1066 |
| totalTokens | 1289 |
| settled before this PR | 1289 |
| settled after this PR | 223 |

So the 1066 cached tokens become free, even at unity weights with no `default_weights` configured. The PR body says Bedrock Converse takes the total-only fallback, but it does not, because the parser publishes both parents.

Two ways to close it, and I'd suggest doing both:

1. Fail-closed guard in `partitioned_cost`: compute the residual `total - (input + output [+ reasoning when additive])` and charge it at the `input` weight, so the weighted cost never drops below what `token.total` covered.
2. Root cause in `token_count`: parse the two Bedrock cache fields and normalize `token.input` the way `parse_anthropic` already does, so Bedrock gets a real cache breakdown and the partition applies.

Either way, a Bedrock cached-usage unit test would be good — none of the new tests cover a provider whose `token.total` exceeds its typed parents.

### Smaller items

- `per_rule_weight_overlay_is_isolated_from_filter_defaults`: the `default-cache` half publishes only `token.total`, so it exercises the fallback path rather than the filter-wide `cached_input: 0.1`. An overlay leak into `filter_defaults` would still pass. Publishing typed counts with `cache_read=90` for that rule and asserting cost 29 would actually prove isolation.
- `weighted_cache_hit_refunds_more_than_token_total`: the comment still carries working notes ("= 240... wait that's still > 100") for a fixture the test doesn't use. Keeping only the 100/90/10 → 29 explanation would help the next reader.
- `parse_u64_meta` in `weights.rs` is the same pattern `reconciliation_context` hand-rolls for `META_RESERVATION_ID` and `META_ESTIMATE`. Making it `pub(super)` and reusing it would keep the metadata parsing in one place.
- The nested-vs-additive heuristic infers provider semantics arithmetically even though `token_count` knows the `ProviderKind`. Not blocking here, but the Bedrock gap is a symptom of that split, and a future provider with a different total identity would miscount silently. Worth a follow-up issue to have `token_count` publish normalized semantics.

Everything else checks out: the partition math is correct for OpenAI, Anthropic and Google in both streaming and non-streaming paths, unity weights reproduce the old behavior for those providers, the overflow path now correctly keeps the estimate instead of charging partial counts, and `cargo xtask lint-filter-docs` confirms the generated docs are in sync.

abdallahsamabd:
@szedan-rh 
fixed

szedan-rh:
@abdallahsamabd - Fix the lint issue please

abdallahsamabd:
@szedan-rh 
done

szedan-rh (approved):
LGTM

nerdalert (changes requested):
Thanks for the thorough implementation and follow-up on the Bedrock accounting path. The normalization and provider coverage look good. I found one arithmetic edge that should be closed before merge.

### Large finite weights can settle a request at zero

`checked_weight` accepts any finite non-negative `f64`, but multiplying a large finite weight by a token count can overflow to positive infinity. `ceil_cost` currently maps every non-finite result to `0`, so that request is reconciled as zero actual usage and the ledger refunds its reservation.

Could we make positive arithmetic overflow fail closed by saturating to `COST_U64_CAP`? For example:

```suggestion
fn ceil_cost(cost: f64) -> u64 {
    if cost.is_nan() || cost <= 0.0 {
        return 0;
    }
    if cost.is_infinite() {
        return COST_U64_CAP;
    }
    let capped = cost.min(COST_F64_CAP).ceil();
    #[expect(
        clippy::cast_possible_truncation,
        clippy::cast_sign_loss,
        reason = "capped is finite, non-negative, and <= 2^53, so ceil is an exact u64"
    )]
    {
        capped as u64
    }
}
```

A regression test using a very large but finite configured weight would pin the behavior. Bounding accepted weights or using explicitly saturating accumulation would also resolve it; the important property is that arithmetic overflow cannot turn usage into a zero-cost settlement.

abdallahsamabd:
@szedan-rh  @nerdalert 
can you please re-review
thanks

nerdalert (approved):
LGTM 🚀 

  - Token-type weighting correctly partitions input, output, cached input, cache writes, and reasoning tokens.
  - OpenAI and Anthropic nested reasoning is not double-counted.
  - Google’s additive reasoning tokens are charged separately.
  - Bedrock Converse cache reads/writes are normalized into input and retain their cache breakdown.
  - Any usage hidden in token.total is charged rather than silently dropped.
  - Streaming and non-streaming provider paths publish compatible metadata.
  - Per-rule weights correctly overlay defaults without leaking into other rules.
  - Missing or unusable usage metadata preserves the original reservation estimate.
  - Partial counts marked as overflow are not used for settlement.
  - Fractional weighted costs round upward to avoid undercharging.
  - Large finite weights that overflow arithmetic saturate at the ledger-safe cap rather than producing zero usage.
  - Negative, NaN, infinite, and unknown configuration values fail validation.
  - Existing configurations without weights retain unity-weight behavior.
  - Both sliding-window and token-bucket reconciliation use the weighted cost.
  - Documentation and example configurations reflect the implemented behavior.

PR #1136, opened 2026-09-14
feat(routing): reload projected provider credentials

### LINKED ISSUE #1135: feat(routing): reload file-backed provider credentials
## Problem\n\nPraxis can load an upstream bearer credential from a mounted file, but the value is resolved only when the filter is constructed. Rotating a projected Secret therefore requires restarting Praxis, even though Kubernetes publishes Secret updates atomically.\n\n## Proposed behavior\n\nWatch file-backed credential sources and atomically replace validated credential snapshots when their files change.\n\n- Keep filesystem I/O and locking off the request path.\n- Ensure each request observes one complete old or new credential.\n- Fail closed when a replacement is missing, empty, oversized, unreadable, or invalid.\n- Never expose credential values through metadata, logs, traces, status, or error bodies.\n- Preserve existing inline and environment-variable behavior.\n- Strip caller Authorization and x-api-key values at the provider boundary before applying trusted upstream credentials.\n\n## Why this is generally useful\n\nAny Praxis deployment that forwards requests to authenticated upstream APIs needs to rotate provider credentials without restarting the proxy or interrupting in-flight requests. Kubernetes projected Secrets are one common delivery mechanism, but the capability is simply safe reload of file-backed upstream credentials and is not tied to a particular control plane.\n\n## Acceptance criteria\n\n- A valid atomic file replacement is observed without reconstructing the filter.\n- Same-length credential replacements are not missed.\n- Invalid replacements cause new requests to fail closed.\n- In-flight requests retain the immutable snapshot they already loaded.\n- Request processing performs no filesystem reads or write locking.\n- Unit tests, workspace tests, and lint pass.

nerdalert:
## Summary

Adds safe, event-driven reloads for file-backed upstream bearer credentials. Praxis now revalidates mounted credential files after atomic filesystem updates and publishes immutable snapshots to request handlers, allowing credentials to rotate without rebuilding the filter or restarting the proxy.

This is the smallest complete change because it keeps the existing credential reference and injection model intact:

- inline and environment-variable credentials retain their current behavior;
- file-backed credentials gain a bounded watcher;
- requests perform only a map lookup and atomic snapshot load, with no filesystem I/O or write lock;
- valid replacements become available to new requests atomically;
- missing, empty, oversized, unreadable, or invalid replacements make new requests fail closed with HTTP 503;
- in-flight requests retain the complete snapshot they already loaded;
- caller Authorization and x-api-key headers are removed at the provider boundary before a trusted upstream credential is applied;
- credential values are never written to filter metadata, traces, logs, or error bodies.

Kubernetes projected Secrets are a common source of atomic file replacements, but the capability is control-plane agnostic and applies to any Praxis deployment forwarding traffic to authenticated upstream APIs.

The generated credential filter documentation includes the updated reload behavior and example configuration.

## Related issue

Closes #1135

## Validation

- [x] Unit tests
- [x] `make lint`

Commands run:

```sh
cargo test -p praxis-ai-filters credential_inject --lib
cargo test --workspace --all-targets
cargo xtask generate-filter-docs
make lint
git diff --check
```

Focused credential validation passed 26 tests, including:

- atomic same-length credential replacement without filter reconstruction;
- invalid replacement failing closed without credential leakage;
- no request-path rereading of unchanged credentials;
- removal of caller Authorization and x-api-key headers;
- selection of the credential matching the trusted routing reference.

The complete workspace test suite and repository lint/documentation consistency checks passed.

## Checklist

The performance-sensitive request path does not perform file I/O, acquire a write lock, or parse credential contents. It loads an immutable `ArcSwap` snapshot. This is covered by a regression test that confirms repeated requests do not trigger file validation. No standalone load benchmark was added.

## Breaking changes

None. Existing inline, environment-variable, and file-backed configurations remain valid. File-backed sources now reload after relevant filesystem changes and fail closed if a replacement becomes invalid.


leseb (changes requested):
P2 – A rotation can be missed during startup. The file is read before watcher registration, with [no authoritative reread before readiness](https://github.com/praxis-proxy/ai/blob/856e1b87567879fa6dd94927b983e86d0f196e01/filters/src/routing/credential_inject.rs#L523-L536). A rotation in that window leaves the old credential active indefinitely.

P2 – Dropped-event rescans are ignored. [EventKind::Other is discarded](https://github.com/praxis-proxy/ai/blob/856e1b87567879fa6dd94927b983e86d0f196e01/filters/src/routing/credential_inject.rs#L542-L547), but notify uses it with need_rescan() when filesystem events were lost. Credential revocations can therefore be silently missed.

P2 – Pipeline teardown blocks synchronously. [Drop busy-waits](https://github.com/praxis-proxy/ai/blob/856e1b87567879fa6dd94927b983e86d0f196e01/filters/src/routing/credential_inject.rs#L225-L237), while cancellation cannot wake the watcher’s 100 ms recv_timeout. Dropping old pipelines can stall runtime threads, cumulatively across filters.

nerdalert:
> P2 – A rotation can be missed during startup. The file is read before watcher registration, with [no authoritative reread before readiness](https://github.com/praxis-proxy/ai/blob/856e1b87567879fa6dd94927b983e86d0f196e01/filters/src/routing/credential_inject.rs#L523-L536). A rotation in that window leaves the old credential active indefinitely.
> 
> P2 – Dropped-event rescans are ignored. [EventKind::Other is discarded](https://github.com/praxis-proxy/ai/blob/856e1b87567879fa6dd94927b983e86d0f196e01/filters/src/routing/credential_inject.rs#L542-L547), but notify uses it with need_rescan() when filesystem events were lost. Credential revocations can therefore be silently missed.
> 
> P2 – Pipeline teardown blocks synchronously. [Drop busy-waits](https://github.com/praxis-proxy/ai/blob/856e1b87567879fa6dd94927b983e86d0f196e01/filters/src/routing/credential_inject.rs#L225-L237), while cancellation cannot wake the watcher’s 100 ms recv_timeout. Dropping old pipelines can stall runtime threads, cumulatively across filters.

@leseb Thanks!

- Fixed: parent directories are now watched before an authoritative startup reread, closing the missed-rotation window.
- Fixed: EventKind::Other and need_rescan() now trigger a full credential rescan; invalid replacements fail closed.
- Fixed: shutdown now wakes the watcher immediately and joins it directly, removing the busy-wait and receive-timeout delay.

leseb (approved):

PR #1155, opened 2026-09-14
fix(openai): bound openai_stream_events streaming accumulation memory (#556)

### LINKED ISSUE #556: Streaming Responses accumulation has no aggregate byte ceiling
## Clawpatch finding

- **Severity:** high
- **Category:** security
- **Confidence:** high
- **Feature(s):** `feat_custom_openai_responses_state_store`
- **Finding ID(s):** `fnd_sig-feat-custom-openai-responses_6072107102`
- **Run ID(s):** `20260727T071056-9838c7`
- **Prior-run overlap:** Overlaps prior finding fnd_sig-feat-custom-openai-responses_b7ced938ce, but no existing exact issue was found.

## Evidence

Finding `fnd_sig-feat-custom-openai-responses_6072107102` (high/security/high):
- `apis/src/openai/responses/stream_events/config.rs:25-28` (`StreamEventsConfig::max_events`) — pub max_events: Option<usize>
- `apis/src/openai/responses/stream_events/config.rs:54-56` (`StreamEventsConfig::validate`) — reject_zero(v, "max_events")?
- `apis/src/openai/responses/stream_events/mod.rs:45-64` (`StreamEventsState`) — tool_call_args: std::collections::HashMap<String, String>
- `apis/src/openai/responses/stream_events/accumulator.rs:134-171` (`handle_output_item_added / handle_output_item_done`) — state.output_items_mut().push(item.clone())

## Description

- **Streaming accumulation has no aggregate memory ceiling**: The parser bounds an incomplete frame and optionally counts events, but accumulated output items and the collection of per-tool argument buffers have no shared byte or item ceiling. The default permits 100,000 events, and configured max_events has no upper bound. A backend can send many individually valid events that are cloned into request-scoped state until the process exhausts memory. This likely overlaps the previously known SSE aggregate-memory issue.

## Reproduction

- **Streaming accumulation has no aggregate memory ceiling**: Use the default configuration and stream many response.output_item.added events, each containing a large but individually valid item below max_buffer_bytes. The filter appends every cloned item to ResponsesState.output without releasing accumulated state; process memory grows with the entire stream.

## Recommendation

- **Streaming accumulation has no aggregate memory ceiling**: Add a mandatory aggregate accumulation budget covering output items, tool_calls, tool_call_args, usage, and response_object. Track bytes incrementally and mark the stream invalid once either the aggregate byte cap or an item-count cap is exceeded. Put a safe upper bound on max_events.

## Notes

Generated by scheduled Clawpatch Codex/xhigh run over `apis/src` excluding `apis/src/token_usage/**` after rebasing on `origin/main`. Please validate before assigning or starting implementation.

leseb:
## Summary

The `openai_stream_events` filter accumulated streaming SSE state — the response output-item list, per-tool-call argument buffers, the local-tool `emitted_output_items` map, and usage — with no aggregate ceiling, so a malicious or compromised backend could stream many individually-valid events until process memory was exhausted. This adds a mandatory aggregate accumulation budget (default 64 MiB of bytes and 100,000 output items, applied even when unconfigured) that charges each event in the parse phase so a chunk aborts atomically before mutating shared state, stays sticky so a poisoned stream cannot be resurrected by a later terminal event, and fails closed (`skip_persist` + `stream_parse_error` metadata and a terminal SSE error event) on overflow. Charged dimensions cover `output_item.added`/`done`, `function_call_arguments.delta`/`done`, and local-tool progress events (`response.web_search_call.*`, `response.mcp_call.*`, `response.mcp_list_tools.*`), and `max_events` now has a hard ceiling as well.

## Related issue

Closes #556

## Validation

- [x] Unit tests — `cargo test -p praxis-ai-apis` (3427 pass, incl. 7 new accumulation tests covering byte/count overflow, poisoned-stream stickiness, local-tool-progress and tool-call-argument charging, `output_item.done` count charging, and the mandatory default)
- [x] `make build`
- [x] `make lint`
- [x] `make doc`

## Checklist

## Breaking changes

The budget is mandatory, so it applies with default caps even when unconfigured; well-behaved streams are unaffected. Configs may optionally set `max_accumulated_bytes` and `max_output_items`, and values that are zero or above their ceilings are rejected at build time.


apis/src/openai/responses/stream_events/mod.rs
```diff
@@ -489,11 +536,137 @@ fn parse_chunk_events(
...
+    ctx: &mut HttpFilterContext<'_>,
+    event: &ResponsesEvent,
+    frame: &SseFrame,
+) -> Result<(), SseParseError> {
+    if !charges_accumulation_bytes(event) {
+        return Ok(());
+    }
+    let accumulated_bytes = {
+        let responses = ctx.extensions.get_or_insert_with(ResponsesState::default);
+        responses.stream_accumulated_bytes = responses.stream_accumulated_bytes.saturating_add(frame.data.len());
```
franciscojavierarceo, line 598: charging only the current frame misses copies of previously retained data. a function-call item without `id`/`call_id` is accepted via `output_index`; each small `function_call_arguments.done` clones the whole item, and `upsert_tool_call` appends another copy because neither identifier can match. with a 64 KiB cap, a 10,000-byte function name and 400 done events retained 4,028,800 bytes of tool calls while this counter was only 42,911 and the stream remained accepted. this can scale to memory exhaustion despite the new limit. we need to charge that retained growth or reject/deduplicate these calls before cloning.

apis/src/openai/responses/stream_events/mod.rs
```diff
@@ -463,9 +501,17 @@ fn parse_and_accumulate(
...
     // recovery still re-synthesizes the executed tool items (#276 finding 3).
-    let events = parse_chunk_events(state, &frames, now)?;
+    let events = parse_chunk_events(state, ctx, &frames, now)?;
     let logical_output = commit_chunk_events(state, ctx, events);
 
+    // The retained item count is a property of the committed output that only
+    // exists after `commit_chunk_events` grows it, so it is enforced here rather
+    // than per-event in phase 1. Any overshoot is bounded to a single chunk and
+    // the request-wide budget stays sticky for every later chunk and round.
+    if let Some(error) = accumulation_count_exceeded(state, ctx) {
```
franciscojavierarceo: this check runs after `commit_chunk_events` has flushed pending local tools and marked their events delivered. when the count overflows, `handle_parse_error` discards those bytes, but EOS recovery trusts the committed milestones and skips the executed tool. with a cap of 1, one previously executed MCP call, and the next model `output_item.added`, i reproduced a terminal error with no client-visible record of the executed call. we should validate the count before committing delivery state, or roll back those milestones when rejecting the chunk.

franciscojavierarceo (changes requested):
two issues reproduced locally: the byte budget misses retained tool-call copies, and the post-commit count check can hide already-executed tool activity. the existing accumulation tests pass, but the focused regressions below fail.

apis/src/openai/responses/stream_events/mod.rs
leseb: c34c1d0f41eafd94f47c6ccff86a223cbf4792dd

apis/src/openai/responses/stream_events/mod.rs
leseb: c34c1d0f41eafd94f47c6ccff86a223cbf4792dd

apis/src/openai/responses/stream_events/mod.rs
```diff
@@ -489,23 +535,211 @@ fn parse_chunk_events(
...
+/// item was added in the same, not-yet-committed chunk it is not found here, but that
+/// chunk's own `output_item.added` frame already charged the item once, bounding the
+/// clone to that chunk's wire bytes.
+fn tool_call_clone_bytes(ctx: &HttpFilterContext<'_>, event: &ResponsesEvent) -> usize {
+    let ResponsesEvent::FunctionCallArgumentsDone(payload) = event else {
+        return 0;
+    };
+    ctx.extensions
+        .get::<ResponsesState>()
+        .and_then(|responses| find_output_item(responses.output_items(), payload))
```
franciscojavierarceo: this lookup still misses the item when `output_item.added` and repeated `function_call_arguments.done` events arrive in the same chunk, since phase 1 has not accumulated any of them yet. charging the added frame once does not cover all the later copies. i coalesced the previous reproduction into one chunk: an ID-less function call with a 10,000-byte name followed by 400 done events targeting `output_index: 0`. with a 64 KiB cap, the stream still accepts 4,028,800 bytes of retained tool calls while charging only 42,911. we need to account for same-chunk items and repeated completions before retaining those copies, and cover this coalesced-chunk case in the regression test.

franciscojavierarceo (changes requested):
the original separate-chunk clone reproduction now passes, and the executed-tool recovery tests confirm the delivery-milestone fix. the byte-budget issue still reproduces when the added item and repeated done events arrive in one chunk; details inline.

apis/src/openai/responses/stream_events/mod.rs
leseb: b0e9af1f13490093343a9ee773b7cafd55c99972

apis/src/openai/responses/stream_events/mod.rs
```diff
@@ -489,23 +535,276 @@ fn parse_chunk_events(
...
+        return Some(item);
+    }
+
+    let output_index = done_payload.get("output_index").and_then(Value::as_u64)?;
+    chunk_events
+        .iter()
+        .rev()
+        .filter_map(chunk_event_output_item)
+        .find_map(|(payload, item)| {
+            (payload.get("output_index").and_then(Value::as_u64) == Some(output_index)).then_some(item)
```
franciscojavierarceo: the pending lookup uses the advertised `output_index`, but `handle_output_item_added` ignores that field and pushes the item onto the output array. these can resolve different items. changing only the added event's index from 0 to 7 in the same-chunk reproduction leaves the large item at actual index 0, so 400 arguments-done events targeting index 0 still clone it while this lookup finds nothing. with a 64 KiB cap, i again get 4,028,800 retained bytes versus 42,911 charged, without rejection. since the budget protects against a malicious backend, we need pending resolution to follow the accumulator's actual append/replace rules, or reject inconsistent indices before committing any copies.

apis/src/openai/responses/stream_events/mod.rs
```diff
@@ -489,23 +535,276 @@ fn parse_chunk_events(
...
+/// resolving it here lets [`tool_call_clone_bytes`] size the clone the commit phase is
+/// about to retain. The most recent matching event wins, matching the last-write-wins
+/// state a replayed `added`/`done` sequence leaves behind.
+fn pending_chunk_output_item<'a>(chunk_events: &'a [ResponsesEvent], done_payload: &Value) -> Option<&'a Value> {
+    let matches_item_id = |item: &Value, item_id: &str| item.get("id").and_then(Value::as_str) == Some(item_id);
+    if let Some(item_id) = done_payload.get("item_id").and_then(Value::as_str)
+        && let Some(item) = chunk_events
+            .iter()
+            .rev()
+            .filter_map(chunk_event_output_item)
```
franciscojavierarceo: each arguments-done event now scans all prior events, twice when both `item_id` and `output_index` are present and unmatched. a coalesced chunk of N such events takes O(N²) work even with no output items. an optimized benchmark of these exact lookup helpers took about 0.05s for 10,000 events, 0.8s for 40,000, and 1.9s for 60,000; the latter charges only 3.24 MB and stays below the default event cap. this work runs synchronously in response-body processing. we should maintain the pending item state incrementally during parsing instead of rescanning the event history for every completion.

franciscojavierarceo (changes requested):
the original coalesced-chunk regression now passes. the pending lookup still disagrees with the accumulator when backend indices are inconsistent, and its repeated history scans introduce quadratic work. reproduced both against this revision; details inline.

apis/src/openai/responses/stream_events/mod.rs
leseb: 434571aeeabcff2dc4e838006ea64f8df57d78e7

apis/src/openai/responses/stream_events/mod.rs
leseb: 434571aeeabcff2dc4e838006ea64f8df57d78e7

franciscojavierarceo (approved):
the current head addresses the prior accumulation findings: the request-wide byte and item guards persist across IRR rounds, retained tool-call clones are charged at the point of retention, and count overflow is checked before delivery milestones are committed. the 129 focused `openai::responses::stream_events` tests pass, including the cross-round, coalesced-clone, mismatched-index, terminal-size, and milestone-atomicity regressions. the added integration coverage uses deterministic synthetic SSE and validates SQLite persistence; no external inference recording is warranted here.

PR #579, opened 2026-07-27
fix(mcp_dispatch): match tool calls by encoded function name

leseb:
### What does this PR do?

The MCP tool resolver encodes function names as `{label}__{tool_name}` for
inference, but the dispatcher compared tool calls from inference against the
raw tool name in the map key — which never matched. This PR changes all
dispatcher lookups to compute `encode_function_name(label, tool_name)` from
each map entry and compare against the encoded name returned by inference.
Lookup functions now return both key and value so callers can extract the
original tool name for MCP server calls, approval policy checks, and result
construction.

### Which issue(s) does this relate to?

Fixes #544

### Checklist

- [x] Signed off all commits (`git commit -s`)
- [x] Tests added or updated
- [ ] Documentation updated (if applicable)
- [x] `make lint && make test` passes locally

### Does this introduce a breaking change?

No.

PR #1161, opened 2026-09-14
Guardrails: migrate guardrails to the v1/checks endpoint

### LINKED ISSUE #1044: feat(ai_guardrails) - Migrate NeMo provider to `/v1/checks`
## Description

The `ai_guardrails` filter's NeMo provider currently calls
`/v1/guardrail/checks`. That endpoint evaluates all messages in one HTTP
call and returns a per-rail breakdown (`rails_status`, `guardrails_data`),
but it does not expose redaction: NeMo collapses `MODIFIED` into
`success`, so the proxy cannot tell whether content passed unchanged or
was sanitized.

NeMo's `/v1/checks` endpoint is the supported path for content
transformation. It returns:

```json
{
  "status": "passed | blocked | modified",
  "content": "text after rails processing",
  "rail": "blocking rail name or null"
}
```

This migration updates the NeMo provider to call `/v1/checks`, parse the
new response shape, and preserve full-conversation coverage by issuing one
HTTP request per user message (cumulative message slices).

**This issue covers the endpoint migration only.** Writing `content` back
into request/response bodies (redaction enforcement) is tracked separately
in [#49](https://github.com/praxis-proxy/ai/issues/49).

Related context:
- Prior discussion (closed): [praxis-proxy#1061](https://github.com/orgs/praxis-proxy/discussions/1061)
- Nvidia Upstream feat request (to support evaluation for all messeges in one call): [NVIDIA-NeMo/Guardrails#2277](https://github.com/NVIDIA-NeMo/Guardrails/issues/2277)

## Motivation

`ai_guardrails` was designed for three verdicts: pass, block, and redact.
The current NeMo integration cannot populate `GuardResult::Redact` because
`/v1/guardrail/checks` never returns `"modified"`.

Operators who configure NeMo with PII-masking rails may believe content is
safe when the proxy records `"passed"` and forwards the original body.
Fixing redaction requires the `/v1/checks` response shape first.

`/v1/guardrail/checks` is also expected to be deprecated upstream. Aligning
now avoids building on a legacy API surface.

### Cost of the N-call workaround

`/v1/checks` evaluates only the last message of the relevant role per
call. To keep checking every user turn, the provider loops with cumulative
slices and fails fast on the first `blocked` result.

Benchmarks I made (on odh/ai-gateway-payload-processing)
(TrustyAI NeMo-Guardrails, PII rails, ~1 ms pod-to-pod RTT):

| User messages | `/v1/guardrail/checks` (1 call) | `/v1/checks` (N calls) | Overhead |
|---|---|---|---|
| 1 | 39 ms | 26 ms | ~0% |
| 5 | 142 ms | 133 ms | ~0% |
| 10 | 223 ms | 309 ms | **+39%** |
| 20 | 437 ms | 597 ms | **+37% (~160 ms)** |

Overhead is on the guardrail hop only, before the upstream model is
reached. Short conversations (1-5 turns) are effectively unchanged.

## Proposed changes

- Update docs, examples, and default endpoint references from
  `/v1/guardrail/checks` to `/v1/checks`.
- Replace `NemoResponse` parsing: `passed` / `blocked` / `modified` +
  `content` + `rail` (treat `content` as optional with a default).
- Map statuses to `GuardResult`:
  - `passed` → `Pass`
  - `blocked` → `Block { reason: rail }`
  - `modified` → `Redact` (record `"redacted"` in filter results; forward
    body unchanged until #49)
- Loop one POST per user message (request phase) and per assistant message
  (response phase) using cumulative slices.
- Handle provider failures via HTTP status codes (no in-body `"error"`).
- Update unit tests, integration tests, and example configs.

## Out of scope

- Applying `content` back into request/response JSON (redaction
  enforcement) — [#49](https://github.com/praxis-proxy/ai/issues/49).

## Alternatives considered

**Stay on `/v1/guardrail/checks`.**
No migration cost; all messages checked in one call. Redaction remains
impossible without scraping `guardrails_data.log` internals. Upstream
deprecation risk.

**Move to `/v1/checks` but only check the last user message.**
One HTTP call, simplest implementation. Earlier turns are not evaluated.
Likely unacceptable for security use cases.

**Wait for upstream multi-message `/v1/checks`.**
Best long-term latency, but blocks redaction work indefinitely.

## Acceptance criteria

- [ ] `NemoProvider` calls `/v1/checks` and parses the new response
      schema.
- [ ] Every user message in a request body is evaluated (N-call loop).
- [ ] `blocked` returns 403 on request phase; `passed` forwards to
      upstream; `modified` records `redacted` but does not rewrite the
      body.
- [ ] Example configs, generated filter docs, and tests updated.
- [ ] Breaking change documented: operators must update the `endpoint` URL
      in YAML.

liavweiss:
## Summary
- Migrate the `ai_guardrails` NeMo provider from `/v1/guardrail/checks` to `/v1/checks`
- Parse the new response shape (`passed` / `blocked` / `modified` + `content` + `rail`)
- Evaluate each user/assistant message with cumulative context (one NeMo call per message)
- Map provider HTTP errors to fail-closed behavior (replaces in-body `"error"` status)

## Breaking change
NeMo `endpoint` URLs in proxy config must use `/v1/checks` instead of `/v1/guardrail/checks`.

## Out of scope
- Applying `modified`/`content` redaction to request/response bodies (follow-up: #49)

## Test plan
- [x] `cargo test -p praxis-ai-filters guardrails` (70 tests)
- [x] `cargo test -p praxis-tests-integration guardrails` (35 tests)
- [x] A real nemo test with a real v1/chat/completion passing forward the ai_guardrails filter.

close: #1044


leseb (changes requested):
P1 – modified short-circuits evaluation. A modified earlier message causes later blocked messages to go unchecked and pass through.

P1 – wrong NeMo rail direction. Mixed user/assistant history auto-runs both rail types; request-phase modifications can be reported as passed. Send phase-specific rail_types.

P1 – unbounded callout amplification. One request can trigger arbitrarily many sequential NeMo calls, each with its own timeout. Add a call limit or overall deadline.

filters/src/guardrails/providers/mod.rs
```diff
@@ -55,10 +55,6 @@ pub enum GuardResult {
         reason: String,
     },
     /// Content contains sensitive data — forward with masked text.
-    #[cfg_attr(
```
christinaexyou, line 58: can we leave this comment here until #673 is merged ? while we can now obtain `modified_text` we still don't forward it

filters/src/guardrails/providers/nemo.rs
```diff
@@ -148,10 +144,49 @@ impl GuardProvider for NemoProvider {
...
+    let target_role = match phase {
+        GuardPhase::Request => "user",
+        GuardPhase::Response => "assistant",
+    };
+
+    messages
+        .iter()
+        .enumerate()
+        .filter(|(_, message)| message.get("role").and_then(|role| role.as_str()) == Some(target_role))
+        .filter_map(|(index, _)| messages.get(..=index).map(<[serde_json::Value]>::to_vec))
```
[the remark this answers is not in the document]
christinaexyou: i concur - this is a glaring opportunity for a DOS attack. can we run slices concurrently and do an early exit when a guardrail is triggered and/or add a limit to the number of slices ?

filters/src/guardrails/providers/mod.rs
liavweiss: The `dead_code` attribute can't be restored - Redact is now constructed when NeMo returns modified, so the lint expectation would be unfulfilled. The doc comment on lines 59-60 serves the same intent: modified_text is populated from NeMo but not applied to the body until #49.

liavweiss:
> 

> P1 – modified short-circuits evaluation. A modified earlier message causes later blocked messages to go unchecked and pass through.
> 
Addressed - modified no longer short-circuits the loop; only blocked fails fast. See apply_slice_result.

> P1 – wrong NeMo rail direction. Mixed user/assistant history auto-runs both rail types; request-phase modifications can be reported as passed. Send phase-specific rail_types.
> 
Addressed - we now send explicit rail_types (["input"] for request phase, ["output"] for response phase) to prevent NeMo from auto-running both rail families on mixed-role slices.
> P1 – unbounded callout amplification. One request can trigger arbitrarily many sequential NeMo calls, each with its own timeout. Add a call limit or overall deadline.
>
Addressed - same max_message_checks limit as item 1.

leseb:
@liavweiss 

P1: timeout_ms restarts for each of up to 32 sequential calls, permitting roughly 320 seconds total by default. Use one overall evaluation deadline.
P2: Valid NeMo modified responses with empty content are rejected. Empty-string redaction must be accepted; only a missing field should fail.
P2 docs: Document the NeMo v0.24+ requirement and the new max_message_checks: 32 fail-closed limit.
P3: Example and PR description still reference #579; the correct redaction issue is #49.

liavweiss:
> @liavweiss
> 
> P1: timeout_ms restarts for each of up to 32 sequential calls, permitting roughly 320 seconds total by default. Use one overall evaluation deadline. 
P2: Valid NeMo modified responses with empty content are rejected. Empty-string redaction must be accepted; only a missing field should fail. 
P2 docs: Document the NeMo v0.24+ requirement and the new max_message_checks: 32 fail-closed limit. 
P3: Example and PR description still reference #579; the correct redaction issue is #49.

Done. 
Regarding the "only a missing field should fail." - nemo will always return the `content` field.

leseb (approved):

leseb:
@liavweiss lint is failing

PR #1172, opened 2026-09-14
feat(routing): add apikey credential injection strategy

### LINKED ISSUE #1171: routing: credential strategy enum only accepts bearer_token, blocking apiKey-authenticated providers
## Problem

The routing overlay wire (`candidates[].credential.strategy`, validated by `routing::descriptor` and `provider_route`) and `credential_inject` accept exactly one strategy value: `bearer_token`. Producers that map provider authentication types onto this enum cannot express the very common `apikey` shape — raw token in a dedicated header (e.g. `x-api-key`) instead of `Authorization: Bearer`.

Concretely, the Kubernetes control plane in [opendatahub-io/ai-gateway-controller](https://github.com/opendatahub-io/ai-gateway-controller) maps `InferenceProvider` CRD `auth.type` values to overlay strategies as part of the IPP migration (epic: opendatahub-io/ai-gateway-controller#10, item O1). Today its `strategyFor()` renders no credential at all for every real CRD auth type, because praxis-ai only accepts the literal `bearer_token`. apiKey is the dominant real-world provider auth shape; sigv4 and oauth2 are tracked separately upstream and should remain honestly unrepresentable.

## Expected behavior

- Overlay `credential.strategy` accepts `bearer_token` (existing behavior, byte-compatible) and `apikey`; unknown values keep failing validation.
- `credential_inject` gains an `apikey` strategy: injects the raw resolved token into a per-entry optional `header` (default `x-api-key`), rejected at config time for `bearer_token` entries, and rejected for reserved internal header prefixes (`x-praxis-`, `x-mcp-`) which the upstream boundary strips in flight.
- Request path cross-checks the strategy declared by the overlay candidate against the configured entry strategy and fails closed (503) on mismatch, so a producer/config disagreement can never send a bearer token where an apikey header is expected.
- Existing guarantees hold unchanged for both strategies: `Zeroizing` storage, tokens never written to metadata/traces/error bodies, caller `Authorization`/`x-api-key` (plus the configured injection header) stripped at the provider boundary, and projected-file rotation without restart.

## Context

Claimed and implemented against the decision recorded in the O1 thread of opendatahub-io/ai-gateway-controller#10: widen the praxis enum, backward compatible, no auto-migration of existing overlays.

yossiovadia:
## Summary

Widens the routing credential strategy enum beyond `bearer_token` with an `apikey` strategy that injects the raw projected token into a per-entry configurable header (default `x-api-key`). This is the data-plane half of item **O1** in the IPP migration epic ([opendatahub-io/ai-gateway-controller#10](https://github.com/opendatahub-io/ai-gateway-controller/issues/10)): the control plane's `strategyFor()` currently renders no credential for any real CRD auth type because this enum accepted only the literal `bearer_token`.

The smallest complete change is:

- `routing::metadata` gains `STRATEGY_APIKEY` and a shared `is_supported_strategy()` so the overlay descriptor, `provider_route`, and `credential_inject` validate against one closed set.
- `credential_inject` gains the `apikey` strategy and an optional `header` field (apikey-only; rejected at config time for `bearer_token` entries, for reserved internal prefixes `x-praxis-`/`x-mcp-` which the upstream boundary would strip in flight, and for `authorization`, which stays owned by the bearer strategy).
- The request path cross-checks the strategy declared by the overlay candidate against the configured entry strategy and fails closed (503) on mismatch, so a producer/config disagreement can never send a bearer token where an apikey header is expected.
- The configured injection header joins `Authorization`/`x-api-key` in the caller-header removal list, so a caller-supplied value in a custom apikey header cannot ride alongside the injected credential.
- File-backed entries carry their strategy and header through the rotation watcher, so projected-Secret reloads rebuild the same header form without a restart.

Existing guarantees hold unchanged for both strategies: `Zeroizing` storage, tokens never written to metadata/traces/error bodies, bounded inputs, and fail-closed lookup. `sigv4` and `oauth2` remain intentionally unrepresentable and stay tracked upstream.

Wire/digest note: this widens only the accepted value set — no existing overlay bytes change, so Go↔Rust golden vectors and content-addressed digests for `bearer_token` producers are unaffected.

## Related issue

Closes #1171

## Validation

- [x] Unit tests — `cargo test -p praxis-ai-filters routing::` → 378 passed (new: strategy set closure, descriptor/provider_route accept+reject tables, config accepts/rejects for `header`, default+custom apikey injection, caller-header stripping including the configured header, strategy-mismatch fail-closed in both directions, apikey file source + rotation, token-never-in-metadata for apikey)
- [x] Integration or functional tests — `cargo test -p praxis-tests-integration --test suite examples::provider_route` → 3 passed; the example test now asserts live apikey injection through the mTLS provider listener (`x-api-key` reaches the backend, caller `Authorization` stripped end-to-end)
- [x] `make lint` — component-equivalent: `cargo clippy --all-targets -- -D warnings` on touched crates, `cargo +nightly fmt --all -- --check`, `cargo xtask lint-separators` / `lint-filter-docs` / `lint-example-tests` / `lint-markdown-links` / `sync-example-readme`, plus `RUSTDOCFLAGS="-D warnings" cargo doc`

Environment note: the file-watcher rotation tests added by #1136 do not receive `notify` events under macOS `/private/var/folders` (they fail on untouched `main` on that machine too); they pass locally with `TMPDIR` pointed at a regular volume and natively under Linux CI.

## Checklist

- [x] New capabilities include an example config and functional example test. (`provider-route.yaml` gained an apikey route + entry; the example integration test exercises it.)
- [x] User-facing behavior and generated documentation are updated. (`docs/filters/credential_inject.md` regenerated; `examples/README.md` verified in sync.)
- [ ] Performance-sensitive changes include appropriate benchmark or load-test evidence. (N/A — request path adds one string compare and one `HeaderName` clone against the existing injection path.)

## Breaking changes

None. `bearer_token` remains the default strategy, existing overlay bytes validate and hash identically, and configs without `header` behave exactly as before. `apikey` is opt-in per credential entry.

Follow-up (control plane, other repo): `ai-gateway-controller` `strategyFor()` can now map `auth.type: apiKey` → `apikey` once this lands; ADR O1 closure after merge.


yossiovadia:
Recording the planned control-plane follow-up so the sequencing is explicit.

The `strategyFor()` half of O1 in the control plane (ai-gateway-controller `pkg/envelope`) is pre-drafted locally on branch `feat/envelope-apikey-strategy`: `auth.type: apikey` → `"apikey"`, exact wire-shape tests for both strategies, credential-omission list narrowed to sigv4/oauth2. Full Go suite and lint are green.

Dependencies for opening that PR:

- **It will not be landed before this PR merges.** The Go change makes the resolver emit an `apikey` credential for every `apiKey` provider, and today's data plane rejects an unaccepted strategy by failing the whole overlay (`validate_credential` in `descriptor.rs`) — the control plane must not run ahead of the consumer. Once this PR is in main, the emitted bytes validate against `is_supported_strategy` as-is.
- The apikey header choice (`x-api-key` default, or a per-entry custom header) stays gateway-side `credential_inject` config — the overlay wire keeps carrying only `{strategy, secretRef}`.
- sigv4 and oauth2 remain unmapped on both sides (faithful representation is still an open §6 item upstream); those routes keep rendering no credential.

yossiovadia:
@nerdalert friendly ping for a review pass when you have a moment. Since the work landed: rebased onto main (picks up the rustls 0.23.45 bump), all 19 checks green, and the CodeQL cleartext-logging annotations on the new Anthropic-path test were addressed with role-based renames rather than suppressions. Happy to walk through any of the fail-closed paths if that's quicker than reading.

filters/src/routing/credential_inject.rs
```diff
@@ -412,45 +477,135 @@ fn selected_credential_ref(ctx: &HttpFilterContext<'_>) -> Option<Result<Credent
...
+        Some(header) => {
+            validate_bounded("header", header, MAX_SOURCE_LEN)?;
+            let name: HeaderName = header.parse().map_err(|error| -> FilterError {
+                format!(
+                    "credential_inject: invalid header '{header}' for '{}/{}/{}': {error}",
+                    entry.name, entry.namespace, entry.key
+                )
+                .into()
+            })?;
+            if RESERVED_HEADER_PREFIXES
```
nerdalert, line 564: P2: Reject transport-controlled headers before accepting the API-key target.

`HeaderName::parse` only establishes syntactic validity, so this currently accepts `host`, `content-length`, `connection`, `transfer-encoding`, and `proxy-authorization`. Injecting the provider credential into one of those fields can corrupt authority selection or HTTP framing, and `proxy-authorization` sends it to an intermediary rather than the selected provider.

The repository already has the required policy in `praxis_ai_apis::promotion::is_transport_controlled_header_lowercase`, including proxy authorization through the hop-by-hop check. Please reuse it here and add table-driven rejection tests for at least `host`, `content-length`, `connection`, and `proxy-authorization`.

Suggested validation after parsing `name`:

```suggestion
            if praxis_ai_apis::promotion::is_transport_controlled_header_lowercase(name.as_str()) {
                return Err(format!(
                    "credential_inject: header '{header}' for '{}/{}/{}' must not be transport-controlled",
                    entry.name, entry.namespace, entry.key
                )
                .into());
            }
            if RESERVED_HEADER_PREFIXES
                .iter()
                .any(|prefix| name.as_str().starts_with(prefix))
            {
```

nerdalert (changes requested):
The API-key strategy and fail-closed strategy matching look sound overall. One custom-header validation gap needs to be closed before merge.

filters/src/routing/credential_inject.rs
yossiovadia: Good catch — fixed in e30bc11d. Reused `praxis_ai_apis::promotion::is_transport_controlled_header_lowercase` right after the reserved-prefix and `authorization` guards in `resolve_header_name`, so config-time acceptance now rejects every name the transport owns (host, content-length, connection, transfer-encoding, proxy-authorization, upgrade, te, keep-alive) via the shared policy the promotion boundary already uses — the two surfaces cannot drift. Added `apikey_transport_controlled_header_rejected` covering host, content-length, connection, proxy-authorization, and transfer-encoding, and extended the field/function docs to state the rule. 44/44 module tests, clippy and rustfmt clean.

nerdalert:
No new blockers found on the current PR head (7aaf27a).

The prior transport-controlled header issue is fixed correctly using the shared policy, with suitable coverage. Credential injection remains fail-closed, rotation preserves strategy/header configuration, caller credential headers are stripped, and unsupported strategies remain rejected.

Waiting on tests then +1

nerdalert (approved):

noyitz:
Re-review landed — the fix for the change-request feedback was pushed on Sep 14 and @nerdalert approved today. All checks are green (lint, test, coverage, MSRV, security audit, DCO, signoff). @jland-redhat could you take a look when you get a chance, or merge if you're happy with the shape?

jland-redhat (approved):

PR #1068, opened 2026-09-11
feat(routing): add static weighted provider selection

nerdalert:
## Summary

Add opt-in static weighted provider selection to the Praxis AI routing overlay.

- add `weightedRandom` selection within the first viable selection group
- accept explicit per-candidate `traffic_weight` values from `1..=1000`
- preserve eligibility, group precedence, and session-affinity behavior
- validate that weights are present only for weighted selection
- reject missing, zero, oversized, or conflicting weighted configuration
- reject duplicate effective stable IDs
- use checked cumulative `u64` weight bounds and an unbiased random draw
- preserve the existing empty-overlay last-known-good behavior

This is static configuration, not metric-driven routing. Grid determines candidate eligibility and grouping before Praxis performs the weighted draw.

## Why

Grid can determine that several providers are equally eligible while those providers have different configured capacities. Existing deterministic, round-robin, and random selection cannot express that capacity ratio. This change lets a trusted routing overlay ask Praxis to distribute new requests proportionally within the first eligible group, without weakening health, policy, locality, admission, or affinity precedence.

Keeping the weighted draw in Praxis also keeps request selection local to the gateway's accepted in-memory snapshot. Requests do not wait for Grid, Kubernetes, metrics collection, or a cross-site control-plane call.

## Related issue

Closes #1042

## API

```yaml
selection_policy:
  mode: weightedRandom
candidates:
  - cluster: provider-a
    selection_group: 0
    traffic_weight: 50
  - cluster: provider-b
    selection_group: 0
    traffic_weight: 30
```

Weights are relative. `50/30/20` and `5/3/2` express the same intended proportions.

## Dependency

The companion [praxis-proxy/grid#142](https://github.com/praxis-proxy/grid/pull/142) change is merged. It publishes validated direct weights in the same `1..=1000` range, propagates provider capacity across sites, and provides the Helm-backed Kind qualification. This AI support must be available before a Grid configuration enables `weightedRandom`.

## Validation

- [x] Unit tests
- [x] Integration or functional tests
- [x] `make lint`

Completed validation:

- AI workspace tests: passed
- focused routing tests: 378 passed
- Clippy with warnings denied: passed
- documentation and lint checks: passed
- two companion Grid Kind qualifications passed all baseline, changed-weight, equal-weight, convergence, attribution, gateway-stability, and cleanup assertions
- follow-up overlay validation: 111 tests passed

[...]

PR #1173, opened 2026-09-14
feat(routing): add apikey credential strategy to overlay candidates

### LINKED ISSUE #1041: Routing-overlay credential strategy vocabulary: wire accepts only bearer_token; control-plane auth types are apikey|sigv4|oauth2
## The gap

The routing-overlay envelope (contract v1) carries per-candidate credential **references** — `{strategy, secretRef{name, namespace, key}}` — never secret bytes. Validation and `credential_injection` accept exactly one strategy value today: `bearer_token` (anything else is rejected at load; `credential_injection` fails closed with 503 on a non-`bearer_token` strategy).

The producer side (`ExternalProvider` auth in the MaaS control plane) is typed `apikey | sigv4 | oauth2` — **zero overlap** with the wire vocabulary. Consequence: the control-plane renderer currently omits the `credential` field entirely for every auth type. Routing works, and provider-gateway-side auth still applies, but per-route credential injection is unexpressible on the wire — a feature regression versus IPP's `apikey-injection` (which honored the merged-config auth contract: per-provider header + prefix, model>provider>built-in precedence).

## Direction (as agreed with control-plane owners)

Widen the strategy enum end-to-end rather than moving credentials out of the envelope, so the frozen CRD auth types stay expressible with no CRD migration:

- `apikey` — inject the secret value into a configurable header with an optional prefix (IPP parity: Anthropic `x-api-key`, Azure `api-key`, default `Authorization: Bearer <value>`); header/prefix may be overridden per candidate by the renderer.
- `sigv4` — interplay with `aws_sigv4_sign` (which today sources static env creds, #813); decide whether the overlay ref feeds it file-mounted creds.
- `oauth2` — interplay with `azure_ad` / `gcp_adc`; token exchange needs more than a static secret ref — scope may shrink to "renderer pre-resolves tokens into a rotated secret file, wire keeps `apikey`".

Until the enum lands, renderers **omit** the credential for unrepresentable types rather than emit a strategy the consumer will reject at cold start (no wire lie).

## Acceptance criteria

- Overlay validation + `credential_injection` accept the widened set; unknown strategies still fail closed.
- Golden-vector fixtures under `tests/fixtures/overlay-contract/v1` extended with accept/reject cases for the new values, so both the Rust consumer and the Go producer CIs pin the same vocabulary.
- Docs (filter reference) updated.

Decision requested from maintainers: **widen the enum** (preferred, backward-compatible) **vs. drop `credential` from the envelope** entirely.

noyitz:
## Summary

Widens the routing-overlay credential strategy vocabulary from `bearer_token`-only to **`bearer_token | apikey`**, givingIPP `apikey-injection` parity for providers that authenticate with a custom header (Anthropic `x-api-key`, Azure `api-key`, raw-token schemes).

Closes #1041

## Design

- **Envelope-carried overrides.** `apikey` candidates may carry optional `header`/`prefix` fields in the overlay envelope (additive, `deny_unknown_fields` preserved). Because they are part of the candidate object, they are covered by the content digest — the same digest-pinned semantics as `traffic_weight` in #1068. Absent overrides default to `Authorization` + `Bearer ` prefix, so plain `apikey` reproduces the bearer shape.
- **Config cross-check.** `credential_inject` entries keep their configured `strategy` as an authority: if the overlay-selected strategy disagrees with the matched config entry, the request fails closed with 503.
- **Fail-closed everywhere.** Unknown strategies (`oauth2`, `sigv4`, case variants) and `header`/`prefix` overrides on `bearer_token` are rejected at snapshot load (hot reload retains last-known-good) *and* at injection time. Malformed overrides (invalid header name, prefix with whitespace/control bytes or > 256 bytes) fail closed per request.
- **Secret hygiene unchanged.** Token bytes stay in `Zeroizing` snapshots, never in `filter_metadata`, traces, or error bodies. Header assembly happens per request from the zeroized token; the caller's `Authorization`, `x-api-key`, and any configured custom target header are stripped before injection.

## Tests / fixtures

- Golden vectors under `tests/fixtures/overlay-contract/v1/` (driven by `manifest.json`): `valid-apikey-credential.json` (accept, with pinned content digest + revision), `credential-unknown-strategy.json` (reject, `unsupported_credential_strategy`), `credential-bearer-with-header.json` (reject, `invalid_credential_combination`).
- Unit coverage across `descriptor`, `overlay`, `metadata`, and `credential_inject`: strategy vocabulary, override validation, metadata round-trip, injection shape for header/prefix/raw-token variants, override stripping, strategy mismatch, and post-injection metadata hygiene.
- `cargo test -p praxis-ai-filters routing::`: 373 passed; the two `projected_file_*` watcher tests are pre-existing macOS file-watch flakes (they fail identically on pristine `origin/main`).

Docs regenerated via `cargo xtask generate-filter-docs`; `make lint` clean.


yossiovadia:
Closing in favor of #1172 — same goal (overlay apikey vocabulary with fail-closed strategy handling), and #1172 is further along with the maintainer: Brent reviewed it today and the single gap he flagged (transport-controlled headers as injection targets) is already fixed and back in his court.

The per-candidate header/prefix override design here doesn't match a live requirement — every provider in scope (Anthropic x-api-key, Azure api-key) needs exactly one fixed header per config entry, so the envelope-carried flexibility is complexity without a customer today. If a real multi-header need ever appears, this PR's digest-covered override pattern is the right fast-follow to resurrect.

Thanks for the quick turnaround on the parallel take — the fixture work and manifest entries in particular were cleanly done.

Issue #532, opened 2026-07-24
Process conversation history without full-history rebuilds or non-spec limits


leseb:
## Description

Praxis currently processes stored OpenAI conversation history through full-history materialization on both write and read paths:

- Conversation item mutations call `sync_conversation_messages`, which pages through every stored item, collects the complete history, and rewrites a denormalized `messages` JSON value.
- Responses rehydration loads that complete JSON value, serializes it again to enforce `max_history_bytes`, and clones accepted items into `ResponsesState`.

This makes latency and peak memory grow with the complete conversation and creates multiple simultaneous representations of the same history.

The existing `max_history_bytes` and `max_history_items` rejection path is not the correct safeguard. The OpenAI API limits each create-items request to 20 items but does not define a total conversation item-count or byte-size ceiling. Model context overflow is governed by the Responses API `truncation` setting.

Related findings: #407 and #409.

## Expected behavior

- Do not impose a Praxis-specific total conversation item or byte ceiling as part of the OpenAI-compatible API contract.
- Preserve the complete ordered conversation history for storage and rehydration.
- Avoid rebuilding the complete denormalized history after every item mutation.
- Avoid redundant full-history serialization and cloning during rehydration.
- Preserve the client's `truncation` setting and let a native Responses backend enforce its model context window.
- Keep operator-configured transport, storage, and resource safeguards separate from conversation API semantics.

## Implementation direction

- Make paginated conversation items the source of truth, or otherwise maintain history incrementally without scanning and rewriting the entire conversation on each mutation.
- Change the store/rehydration boundary so history can be read incrementally and serialized once, without a second sizing allocation or `messages.to_vec()`-style full copy.
- Remove or redefine `max_history_bytes` and `max_history_items` so valid stored history is not rejected with a non-spec 413.
- Preserve tenant isolation, item ordering, SQLite/PostgreSQL parity, and existing append-back behavior.

The exact storage representation is intentionally left open; the invariant is that processing should not create multiplicative full-history copies or require a full cache rebuild for each mutation.

## Acceptance criteria

- Rehydration no longer returns 413 solely because stored history exceeds a Praxis-specific history limit.
- Adding or deleting a conversation item does not page through and materialize every existing item merely to rebuild a cache.
- Rehydration does not serialize the full history solely to measure it and does not clone the full accepted history unnecessarily.
- Stored item order and complete-history behavior remain unchanged.
[...]

Issue #32, opened 2026-06-01
Epic: OpenAI  background_jobs


leseb:
## Purpose

Implement OpenAI Responses background mode for locally orchestrated Responses requests: asynchronous execution, bounded scheduling, polling, cancellation, timeout handling, and resumable streaming.

This filter is opt-in. When it is not configured, backends that natively support background mode continue to receive the request unchanged.

## Current OpenAI contract

The implementation must follow the current [OpenAI background mode documentation](https://developers.openai.com/api/docs/guides/background):

- `background=true` starts asynchronous processing.
- Clients poll `GET /v1/responses/{id}` while status is `queued` or `in_progress`.
- `POST /v1/responses/{id}/cancel` cancels an in-flight response; repeated cancellation is idempotent.
- `background=true` is valid with `store=false`. Temporary operational state must still be retained long enough for polling and cancellation, without turning `store=false` into durable application storage.
- `background=true` is valid with `stream=true`. A client can resume the stream with `GET /v1/responses/{id}?stream=true&starting_after={sequence_number}` when the response was originally created with streaming enabled.

The existing `stream + background` and `background + store=false` rejection rules are obsolete and must be removed as part of this work or a prerequisite conformance change.

## Architecture boundary

The filter schedules and exposes the OpenAI lifecycle; Praxis core owns background task lifecycle and execution through the configured proxy machinery.

A worker must run the request through the applicable routing, load balancing, upstream exchange, response filters, persistence, guardrails, and observability path. Do not implement this as:

- a hidden direct backend call from the filter;
- a loopback HTTP request into the public listener; or
- a second proxy implementation embedded in `background_jobs`.

The required Praxis jobs/task lifecycle work is tracked by praxis-proxy/praxis#807. The AI implementation must be based on the resulting core contract rather than inventing an incompatible local runtime.

### Enqueue

For a locally handled `POST /v1/responses` request with `background=true`:

1. Generate the stable response ID.
2. Reserve bounded queue capacity before persisting `queued`.
3. Persist the initial response object with `status=queued`.
4. Return the queued response immediately for non-streaming requests.
5. The worker atomically claims `queued -> in_progress`.
6. The worker executes the configured Responses path.
7. Completion atomically transitions `in_progress` to `completed`, `incomplete`, or `failed`.

The locally generated response ID must remain the public and stored ID even if the inference backend generates a different provider response ID.

[...]

PR #1176, opened 2026-09-14
fix(responses): reject background and truncation the Chat adapter cannot honor

### LINKED ISSUE #951: Responses-to-Chat translation silently downgrades background and truncation semantics
## Clawpatch finding

- **Severity:** medium
- **Category:** api-contract
- **Confidence:** high
- **Finding ID(s):** `fnd_sig-feat-custom-apis-openai-prot_48e25126de`
- **Run ID(s):** `20260907T071353-20798f`
- **Prior-run overlap:** No prior-run overlap found.

## Evidence

- `apis/src/openai/translation/chat_completions.rs:332-350` (`map_request_parameters`) — The Chat request mapping omits both background and truncation.
- `apis/src/openai/translation/chat_completions.rs:1331-1362` (`response_resource`) — The translated response unconditionally reports truncation as disabled and background as false.
- `apis/src/openai/translation/mod.rs:2578-2624` (`response_resource_includes_all_required_fields`) — Response tests check only the hard-coded defaults and do not provide non-default request values.

## Description

A canonical Responses request may ask for `background: true` or `truncation: auto`. The translator neither implements those behaviors nor rejects them: it sends a foreground Chat request with no truncation equivalent, then reports `background: false` and `truncation: disabled`. This is not delegated backend validation; the adapter has accepted request semantics that it cannot represent and has returned a response claiming different semantics.

## Reproduction

Translate canonical state for `{"model":"m","input":"hello","background":true,"truncation":"auto"}`. Inspect the Chat request and observe that both fields are absent. Translate a successful Chat response and observe that the resulting Responses resource contains `background:false` and `truncation:"disabled"`.

## Recommendation

Fail closed during request translation when non-default background or truncation behavior cannot be represented, or implement dedicated background-job and truncation semantics. Echoing hard-coded defaults is safe only after verifying the request used those defaults.

---
Generated by the scheduled Clawpatch Codex `gpt-5.6-sol` / `xhigh` review of `apis/src`. Validate before assigning or implementing.

cdoern:
## Problem

A canonical Responses request may ask for `background: true` or `truncation: auto`. The Chat Completions translation implemented neither and rejected neither:

- `map_request_parameters` dropped both fields, so a foreground, untruncated Chat request went upstream.
- `response_resource` then reported `background: false` and `truncation: "disabled"` back to the caller.

The adapter accepted request semantics it cannot represent and answered with semantics the caller never asked for. This is not delegated backend validation — because neither field is forwarded, the backend never sees them and cannot reject them on our behalf.

## Change

Fail closed during request translation. Implementing real background and truncation behavior belongs to the background-jobs filter (#32) and a truncation design, not to this adapter, so the honest answer here is a 400 rather than a silent downgrade.

`validate_representable_parameters` rejects:

| Field | Accepted | Rejected |
| --- | --- | --- |
| `background` | absent, `null`, `false` | `true`, and any non-boolean |
| `truncation` | absent, `null`, `"disabled"` | `"auto"`, and any non-string |

Malformed values are rejected rather than forwarded because these fields are dropped, never sent. Anything not demonstrably the default would otherwise be discarded and then reported back as the default — the same defect.

The existing `TranslationError` path already renders as a Responses-shaped `400 invalid_request_error`, so no new plumbing was needed.

This is also what makes the response resource truthful: the hard-coded `background: false` and `truncation: "disabled"` are now correct by construction. The two `TODO(responses)` comments are replaced with that reasoning.

## Scope

Only the Responses → Chat Completions adapter. Native passthrough is unaffected and still forwards `truncation` verbatim, as `test_truncation_forwarded_to_backend_through_rehydration` (#532) requires.

## Tests

- Six unit tests: `background: true` and `truncation: "auto"` rejected with exact messages; explicit defaults and `null` still translate; malformed values rejected; a rejected request produces no Chat request at all.
- New inference scenario `responses/chat-unrepresentable-parameters` with a synthetic recording, covering both turns end to end through the proxy (`client_status: 400`, no upstream call). Manifest feature `responses.chat.unrepresentable_parameters`.

Verified: `make lint` clean, `cargo test -p praxis-ai-apis` 3392 passed, `cargo test -p praxis-tests-integration` 624 passed, `make test-inference-fixtures` passes.

closes #951


leseb (changes requested):
[P2] Bound malformed values before formatting errors — [chat_completions.rs:399](https://github.com/praxis-proxy/ai/blob/6f4ebbfdc5f8d677df649e41b7ba90c002ad646d/apis/src/openai/translation/chat_completions.rs#L399-L416). A malformed background or truncation can contain nearly 64 MiB of JSON. Calling to_string() copies that entire subtree, then the rejection path formats and JSON-escapes it again without an output-size limit. Repeated requests can therefore create multiple large allocations, large responses, and potentially large debug logs. Use a bounded/type-based message for malformed values; only the recognized true/"auto" cases need their value echoed.

cdoern:
Fixed in fd126ba5.

Only the two recognized values, `true` and `"auto"`, are echoed now. Anything else is reported by JSON type, so the message stays small no matter what the client sends. The error field is a `&'static str` rather than a `String`, so the bound is enforced by the type instead of by convention.

Updated the unit tests and the inference fixture recording to match the new wording.

`make lint` clean, apis and integration suites pass, fixtures replay.

leseb (approved):

PR #1177, opened 2026-09-15
feat(routing): score candidates on live load signals

### LINKED ISSUE #1174: feat(routing): load-signals data source for cross-site load-aware routing
## What

The consumer half of the cross-site signals loop, whose publisher is the grid operator's signals endpoint. A data source polls that endpoint over HTTP(S), keeps a bounded per-(site, cluster, metric) window of Prometheus samples, and exposes them as routing-scorer inputs, so the router compares live load across every site serving a model and routes to the site with the most headroom.

## Behavior

- Poll each peer operator's signals endpoint on a coarse interval, keeping a bounded freshness window per (site, cluster, metric).
- Expose latest, fresh, and window-worst readings as scorer inputs.
- Score routing candidates on that live load, so a request lands on a site with headroom rather than a loaded one.
- Land the data source additively first, inert until a consumer exists, then wire the scorer and the routing filter in a follow-up.

## Why

Today the router picks a candidate without seeing the other sites' live load, so it cannot route away from a loaded site. The grid operator already scrapes and publishes that load when a site enables it. Reading it here makes cross-site routing load-aware, closing the loop between what the grid observes and how the gateway routes.

## Delivery

Two PRs, so each lands reviewable on its own. The first adds the data source module, self-contained and unit-tested, inert until wired. The second adds the scorer and wires it into the routing path.

## Parent

Part of praxis-proxy/ai#197.

hexfusion:
Built on #1175

## Summary

Wire the load-signals consumer onto the routing picker. Selection can score candidates on live load from the local grid operator, opt-in and off by default: with no `signals` block configured, no collector runs and the pick is byte-identical to before.

## Motivation

The data source (#1175) polls the operator and keeps a bounded per-series window, but nothing reads it yet. This is the consumer half of the cross-site signals loop: within the winning priority and locality group, route to the least-loaded member instead of taking the first by policy, while never pulling a request across a group, so locality still leads.

## What changed

A `signals` block on the intelligent_route filter opts in. Present, it starts the background collector and holds one scorer per configured signal on the filter; an empty block opts in on defaults (the operator signals endpoint and the two endpoint-picker multicluster signals). Absent, selection is unchanged.

The seam lives where selection already lives. `picker::select_candidate` takes the scorer set and a request timestamp; inside the winning group it scores the members and the best-scored wins, and a group the signal says nothing about falls back to the existing deterministic, round-robin, or weighted policy. The request clock is read once and only when scoring is active.

Scoring follows the endpoint picker's model: each signal is normalised so higher is better and comparable, scores combine by weight, a deadband suppresses spreads too small to justify moving off locality, and an unscored candidate is skipped rather than disqualifying the set. The signal weight and deadband are validated at config parse, so a non-finite or out-of-range value is rejected rather than silently poisoning the score. The load-store key is precomputed on each candidate at snapshot-build time, so the request path scores with no per-candidate allocation.

Control and data planes stay separated: the collector writes observations into the store, the request path reads a windowed aggregate from it, and the two meet only at that store.

Closes #1174


leseb (changes requested):
P1: The advertised static load-scoring configuration never scores anything.
The new example uses inline static candidates with signals. Static snapshots have no selection groups, so select_candidate takes select_legacy, which ignores the scorer list and always picks the first admitted candidate. The collector runs and consumes resources, but its data cannot change routing. Either score the matching static candidate set or reject signals with ungrouped/static candidates. [Advertised example](https://github.com/praxis-proxy/ai/blob/9706c4bfd3296673f6e37c0455b0df71bd287afd/examples/configs/intelligent-route-load-scoring.yaml#L27-L58), [picker bypass](https://github.com/praxis-proxy/ai/blob/9706c4bfd3296673f6e37c0455b0df71bd287afd/filters/src/routing/picker.rs#L26-L48).

P2: max_age_ms is silently widened to 15 seconds.
The field is documented as the age after which a sample is ignored, but scoring uses max(max_age_ms, 15_000). Configuring a stricter bound—such as 2 seconds—still permits readings to influence routing for 15 seconds. Use the configured value directly or reject unsupported shorter values during configuration. [Code](https://github.com/praxis-proxy/ai/blob/9706c4bfd3296673f6e37c0455b0df71bd287afd/filters/src/routing/scoring.rs#L115-L150).

hexfusion:
this is a grid only filter moving to that repo

Issue #1051, opened 2026-09-10
feat(anthropic): register Messages operations for request-head classification


leseb:
## Description

Declare supported Anthropic Messages operations in a protocol-owned registry so
their identities come from the HTTP request head rather than body-format
heuristics.

The registry will publish `application_protocol: anthropic_messages` through
the shared AI operation classifier. Payload validity and message extraction
remain responsibilities of Anthropic-specific request processing.

Parent epic: #740

## Scope

- Add an Anthropic Messages operation registry using the provider-neutral
  runtime operation descriptor and matcher.
- Register every Anthropic Messages operation Praxis supports, including
  `POST /v1/messages` at minimum.
- Declare exact methods, normalized runtime paths, stable operation IDs,
  transport qualifiers, and request-body shapes.
- Treat required Anthropic headers such as `anthropic-version` as contract
  validation rather than changing operation identity when the method and path
  already identify the operation.
- Publish matches through `ai_operation` with
  `application_protocol: anthropic_messages`.
- Add an explicit, checked-in source-of-truth and provenance for the supported
  operation inventory. Do not claim machine-readable specification conformance
  unless an authoritative specification is pinned.
- Prove Chat Completions-shaped, malformed, empty, and absent bodies do not
  change the operation selected from the same request head.

## Acceptance Criteria

- `POST /v1/messages` resolves to the expected Anthropic Messages operation
  without reading or buffering its body.
- Unsupported methods and malformed subresource paths publish no operation
  match.
- OpenAI and Anthropic operations use the same classifier and typed match.
- The Anthropic registry requires no Anthropic-specific branch in the generic
  matcher or classifier.
- Missing or invalid required payload fields and headers remain distinguishable
  from operation identity and are left to Anthropic request processing or the
  backend as appropriate.
- Unit tests, functional integration tests, example configuration, generated
  example inventory, and filter documentation are complete.

## Dependencies

- #1050 — generic provider-neutral operation identity

## Blocks

- #1052 — Anthropic Messages request-processing consolidation
- #1043 — operation-aware multi-protocol guardrail message extraction

## Non-Goals

- Anthropic request or response schema validation
- Anthropic-to-OpenAI translation
- Body-format classification
- Provider inference

## Area

Provider APIs (OpenAI/Anthropic)

skamenan7:
please assign it to me, @leseb . Thanks.

PR #1183, opened 2026-09-15
refactor(ai): make operation identity provider-neutral

### LINKED ISSUE #1050: refactor(ai): make operation identity provider-neutral
## Description

Generalize the runtime operation identity introduced by #746 from an
OpenAI-family model into a provider-neutral application-protocol model that
OpenAI, Anthropic, and future AI APIs can share.

The OpenAI registry and Responses operation work already merged through #746
and #743 remain the first implementation. This issue extracts the reusable
runtime identity without rewriting that completed history or coupling
provider-owned contract generation to the generic matcher.

Parent epic: #740

## Scope

- Introduce an open-ended application-protocol identifier for runtime operation
  identity, with canonical values such as `openai_responses`,
  `openai_chat_completions`, and `anthropic_messages`.
- Replace classifier-facing API-family identity with
  `application_protocol`.
- Define provider-neutral runtime operation and match types covering:
  - application protocol
  - stable operation ID
  - exact HTTP method
  - runtime path template
  - transport qualifiers
  - request-body shape
- Keep OpenAI implementation `OpenAPI` generation and owned contract metadata
  in the OpenAI boundary; it may wrap or reference the generic runtime
  descriptor.
- Migrate the existing Conversations and Responses registries without changing
  their supported methods, paths, operation IDs, body shapes, or ownership.
- Keep matching allocation-free on the normal request path where practical.
- Treat the application provider as selected-cluster metadata; do not infer or
  publish `application_provider` from the incoming request.

## Acceptance Criteria

- OpenAI and Anthropic registries can use the same runtime matcher without
  provider-specific branches in the matcher.
- The authoritative match exposes `application_protocol`, `operation_id`, and
  transport with no request-body inspection.
- Existing Responses and Conversations registry, conformance, and runtime tests
  remain green.
- Published runtime identity uses canonical application-protocol identifiers
  rather than the current `responses` / `conversations` family labels.
- Provider-specific contract generation remains independent from generic
  runtime identity.
- No request, response, header, path, or body data is cloned or buffered by the
  matcher.

## Dependencies

- #746 — initial OpenAI registry and matcher foundation (complete)

## Blocks

- #744 — generic request-head operation classifier
- #1051 — Anthropic Messages operation registration

## Non-Goals

- Selected-upstream `application_protocol` or `application_provider`
  propagation
- Request-body parsing or validation
- Provider translation
- Backend selection

## Area

Provider APIs (OpenAI/Anthropic)

cdoern:
## Problem

The runtime operation identity introduced by #746 is modelled around the OpenAI family. Anthropic Messages (#1051) and every future AI API need the same identity, so the reusable runtime parts have to leave the OpenAI boundary before a second protocol adopts them.

## Change

Adds `praxis-ai-apis::operation`, carrying the provider-neutral runtime identity and the shared matcher:

| Moved out of `openai::operation` | Stayed in `openai::operation` |
| --- | --- |
| `ApplicationProtocol` (was `OpenAiApiFamily`) | `OpenAiOperationSpec` wrapper |
| `HttpMethod`, `Transport`, `HandlingMode`, `RequestBody` | `SchemaBinding`, `ParameterSpec`, `MediaTypeSpec` |
| `OperationSpec`, `OperationEntry` | `OwnedOperationContract`, `RequestBodySpec`, `ResponseSpec` |
| `RouteParams`, `MatchedOperation`, `match_operation` | `implementation_openapi`, `owns_contract` |

### The protocol identifier is open-ended

`ApplicationProtocol` is a newtype over `&'static str`, not an enum, so registering a protocol does not mean editing one shared list. Each registry declares its own identifier beside its operations:

```rust
// apis/src/openai/conversations/routes.rs
const APPLICATION_PROTOCOL: ApplicationProtocol = ApplicationProtocol::new("openai_conversations");
```

Only the two protocols that have a registry today are declared. Empty identifiers are rejected.

### Contract generation stays provider-owned

`OpenAiOperationSpec` wraps `runtime: OperationSpec` and keeps what OpenAI owns: `spec_path` and `owned_contract`. `owns_contract()` lives on that wrapper too — it is `OpenAPI` generation policy rather than runtime identity. The matcher reads the shared half through `OperationEntry`, so it carries no `OpenAPI` dependency and no provider branches. Utoipa's own `HttpMethod` is aliased and converted locally by `openapi_method()`.

### Registries migrate through their macros

Conversations and Responses both declare specs inside a macro, so the shape change touched one place each. Supported methods, paths, operation IDs, body shapes, and ownership are unchanged.

## Acceptance criteria

- [x] The authoritative match exposes `application_protocol`, `operation_id`, and transport with no request-body inspection
- [x] Existing Responses and Conversations registry, conformance, and runtime tests stay green
- [x] Published runtime identity uses canonical application-protocol identifiers
- [x] Provider-specific contract generation remains independent of generic runtime identity
- [x] Nothing is cloned or buffered by the matcher — path parameters still borrow from the request path
- [ ] **OpenAI and Anthropic registries use the same runtime matcher with no provider-specific branches.** Not demonstrated here. There is no Anthropic registry yet and `classify()` is still an OpenAI chain. `the_matcher_does_not_branch_on_protocol` matches a hand-built slice, which shows the matcher is protocol-agnostic but is not evidence that two shipped registries share it. #1051 closes this.

## Not in scope

Renaming the filter to `ai_operation`. The issue does not ask for it and it would break existing configuration.

## Tests

Six tests in the shared module: protocol-agnostic matching, literal-over-parameter precedence, borrowed parameters, transport as identity, path normalization, open-ended identifiers. The matcher tests that previously lived in `openai::operation` moved with the code, including the duplicate-parameter and capacity invariants.

Verified: `make lint` exit 0, `cargo clippy --workspace --all-targets` clean, `cargo test -p praxis-ai-apis` 3464 passed + 4 doctests, `cargo test -p praxis-tests-integration` 644 passed.

closes #1050


leseb (changes requested):
since it'zs provider neutral now it's should be outside of the openai dir

apis/src/operation.rs
```diff
@@ -0,0 +1,480 @@
...
+//! template, and request-body shape — without naming any provider.
+//!
+//! Contract generation stays with the provider that owns it. A registry may
+//! wrap [`OperationSpec`] in a richer type carrying provider-specific metadata
+//! and expose the shared part through [`OperationEntry`], so the matcher reads
+//! one representation while `OpenAPI` generation keeps its own.
+
+/// Application protocol an operation belongs to.
+///
+/// Deliberately open-ended rather than an enum: registering a new protocol is
```
mkoushni: The comment says a new protocol must not edit a central list, then this impl is that list. ANTHROPIC_MESSAGES, OPENAI_CHAT_COMPLETIONS, OPENAI_FILES, and OPENAI_VECTOR_STORES have no registry in this PR. Keep the newtype and new(); put each constant beside the registry that owns it.

apis/src/operation.rs
```diff
@@ -0,0 +1,480 @@
...
+    /// `OpenAI` Files API.
+    pub const OPENAI_FILES: Self = Self("openai_files");
+    /// `OpenAI` Responses API.
+    pub const OPENAI_RESPONSES: Self = Self("openai_responses");
+    /// `OpenAI` Vector Stores API.
+    pub const OPENAI_VECTOR_STORES: Self = Self("openai_vector_stores");
+
+    /// Declare a protocol identifier not covered by the canonical constants.
+    #[must_use]
+    pub const fn new(value: &'static str) -> Self {
```
mkoushni, line 38: ApplicationProtocol::new("") is legal. Empty identifiers should fail construction or a registry test.

apis/src/operation.rs
```diff
@@ -0,0 +1,480 @@
...
+    pub const fn as_str(self) -> &'static str {
+        match self {
+            Self::Passthrough => "passthrough",
+            Self::Inspect => "inspect",
+            Self::Transform => "transform",
+            Self::Local => "local",
+        }
+    }
+
+    /// Whether Praxis owns the operation's externally visible contract.
```
mkoushni: owns_contract is OpenAPI generation policy, not runtime identity. OwnedOperationContract already lives on OpenAiOperationSpec. Move this method there (or into the generator) and drop it from the shared HandlingMode.

apis/src/operation.rs
```diff
@@ -0,0 +1,480 @@
...
+pub struct RouteParams<'a> {
+    /// Captured parameter names paired with their borrowed path segments.
+    pairs: [(&'static str, &'a str); MAX_PATH_PARAMS],
+    /// Number of occupied slots.
+    len: usize,
+}
+
+impl<'a> RouteParams<'a> {
+    /// Record one captured parameter, failing when capacity is exhausted.
+    fn insert(&mut self, name: &'static str, value: &'a str) -> Option<()> {
```
mkoushni, line 274: insert no longer rejects a repeated parameter name. {id}/{id} now matches; get still returns the first value. Restore the if self.get(name).is_some() { return None; } check.

apis/src/operation.rs
```diff
@@ -0,0 +1,480 @@
...
+                } else if template_segment != path_segment {
+                    return None;
+                }
+            },
+            _ => return None,
+        }
+    }
+}
+
+#[cfg(test)]
```
mkoushni, line 380: rejects_duplicate_parameter_name and rejects_more_parameters_than_capacity were deleted with the old openai/operation.rs tests and not moved here. Put them back next to parameters_borrow_from_the_request_path.

apis/src/operation.rs
```diff
@@ -0,0 +1,480 @@
...
+            method: HttpMethod::Post,
+            transport: Transport::Http,
+            runtime_path: "/v1/messages",
+            mode: HandlingMode::Passthrough,
+            request_body: RequestBody::Json { required: true },
+        },
+    ];
+
+    #[test]
+    fn one_matcher_serves_several_protocols() {
```
mkoushni: one_matcher_serves_several_protocols matches a hand-built slice, not OpenAI and Anthropic registries. That does not satisfy the “two registries, one matcher” checkbox. Drop or reword the claim until Anthropic has a real registry. Same in line 388

apis/src/operation.rs
```diff
@@ -0,0 +1,480 @@
...
+        }
+    }
+}
+
+/// Runtime identity of one operation, shared by every registry.
+///
+/// Carries only what resolving a request head requires. Provider-owned
+/// specification provenance and generated contracts live in the provider's own
+/// wrapper type.
+#[derive(Clone, Copy, Debug)]
```
mkoushni, line 196: These were pub(crate) and should stay that way. Anthropic is in this crate; pub freezes API. Clippy will not catch it because of #![allow(unreachable_pub)] in apis/src/lib.rs:4. Same for pub mod operation in apis/src/lib.rs:21 and code in current file lines 250 / 266 / 274 / 303 / 333

apis/src/openai/operation.rs
```diff
@@ -567,14 +284,14 @@ where
         let Some(contract) = spec.owned_contract() else {
             continue;
         };
-        if !spec.mode.owns_contract() {
+        if !spec.mode().owns_contract() {
```
mkoushni: spec.mode().owns_contract() pulls shared HandlingMode into contract generation. After moving owns_contract onto OpenAiOperationSpec, call it here without going through the runtime enum.

apis/src/openai/operation_classifier/mod.rs
```diff
@@ -176,23 +173,25 @@ fn publish_match(ctx: &mut HttpFilterContext<'_>, matched: OpenAiOperationMatch)
 ///
 /// Families are consulted in registration order. Their path spaces do not
 /// overlap, so at most one can match a given method and path.
-fn classify(method: &str, path: &str, transport: OpenAiTransport) -> Option<OpenAiOperationMatch> {
+fn classify(method: &str, path: &str, transport: Transport) -> Option<OpenAiOperationMatch> {
```
mkoushni, line 181: Production classify() is still Conversations then Responses. Unmatched still logs “no OpenAI operation matched” (122-128). Fine for this PR; do not claim Anthropic already shares this path. When a second registry lands, iterate registries instead of adding another if let.

apis/src/openai/conversations/openapi.rs
```diff
@@ -72,8 +72,8 @@ mod tests {
             .collect::<BTreeSet<_>>();
         let expected = operation_specs()
             .iter()
-            .filter(|spec| spec.owned_contract().is_some() && spec.mode.owns_contract())
-            .map(|spec| (spec.method.as_str(), spec.spec_path))
+            .filter(|spec| spec.owned_contract().is_some() && spec.mode().owns_contract())
```
mkoushni: OpenAPI export still gates on spec.mode().owns_contract(). After moving owns_contract onto OpenAiOperationSpec, this filter should use that wrapper, not shared HandlingMode.

apis/src/openai/conversations/routes.rs
```diff
@@ -412,24 +419,24 @@ mod tests {
...
         let operation_ids = OPERATION_SPECS
             .iter()
-            .map(|spec| spec.operation_id)
+            .map(|spec| spec.operation_id())
             .collect::<BTreeSet<_>>();
         assert_eq!(operation_ids.len(), OPERATION_SPECS.len());
-        assert!(OPERATION_SPECS.iter().all(|spec| spec.mode == OpenAiHandlingMode::Local
-            && spec.mode.owns_contract()
+        assert!(OPERATION_SPECS.iter().all(|spec| spec.mode() == HandlingMode::Local
+            && spec.mode().owns_contract()
```
mkoushni: Registry invariant uses spec.mode().owns_contract() for contract ownership. Same coupling as above; switch to OpenAiOperationSpec once owns_contract moves off the shared enum.

apis/src/openai/conversations/routes.rs
```diff
@@ -459,22 +466,22 @@ mod tests {
...
         }
     }
 
     #[test]
     fn handling_modes_classify_contract_ownership() {
-        assert!(!OpenAiHandlingMode::Passthrough.owns_contract());
-        assert!(!OpenAiHandlingMode::Inspect.owns_contract());
-        assert!(OpenAiHandlingMode::Transform.owns_contract());
-        assert!(OpenAiHandlingMode::Local.owns_contract());
+        assert!(!HandlingMode::Passthrough.owns_contract());
```
mkoushni: Unit tests assert HandlingMode::*::owns_contract() on the shared type. If owns_contract moves to OpenAI-only code, these tests should move with it (e.g. next to openai/operation.rs tests), not stay on crate::operation::HandlingMode.

mkoushni (commented):
Good points:
OperationSpec vs OpenAiOperationSpec { runtime, spec_path, owned_contract } is the correct seam.
Matcher still ranks literal segments over parameters; transport is still part of identity; path params still borrow.
openapi_method() stays next to Utoipa. Shared matcher has no OpenAPI dependency.
Not renaming openai_operation is the right product call.

Point to fix:
Runtime identity belongs in praxis-ai-apis::operation; OpenAPI generation can stay in openai::operation. The PR over-claims “Anthropic uses this matcher,” quietly drops a matcher invariant, and recreates the central protocol list it says it is removing.

@leseb’s “move it out of openai/” is already half-done. apis/src/operation.rs is the shared module. What remains in openai/ is the OpenAI contract wrapper plus openai_operation. Do not move OpenAPI generation. Do not rename the shipped filter in this PR.

Acceptance criteria are not met.

The write-up says OpenAI and Anthropic registries share one matcher. There is no Anthropic registry in this PR. one_matcher_serves_several_protocols builds a synthetic slice. Production classify() is still an OpenAI if-chain (Conversations, then Responses).

That is a fine shape for a later Anthropic PR. It is not evidence that two real registries already share the matcher. Drop or reword that checkbox.

apis/src/operation.rs
cdoern: Fixed. I retyped this instead of moving it and lost the guard. Restored, and `{id}/{id}` now fails to match again.

apis/src/operation.rs
cdoern: Both restored, next to the borrow test. Added a third that drives a repeated-name template through `match_operation`, so the invariant is covered end to end and not just on `insert`.

apis/src/operation.rs
cdoern: Fixed. The shared module no longer names any protocol. Each registry declares its own beside its operations, and only the two that exist are declared — the other four were placeholders and are gone.

apis/src/operation.rs
cdoern: Fixed. `new()` asserts non-empty, so a const call site fails to compile. Added a `should_panic` test for the runtime path.

apis/src/operation.rs
cdoern: Moved to `OpenAiOperationSpec`, next to `owned_contract`. Dropped from `HandlingMode`.

apis/src/openai/operation.rs
cdoern: Updated — now `spec.owns_contract()`.

apis/src/openai/conversations/openapi.rs
cdoern: Updated — now `spec.owns_contract()`.

apis/src/openai/conversations/routes.rs
cdoern: Updated — now `spec.owns_contract()`.

apis/src/openai/conversations/routes.rs
cdoern: Moved. It is now a registry assertion in this file (every Conversations operation owns its contract), and the mode-by-mode check sits with the OpenAI tests.

apis/src/operation.rs
```diff
@@ -0,0 +1,480 @@
+// SPDX-License-Identifier: MIT
```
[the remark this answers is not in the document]
cdoern: Fixed — Apache-2.0.

apis/src/operation.rs
cdoern: All six are back to `pub(crate)`. You are right that the crate-level `allow(unreachable_pub)` is why clippy stayed quiet.

One I did not change: `pub mod operation` and `HandlingMode`. xtask reads `praxis_ai_apis::operation::HandlingMode` for conformance reporting, so making them crate-private breaks the build. Happy to re-export it elsewhere and tighten the module if you'd prefer, but that felt bigger than this PR.

apis/src/operation.rs
cdoern: Agreed, that was an over-claim. Renamed to `the_matcher_does_not_branch_on_protocol` and documented that the slice is hand-built and not two shipped registries. Also unticked that box in the PR description and pointed it at #1051.

apis/src/openai/operation_classifier/mod.rs
cdoern: Noted in the code — `classify()` now says only OpenAI protocols are registered, and that a second provider should iterate a registry list rather than add another `if let`. No Anthropic claim anywhere now.

cdoern:
Thanks both, pushed in b66bc676.

The important one was the dropped duplicate-parameter guard. I retyped `insert` instead of moving it, lost the check, and deleted the two tests that would have caught it. Restored, plus a third test that drives a repeated-name template through the matcher.

Also done: protocol constants now live beside the registry that owns them (the shared module names none, and the four speculative ones are gone), `owns_contract` moved onto `OpenAiOperationSpec`, visibility back to `pub(crate)`, empty identifiers rejected, license header corrected, and the misleading matcher test renamed.

I unticked the "OpenAI and Anthropic share one matcher" box in the description and pointed it at #1051 — you were right that a hand-built slice is not evidence for it.

One deviation, called out inline: `pub mod operation` and `HandlingMode` stay public because xtask reads them across the crate boundary. Everything else on that list is crate-private again.

cdoern:
Merged main in (0e26bc68) now that #1204 has landed.

The new Chat Completions registry migrated onto the shared types and declares its own `APPLICATION_PROTOCOL` beside its operations, so it never touches the shared module — which is the point of the change.

I took main's `classify()` over mine. It consults HTTP-only registries from a list instead of chained `if let`s, which is what @mkoushni asked for in review; I just adapted it to the generic types. Same for the extracted `registry_check` module.

Verified locally: workspace builds, clippy clean, `make doc` exit 0, 3751 lib tests pass.

leseb (approved):

PR #1202, opened 2026-09-16
fix(ci): test Basic Auth at the package that exposes it

nerdalert:
## Summary

- separate the experimental feature lists used by `praxis-ai-filters` and the integration-test crate
- stop passing `basic-auth-filter` to `praxis-ai-filters`, which does not expose that feature
- test Basic Auth directly through `praxis-ai-proxy`, where the workspace feature is exposed
- retain Basic Auth in the integration feature set forwarded through `praxis-tests-integration`

## Why

`make test-unit` passed `basic-auth-filter` to `praxis-ai-filters`, causing Cargo to fail before running that package's tests:

```text
error: the package 'praxis-ai-filters' does not contain this feature: basic-auth-filter
```

Basic Auth is implemented by the underlying filter crate and exposed in this workspace through `praxis-ai-proxy`. The filters package should only receive features it actually defines. This change preserves Basic Auth coverage rather than removing the feature from validation.

## Validation

- `make test-unit`
- existing test suite completed successfully after correcting the package/feature pairing


nerdalert:
@leseb apologies, I managed to break this with #980.

shaneutt (approved):

leseb:
> @leseb apologies, I managed to break this with #980.

Np!

PR #1055, opened 2026-09-10
ci: add vLLM GPU smoke workflow with dynamic EC2 runner lifecycle (#541)

skamenan7:
## Summary

Adds a manual GPU smoke workflow for [#541](https://github.com/praxis-proxy/ai/issues/541). It runs the existing stateless Responses API test through Praxis against vLLM on an ephemeral AWS GPU runner.

## Flow

1. **Start runner**: GitHub Actions uses OIDC to assume the Praxis AWS role and starts an ephemeral `g4dn.xlarge` runner.
2. **Run smoke test**: The runner verifies the GPU, starts the pinned vLLM image with `Qwen/Qwen3-0.6B`, starts OGX, builds Praxis, waits for both services, and runs `test_stateless_request`.
3. **Clean up**: A final job always terminates the EC2 runner, including when the test fails.

## Safety and validation

- Manual dispatch only.
- AWS access uses GitHub OIDC and the repository-scoped role from [aaet-infra MR !33](https://gitlab.com/redhat/ai/midstream-integration/aaet-infra/-/merge_requests/33).
- The [vLLM GPU workflow](https://github.com/praxis-proxy/ai/blob/feat/541-vllm-gpu-smoke-ec2/.github/workflows/vllm-gpu-smoke.yaml) uses an image pinned by digest.
- Failure logs are captured and the GPU runner is cleaned up.
- Action lint, shellcheck, Rust build, and test collection pass locally.

The first live GPU run is still pending access to the configured repository variables and runner-registration secret.

PR #1225, opened 2026-09-17
test: trigger vLLM GPU smoke workflow (#541)

leseb:
Base-repo copy of #1055 to enable manual `workflow_dispatch` of the vLLM GPU smoke workflow (needs base-repo AWS OIDC role + runner secrets that a fork cannot access).

Temporary/test branch for triggering the workflow. Original change: #1055.

Refs #541.


franciscojavierarceo (changes requested):
the workflow still has two medium issues to resolve. first, HUGGING_FACE_HUB_TOKEN is passed into the vLLM container without being sourced from repository secrets or variables, so it is always empty; remove the unused export for this public model or wire it to an explicit secret. second, both ogx installs resolve an unpinned ogx[starter], while tests/integration/ogx-constraints.txt already provides the repository's single source of truth (currently ogx[starter]==1.3.1); use that pin/constraints file so the GPU smoke environment cannot drift. please rerun the workflow after these changes. this is a live GPU smoke test and the repository's fixture policy does not require a committed inference recording for it.

PR #858, opened 2026-09-01
fix(token_count): preserve OpenAI Responses cache writes and parse once

leseb:
## Summary

Follow-up to #714 fixing two issues in the newly added OpenAI Responses token-usage parsing. The Responses parser discarded `cache_write_tokens` (always emitting no `token.cache_write` metadata), which could undercount billable cache activity, and it deserialized each payload up to three times, so most SSE events were parsed three times. The parser now reads each payload once into a single envelope with an untagged top-level usage enum—classifying it as Chat Completions, non-streaming Responses, or nested streaming Responses usage—and preserves both Responses cache reads and writes. The OpenAI projection types stay beside the parser because they are intentionally partial, filter-specific wire types, not reusable canonical OpenAI models.

## Related issue

Follow-up to #714, addressing the code-review findings on the newly merged OpenAI Responses token extraction. No separate tracking issue.

## Validation

- [x] Unit tests — `cargo test -p praxis-ai-filters token_usage` (JSON + SSE token totals and both cache-metadata fields, incl. a nonzero cache-write count); full `make test` (17 suites, 0 failures)
- [x] Integration or functional tests — covered by the workspace `make test` run
- [x] `make lint` — clippy, rustfmt, dependency, docs, and example/README-sync checks all pass

## Checklist

- [x] I reviewed every changed line and can explain the change.
- [x] New capabilities include an example config and functional example test. — N/A: bug fix to an existing parser, no new capability or config.
- [x] User-facing behavior and generated documentation are updated. — `token.cache_write` is now reported for OpenAI Responses; generated READMEs verified in sync.
- [x] Performance-sensitive changes include appropriate benchmark or load-test evidence. — Removes two redundant deserializations per payload; behavior covered by tests, no separate benchmark.
- [x] Commits are signed and include a `Signed-off-by` trailer.

## Breaking changes

None. Metering now reports `token.cache_write` for OpenAI Responses where it previously reported nothing; token totals are unchanged.

PR #1233, opened 2026-09-17
fix(token_usage): extract cache_write_tokens from OpenAI Chat Completions usage

### LINKED ISSUE #1034: fix(token_count): cache_write_tokens on Chat Completions path — verify premise, then add or close
## Problem

OpenAI reports `cache_write_tokens` in `input_tokens_details` for both Chat Completions and Responses API formats. The parser only extracts `cached_tokens` (cache reads), causing `cache_creation_tokens` to always be 0 in the metering database for OpenAI requests.

This means cost calculations cannot distinguish fresh input from cache writes, and the cache-write cost tier is never applied.

## Fix

Add `cache_write_tokens: Option<u64>` to both:
- `OpenAiPromptTokensDetails` (Chat Completions)
- `ResponsesApiInputDetails` (Responses API)

Wire through all three OpenAI parse paths (Chat Completions, Responses API non-streaming, Responses API SSE).

## Impact

- OpenAI cache write costs are billed at $0 instead of the cache-write rate
- For providers like gpt-4.1 and gpt-5.6-luna, this understates actual cost
- Fix deployed on dogfood (commit `68196cae`)

## Ref

- Dogfood commit: `68196cae`
- Part of the Responses API parsing work (#1033)

noyitz:
## Problem

Issue #1034: Chat Completions prompt caching reports cache writes as `usage.prompt_tokens_details.cache_write_tokens` once a cache entry is created — the same field the Responses API exposes under `input_tokens_details`. The OpenAI parser hardcoded the Chat Completions cache write count to `None`, based on the assumption (introduced with the #858 restructuring) that OpenAI Chat Completions never reports cache writes. Dogfood gateway traffic shows that assumption is wrong: real provider responses do include `cache_write_tokens`, and those writes were metered as plain prompt tokens with no cache breakdown.

## Change

- Add `cache_write_tokens` to `OpenAiPromptTokensDetails` and extract it in `chat_completions_usage`, alongside the existing `cached_tokens` read.
- Both cache counts stay breakdowns of `prompt_tokens`, not additions to it (matching the Responses API path).
- An absent field still yields an absent count, not a zero — the `NO_CACHE_WRITE` constant now only documents the Google path, which genuinely has no cache write field.
- Tests: new `openai_chat_completions_reports_cache_writes`; the two tests that asserted "OpenAI never reports cache writes" were corrected (absent-field cases still assert `None`).

Complements #858 (which added cache-write extraction for the Responses API path only).

## Testing

```console
cargo test -p praxis-ai-filters token_usage   # 182 passed
cargo clippy -p praxis-ai-filters --all-features
```

Closes #1034


filters/src/token_usage/providers.rs
```diff
@@ -9,10 +9,11 @@ use super::TokenUsage;
...
-/// `OpenAI` Chat Completions and Google expose how much of the prompt was *read*
-/// from their cache but not how much was written to it, so those parsers leave
-/// the cache write count absent rather than claiming a zero the provider never
-/// reported.
+/// Google exposes how much of the prompt was *read* from its cache but not how
+/// much was written to it, so that parser leaves the cache write count absent
+/// rather than claiming a zero the provider never reported. (`OpenAI` reports
+/// cache writes as `cache_write_tokens` in both Chat Completions
+/// `prompt_tokens_details` and Responses API `input_tokens_details`.)
 const NO_CACHE_WRITE: Option<u64> = None;
```
alexsnaps: I'd drop that entirely, and just pass `Option::default()` to the one `.with_cache` remaining.
Probably much more idiomatic if anything, and the `const` really adds nothing anymore. 
The comment there as well, is _now_ (when you don't have the PR & context) more confusing... why is it even talking about OpenAI _not_ using it  ¯\_(ツ)_/¯

alexsnaps:
Your DCO signoff not matching `author: `?

alexsnaps (changes requested):

yossiovadia:
@alexsnaps both flags point at the pre-rebase history. The bot's unsigned commit (`c725d05`) and the signoff/author mismatch refer to commits that no longer exist on this branch — it was force-pushed at 20:25Z (after your comment) down to a single commit, `ab19b96`.

At the current head:

- author: `Yossi Ovadia <yovadia@redhat.com>`
- trailer: `Signed-off-by: Yossi Ovadia <yovadia@redhat.com>` — exact match
- DCO, signoff-check and signed-commits-check all pass

Nothing outstanding from the commit-author side; happy to take another look when you get to it.

filters/src/token_usage/providers.rs
leseb: how about just using None instead of Option::default() to convey the absence more clearly?

leseb (commented):
[MINOR] Example contradicts OpenAI cache-write support
providers.rs:15
What: token-counting.yaml:15 still says OpenAI never sets token.cache_write.
Impact: Operators receive the opposite of the new behavior.
Fix: Remove OpenAI from that statement.

filters/src/token_usage/providers.rs
alexsnaps: Sure! I have no strong opinion. I prefer `Option::default()` over `None` when you actually want to convey that the receiver will treat the absence as "let's use the default". But that's not what this does. 

I'm mostly caring about getting rid of that `const` that's really useless _and_ now has more confusing comments associated with it than actually being useful.

alexsnaps:
@yossiovadia see the comments on the code. It probably would be good to address @leseb 's comment too, which is an another instance of exactly why I don't think we should that `const`. This decouples comments in space (_and_ in time with this PR) from where they actually might be useful.

filters/src/token_usage/providers.rs
alexsnaps: To be clear, the builder pattern _is_ using `..Default()` for those fields. Just then that `.with_cache` "just" forces writing to both fields. That's how I settled on `Option::default()` (which itself calls is `None`); sorry long winded story, but that was the reasoning  ¯\_(ツ)_/¯

noyitz: Good catch — the const was a leftover from when every provider path referenced it. Dropped in d4fa7ca6: the sole remaining call site now passes `None` inline (taking leseb's wording over `Option::default()` — per the last two replies, absence here really is just absence, no default-semantics to convey). The Google-specific rationale the const doc carried moved into the `parse_google` docs, so the "why absent, not zero" reasoning survives at the site that exercises it.

noyitz:
Review feedback addressed in d4fa7ca6:

- **`NO_CACHE_WRITE` const dropped** (per @alexsnaps / @leseb) — the sole remaining `.with_cache` call site now passes `None` inline; the Google-specific "absent, not zero" rationale moved into the `parse_google` docs where it is actually exercised.
- **Example docs corrected** (@leseb's [MINOR]) — `token-counting.yaml` no longer claims OpenAI never sets `token.cache_write`; it now states OpenAI reports cache writes as `cache_write_tokens` on both Chat Completions and Responses usage, with Google as the provider that still never sets the count.
- Grep sweep found no other occurrences of the stale claim in code, docs, or examples.

`cargo test -p praxis-ai-filters token_usage` 182/182, `fmt --check` and clippy clean locally.

alexsnaps (approved):

leseb (approved):

PR #1234, opened 2026-09-17
feat(guardrails): support NeMo guardrail configuration IDs

pierDipi:
NeMo checks can require an explicit guardrails.config_ids selection, but our provider only sent model and messages and rejected this setting. Add optional provider.guardrails configuration and include it in both request- and response-phase checks. Omit the field when unconfigured to preserve the existing payload.

Borrow the configured model and guardrail selection while serializing the outgoing request. Serialization only reads this data, so owning another model String would allocate and copy it unnecessarily on every callout.

Update both example configurations and generated NeMo documentation. Cover configured and omitted selections in unit tests and assert the outbound payload through both functional example tests.


leseb (changes requested):
[P2] The examples hard-code model: "check-model" (request:47, response:31). NeMo 0.24 treats any non-empty request model as an override of the selected configuration’s main model ([source](https://github.com/NVIDIA-NeMo/Guardrails/blob/v0.24.0/nemoguardrails/server/api.py#L439-L484)). Thus, replacing only the explicitly marked config_ids placeholder can make the runnable example call a nonexistent model. Omit model to use the selected config’s model, or mark it as another required replacement. The mock integration tests currently lock in the problematic placeholder without exercising real NeMo behavior.

pierDipi:
Agree but it's a totally valid use case to support, I'll remove the model from examples and add only comments about the usage of model in combination with config_ids

leseb (approved):

leseb:
@pierDipi see CI failures

pierDipi:
@leseb when you can please take a look again, small formatting issue and it got removed from the merge queue

christinaexyou (approved):

PR #1235, opened 2026-09-18
fix(file-resolve): make outbound_chain optional to match callout filters

leseb:
## Summary

PR #1186 made `openai_file_resolve`'s `outbound_chain` effectively required — registration failed the pipeline build whenever the key was omitted — which diverged from the sibling SSRF-hardening filter `openai_file_search_callout` (PR #1185) that treats the chain as optional and defaults it to an empty inline passthrough. This change makes `outbound_chain` optional on `openai_file_resolve` too (`#[serde(default)]` → empty inline passthrough chain), so all three outbound-callout filters (`openai_web_search`, `openai_file_search_callout`, `openai_file_resolve`) share one identical pattern. An omitted chain still routes configured `file_id` callouts through the bound outbound pipeline with `allow_private_upstreams` enforcement, so SSRF protection is unchanged; a provided-but-unbuildable chain still fails the build. `file_resolve` continues to accept named chains because it runs at the top pipeline level, not nested inside an `iterative_request_router` step.

## Related issue

Refs #1186

## Validation

- [x] Unit tests — `cargo test -p praxis-ai-apis -- file_resolve` (252 passed), `cargo test -p praxis-ai-filters -- register` (7 passed, incl. new `file_resolve_binds_when_outbound_chain_omitted` and `file_resolve_rejects_unbuildable_outbound_chain`)
- [x] Functional/registration tests — pipeline `build_with_chains` proves an omitted chain binds and an unbuildable chain fails closed; existing `examples/configs/openai/responses/file-resolve.yaml` remains valid
- [x] `make lint` (exit 0), `make doc` (exit 0), `make build` (exit 0)

## Checklist

- [x] No new capability; the existing example config already exercises `outbound_chain` and stays valid.
- [x] User-facing behavior and generated documentation are updated (regenerated `docs/filters/openai_file_resolve.md`).
- [x] Not performance-sensitive; no benchmark needed.
- [x] Commit is SSH-signed and carries a single `Signed-off-by` trailer.

## Breaking changes

None — strictly permissive. A configuration that previously failed the build (omitting `outbound_chain`) now succeeds with an empty inline passthrough default; configurations that already specify `outbound_chain` are unaffected.


franciscojavierarceo (approved):
reviewed the current head against main and the existing review context. the config defaulting, registration path, and focused tests are consistent with the file-search callout pattern. no actionable correctness, security, or architectural issues found, and no inference recording is needed for this deterministic config/registration change.

PR #1243, opened 2026-09-20
fix/385 forwarding audit

szedan-rh:
- **fix(anthropic): reject Chat Completions fields whose output the response discards**
- **fix(anthropic): treat null output_config values as unset**


leseb (changes requested):
[request.rs:34](https://github.com/praxis-proxy/ai/blob/eb255f450eb3453db800ee16ba86fb7dacfd1cab/apis/src/anthropic/messages_to_chat_completions/request.rs#L34): rejects harmless defaults including n:1, logprobs:false, modalities:["text"], and function_call:"none". Reject only effectful values.
[request.rs:41](https://github.com/praxis-proxy/ai/blob/eb255f450eb3453db800ee16ba86fb7dacfd1cab/apis/src/anthropic/messages_to_chat_completions/request.rs#L41): moderation remains forwarded, but both response translators discard its results, silently losing requested safety data.
[request.rs:731](https://github.com/praxis-proxy/ai/blob/eb255f450eb3453db800ee16ba86fb7dacfd1cab/apis/src/anthropic/messages_to_chat_completions/request.rs#L731): broad null removal accepts arbitrary output_config.foo:null, bypassing the documented unsupported-key 400 and additionalProperties:false. Remove nulls only for known nullable keys.

szedan-rh:
@leseb handled, thanks for the review

leseb (approved):

